PR #52, opened 2019-06-17
fix(delete_job): keyword argument postData

Sijoma:
- keyword argument postData should not be passed to the delete request


m-mohr:
I think this PR doesn't cover all changes:
* It should also be removed here: https://github.com/Open-EO/openeo-python-client/blob/379088a8247df2ff8ecc25e4b64da62fec8a5a20/openeo/rest/rest_file.py#L59
* `postdata` should be removed here: https://github.com/Open-EO/openeo-python-client/blob/4aadc1e3992a4aa4ec4d8b60f373d66185475b0f/openeo/rest/rest_connection.py#L467
* https://github.com/Open-EO/openeo-python-client/blob/4aadc1e3992a4aa4ec4d8b60f373d66185475b0f/openeo/rest/rest_connection.py#L477 `jsondata=postdata` should be removed (or set to `None`).

jdries:
Hi, 
 could you please indicate whether work is still ongoing here, or if I should review and merge?

m-mohr:
@Sijoma Could you do these changes, check back with the validation engine and remove the custom python client from there?

soxofaan:
FYI: the `postData` argument was removed from `RESTConnection.delete` by PR#63
which covers the changes listed by @m-mohr above

m-mohr:
Very much appreciated, thanks!

soxofaan:
I looked into why I didn't find the delete calls in rest/job.py and it's because of the `Connection` type hint of https://github.com/Open-EO/openeo-python-client/blob/6f16dfcbdf00d7db88f8ad75dbf18bb4a2456441/openeo/rest/job.py#L19

If you change that to 
```
from openeo.rest.rest_connection import RESTConnection
...
    def __init__(self, job_id: str, connection: RESTConnection):
```

my IDE (PyCharm) finds them without problem

Morever, it also marks the following wrong `postdata` usage in a get call: 
 https://github.com/Open-EO/openeo-python-client/blob/6f16dfcbdf00d7db88f8ad75dbf18bb4a2456441/openeo/rest/job.py#L43-L46


So @Sijoma , you could also include these two changes

Sijoma:
@soxofaan using RESTConnection does not work. I get an import error. I now use the Connection again and removed postData.

I also removed postData in the stop_job call.

https://github.com/Open-EO/openeo-python-client/pull/52/commits/1faf5c5e85ca4357555e4d8475d8c00ea8d6f316#diff-6c49aca1f62427073bef28459af2efc2R60

jdries:
It seems like we indeed need to resolve the usage of the connection vs restconnection class, as the current state violates the type hints. (It wouldn't even compile in a typed language.) This is however out of scope for this pull request, so will merge it.

PR #54, opened 2019-06-21
use `load_collection` instead of `get_connection` for API version 0.4.0 and above

soxofaan:
- use "load_collection" instead of "get_collection" for API version >= 0.4.0
- in unit tests: just use `assert a == b` instead of `print(json.dumps(...)); self.assertDictEqual(...)` as pytest  already gives actionable difference information by default


soxofaan:
note: this PR conflicts a bit with #55 in the sense that #55 makes version checking more generic/simple

PR #96, opened 2019-10-31
Adds process linear_scale_range

lforesta:
in the client the process stretch_colors was correctly deprecated in favor of linear_scale_range, but this one was not present.

### DIFF
--- a/openeo/rest/imagecollectionclient.py
+++ b/openeo/rest/imagecollectionclient.py
@@ -617,7 +617,7 @@ class ImageCollectionClient(ImageCollection):
         process_id = 'apply'
         arguments[data_argument] = \
             {
-                "from_argument": "data"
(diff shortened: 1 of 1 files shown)


soxofaan:
minor note: this `linear_scale_range(self, inputMin, inputMax, outputMin, outputMax) `  does not follow the standard python code style convention of using under_scores instead of dromedaryCase for variables and arguments. It should be something like

```
def linear_scale_range(self, input_min, input_max, output_min, output_max) -> 'ImageCollection':
```

soxofaan:
I can quickly fix it myself, unless you want to do it @lforesta

lforesta:
ah yes, I actually copy-pasted that from the openeo-processes specification, please @soxofaan feel free to change that, thanks! I'll try to remember that in case of other PRs :)

Issue #74, opened 2019-08-28
Drop requirements.txt


soxofaan:
There is duplication between `requirements.txt`  https://github.com/Open-EO/openeo-python-client/blob/e72ca6f3f0de109ea70c41ba67b7d22b503188f5/requirements.txt#L1-L6
and install_requires in `setup.py` https://github.com/Open-EO/openeo-python-client/blob/e72ca6f3f0de109ea70c41ba67b7d22b503188f5/setup.py#L46-L51

As far as I understand the python packaging world, projects that are to be used as library in an existing env or application, like the op openeo client, should only define dependencies in install_requires of `setup.py` and should not provide a `requirements.txt` file. The latter is more intended for projects that are standalone applications.

soxofaan:
Example: when adding a dependency in requirements.txt recently, I forgot to also add it to setup.py, which broke a build of another project that depended on openeo-python-client. see ff71fbcfc5

soxofaan:
https://packaging.python.org/discussions/install-requires-vs-requirements/

soxofaan:
I think it's fine to keep `requirements-dev.txt` for installing the development related dependencies. There are ways to also handle that through setup.py instead (e.g. see https://stackoverflow.com/a/28842733) but these solutions are less well known

PR #99, opened 2019-11-19
Some fixes for pyton Sentinel example

ghtmtt:
Making some attempts with the `openEO_Python_Sentinel2_EURAC.ipynb` files I noticed some errors in imports and other small stuff. This PR should fix the imports (when creating a clean py3 venv)


soxofaan:
any reason you changed requirements.txt  that much? The set of dependencies of the python client should be as small as possible.

note that we are actually considering to drop requirements.txt: #74 

I think it would be better to have a requirements.txt file in the examples forlder, dedicated to the example notebooks

ghtmtt:
> any reason you changed requirements.txt that much? The set of dependencies of the python client should be as small as possible.

I just tried to use the example in a clean py env but I had to pip install all the stuff in the requirements file

> note that we are actually considering to drop requirements.txt: #74

I'll suggest to keep a setup method for new comers so that they can easily try the examples. But of course this is just a suggestion

> I think it would be better to have a requirements.txt file in the examples forlder, dedicated to the example notebooks

that could be also an alternative. Please let me know if I can be helpful in some way. 

Thanks!

soxofaan:
I added a requirements.txt file in the examples folder, as discussed above

examples/notebooks/openEO_Python_Sentinel2_EURAC.ipynb
```diff
@@ -126,7 +126,7 @@
     "from matplotlib.colors import NoNorm\n",
     "from matplotlib_scalebar.scalebar import ScaleBar\n",
     "import numpy as np\n",
-    "from osgeo import gdal\n",
+    "import gdal\n",
```
soxofaan: as far as I understand `import gdal` is deprecated and `from osgeo import gdal` should be used instead

soxofaan:
I also addressed the plot issue

I'll close this PR for now. Feel free to reopen if you want to further finetuning

PR #110, opened 2020-01-30
Allow specifying OAuth redirect port (and host)

soxofaan:
based on request of @bgoesswe to allow specifying the port for the OAuth redirect in the OIDC PKCE flow
e.g. like this:
```
con.authenticate_OIDC("openeo-public", timeout=3, server_address=('', 1234))
```


bgoesswe:
This works for me!

Issue #40, opened 2018-11-21
Dynamic Process function generation


bgoesswe:
Since the back ends may be capable of a different amount of processes and they can be retrieved by the GET /processes end point, it would be a major improvement to generate the process functions dynamically when a back end provider is chosen.  
e.g.: https://stackoverflow.com/questions/23812760/dynamic-functions-creation-from-json-python

It is at least something I want to look into.

soxofaan:
I think there are 2 separate aspects to this:

1. make sure all "official" processes from the openEO API are supported by the python client
2. detect that a backend does not support a process and fail early in the client instead of waiting for the backend to fail

Concerning 1.: I'm not a big of dynamic generation of functions/methods as this breaks some features that are important for the end user: normal discovery and documentation of methods (by looking at the source code of ImageCollectionClient, or using the code inspection features of their IDE), straightforward exceptions and backtraces when something goes wrong, lower barrier to entry to contribute/fix things, .... 

An alternative solution for 1. is still using traditional hardcoded methods but using unit tests that compare the openEO API description with the available methods of ImageCollectionClient and fail when something is missing. We are using this approach in the python driver to support dedicated python exceptions for each openEO error code (as defined in https://open-eo.github.io/openeo-api/errors/):
- we hardcoded an exception class for each error code in https://github.com/Open-EO/openeo-python-driver/blob/master/openeo_driver/errors.py 
- the unit tests in https://github.com/Open-EO/openeo-python-driver/blob/master/tests/test_errors.py  compare the exception classes with the expected openEO error codes

That being said, there are probably some ways to reduce the necessary boilerplate code and overhead to implement a process as a method in the client.


About 2.: this should be relatively straightforward to implement. It should be optional at the moment however, because probably not all backends properly declare which processes are supported in the capabilities endpoint (the VITO backend doesn't for example)

jdries:
There exists a way to add custom/unsupported processes:
https://open-eo.github.io/openeo-python-client/#openeo.rest.imagecollectionclient.ImageCollectionClient.graph_add_process

Perhaps we need to improve documentation so that people find it more easily?

About dynamically generated processes, I agree with Stefaan. I would however not object to someone showing how this can be done in the python client (as a separate way of building process graphs, separate from the ImageCollection class).

bgoesswe:
I'll investigate on possible dynamic generation strategies in the "process_generation" branch.

bgoesswe:
[...]

Issue #111, opened 2020-01-30
Clean up examples


soxofaan:
(This was brought up in the telco today)
The examples folder could use some clean up and finetuning, e.g. (but not limited):
- make sure the script work with latest version of client (and backends)
- make it more clear where to start, e.g.:
-- add a README.md 
-- reduce the number of examples in the root of the examples folder and move more advanced examples to subfolders
-- prefix the filenames with number 
- harmonize code style and terminology a bit better (e.g. #108, #105)

jonathom:
only saw this now, I'm on it :+1:

soxofaan:
This issue was brought to attention today: the examples in this repo are massively outdated.

Had a quick look and found various issues:
 
- usage of `con.capabilities().version()` which doesn't work anymore (was long deprecated and removed earlier this year)
- usage of `authenticate_basic()`, which is heavily deprecated since long
- usage of outdated functions/methods 
- outdated process graph building paradigms
- usage of boilerplate handling of things that are now built-in in the client
- sloppy/inconsistent coding style
- there is a `requirements.txt` file but it's completely unclear when/if it should be used

I think we should, at least, make it more clear that these examples are outdated by now, e.g. by putting them in an "archive" folder or something like that. We could pick a small subset to port to more recent standards, but maybe it's better to just point to the community examples repo. Also note that we have better maintained usage examples in the docs anyway.

soxofaan:
Already moved all examples to an "archive" subfolder and added some readme's to better describe the situation

PR #112, opened 2020-01-30
Process generation

bgoesswe:
Dynamic process generation. 
See #40 for more information.


openeo/internal/filter_bbox.json
```diff
@@ -0,0 +1,123 @@
+{
```
soxofaan: I think we should avoid mirroring json files from openeo-api and -processes

openeo/internal/processes.py
```diff
@@ -0,0 +1,902 @@
+def ln(cube, x):
```
soxofaan: I would also add some metadata in a comment in the header of this file with information like:
- which backend is this from
- at what time was it generated

openeo/internal/processes.py
```diff
@@ -0,0 +1,902 @@
...
+    """The natural logarithm is the logarithm to the base *e* of the number `x`. This
+process is an alias for the *log* process with the base set to *e*: `log(x,
```
soxofaan: FYI: the `textwrap` module has an `indent` function (https://docs.python.org/3/library/textwrap.html#textwrap.indent)  with which you can make the indentation of these docblocks nicer

soxofaan:
FYI: regarding the notes about the example scripts: we can delegate clean up also to #111

examples/eurac_example.py
```diff
@@ -1,43 +1,56 @@
 import openeo
 import logging
 import json
+from openeo.rest.imagecollectionclient import ImageCollectionClient
```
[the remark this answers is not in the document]
bgoesswe: fixed in the last commit

examples/eurac_example.py
```diff
@@ -1,43 +1,56 @@
...
 print(cap.list_plans())
 
-#Example using the 'ImageCollection' API.
-
-datacube = con.imagecollection("S2_L2A_T32TPS_20M")
-datacube = datacube.filter_bbox( west=652000, south=5181000, east=672000, north=5161000, crs="EPSG:32632")
-datacube = datacube.filter_daterange(extent=["2016-01-01T00:00:00Z", "2016-03-10T23:59:59Z"])
-datacube = datacube.ndvi( nir="B04", red="B8A")
+# load a specific dataset
+datacube = ImageCollectionClient.load_collection(session = con, collection_id = "S2_L2A_T32TPS_20M",
```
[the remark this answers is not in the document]
bgoesswe: fixed in the last commit

openeo/internal/filter_bbox.json
bgoesswe: Sry that was just for testing, so should have never bin in the repository:
Removed in the last commit

openeo/internal/processes.py
bgoesswe: Added a header text with the suggested metadata, also added path to the generation tool.

openeo/internal/processes.py
bgoesswe: Thanks! I used it to make the comments prettier.

soxofaan:
#153  introduces a 1.0-style variation of the process source code generator

I think we can close this PR #112

PR #113, opened 2020-02-04
updated EURAC example and Jupyter notebook, removed deprecated one.

bgoesswe:
- Updated EURAC example and Jupyter notebook, removed deprecated one. 
- Added operator skeleton ImageCollection base class, since some IDEs (Pycharm) mark there a warning, because they just know at runtime if the object has these methods.


soxofaan:
Hi @bgoesswe, sorry for the delayed review

Is it easy for you to separate the two things (example update and operator skeleton)? If not, I can cherrypick from your commit myself.

The thing is that we we might have to rethink operator handling a bit (especially on base class level), because it is not really well defined at the moment (e.g. see #123 ). For example, we might only define operators of specific subclasses.

bgoesswe:
Ok sure, I will remove the sceleton stuff.

soxofaan:
merged and squashed into master: d11543c7aa37023338c73eea07aa094164cc6ca7

PR #118, opened 2020-02-11
(WIP) Issue #40: add delegator for dynamically detected cube processes

soxofaan:
This is an experimental/proof of concept PR to address handling of dynamically detected backend processes

it adds a property `dynamic` to ImageCollectionClient instances,
which allows to call processes that are dynamically detected from the
backend process listing (and not necessarily predefined in the client)


bgoesswe:
I really like it, because the user does not have to do anything to use the existing processes. so it solves the dynamic usage of processes. It is just up for discussion if we want it locally for the convenience of code completion or not. For now, it is an uncomplicated solution for the users to use all available processes, which is nice.

Issue #125, opened 2020-03-11
openEO API 1.0.0 support


soxofaan:
openEO API 1.0.0 is in RC2 phase at time of this writing.
There is quite some work to support this fully in the python client

soxofaan:
pretty large refactor (targeting various issues and bugs) is going on under #117

soxofaan:
- [ ] context in reduce_dimension

soxofaan:
corresponding ticket for openeo python driver:  openeo-python-driver#33

soxofaan:
I think it's time to close this issue, as development focus has completely shifted to API 1.0 anyway

Issue #116, opened 2020-02-10
hardcoded dimension names


lforesta:
in openeo/rest/imagecollectionclient.py some dimension names are hardcoded

- _reduce_time() -> 'temporal'
- band() -> 'spectral_bands'

these clash with back-ends defining different names for these dimensions, e.g. see [https://earthengine.openeo.org/v0.4/collections/COPERNICUS/S2](https://earthengine.openeo.org/v0.4/collections/COPERNICUS/S2) under properties/cube:dimensions .

soxofaan:
related to #76 and #93

soxofaan:
merged in master

lforesta:
Can I ask how/where does the client get this information from the backend collections endpoint?
With which version(s) of STAC is the client compatible?

soxofaan:
The client tries to be bit forgiving about this and currently looks for dimension names here:
- `cube:dimensions` (stac 0.9.0 style per https://openeo.org/documentation/1.0/developers/api/reference.html#operation/describe-collection)
- `properties > cube:dimensions` (stac 0.6.2 per https://openeo.org/documentation/0.4/developers/api/reference.html#tag/EO-Data-Discovery/paths/~1collections~1{collection_id}/get

also see https://github.com/Open-EO/openeo-python-client/blob/master/openeo/metadata.py#L133-L204

PR #128, opened 2020-04-01
Fix gee example

bgoesswe:
- Fixed GEE example for the 1.0.0rc2 version of it. 

- Made some changes in the client to be compliant with 1.0.0rc2.

- Added the possibility to apply complex apply process graphs.

- Implemented multiple file result download.


openeo/imagecollection.py
```diff
@@ -52,7 +52,14 @@ def bands(self) -> List[Band]:
 
     def _get_bands(self) -> List[Band]:
         # TODO refactor this specification specific processing away in a subclass?
-        # First try `properties/eo:bands`
+        # First try `summaries/eo:bands`
+        eo_bands = self.get('summaries', 'eo:bands')
```
soxofaan, line 56: ah ok, I didn't know about this `summaries` field as replacement for the `properties` field.
Maybe we can drop support for `properties` in favor of `summaries`. I have no idea if it is necessary to keep supporting that

openeo/rest/connection.py
```diff
@@ -422,7 +426,10 @@ def download(self, graph, outputfile):
         :param format_options: formating options
         :return: job_id: String
         """
-        request = {"process_graph": graph}
+        if self._api_version.at_least("1.0.0"):
+            request = {"process": {"process_graph": graph}}
+        else:
+            request = {"process_graph": graph}
```
soxofaan, line 432: reference for self: Open-EO/openeo-python-driver#33

openeo/rest/connection.py
```diff
@@ -456,17 +463,22 @@ def create_job(self, process_graph: Dict, title: str = None, description: str =
...
-            "process_graph": process_graph,
+        pg_meta = {
             "title": title,
             "description": description,
             "plan": plan,
             "budget": budget
         }
+
+        if self._api_version.at_least("1.0.0"):
+            pg_meta["process"] = {"process_graph": process_graph}
```
soxofaan, line 474: reference for self:  Open-EO/openeo-python-driver#33

openeo/rest/datacube.py
```diff
@@ -523,34 +523,48 @@ def reduce_tiles_over_time(self, code: str, runtime="Python", version="latest"):
...
+                arguments["process"] = {"process_graph": PGNode(
+                        process_id=process,
+                        arguments={data_argument: {"from_parameter": "x"}})}
+            else:
+                arguments["process"] = {"process_graph": process}
+        else:
+            if isinstance(process, str):
+                arguments["process_graph"] = PGNode(
+                        process_id=process,
+                        arguments={data_argument: {"from_parameter": "x"}})
```
soxofaan, line 542: for 0.4 it has to be `from_argument` I think

openeo/rest/datacube.py
```diff
@@ -523,34 +523,48 @@ def reduce_tiles_over_time(self, code: str, runtime="Python", version="latest"):
...
-                    process_id=process,
-                    arguments={data_argument: {"from_parameter": "x"}}
-                )},
+            arguments=arguments
                 # TODO #125 context
-            }
         ))
 
-    def reduce_temporal_simple(self, process_id="max") -> 'DataCube':
+    def reduce_temporal_simple(self, process_id="max", dim_abbr="temporal") -> 'DataCube':
```
soxofaan, line 552: I would use `dimension` instead of `dim_abbr`,  as used in other places in DataCube methods

also, picking the correct dimension should be done automatically instead of manually by user: see #116. Can you put a comment `# TODO #116 determine dimension based on datacube metadata` here as well?

openeo/rest/datacube.py
```diff
@@ -523,34 +523,48 @@ def reduce_tiles_over_time(self, code: str, runtime="Python", version="latest"):
...
         """Finds the minimum value of a time series for all bands of the input dataset.
 
+            :param dim_abbr: Dimension name to reduce to.
+
             :return: a DataCube instance
         """
-
-        return self.reduce_temporal_simple("min")
+        # TODO: maybe find a better solution than dim_abbr (atm: time dimension: GEE: "t", VITO: "temporal")
+        return self.reduce_temporal_simple("min", dim_abbr=dim_abbr)
```
soxofaan, line 567: see higher about `dimension` vs `dim_abbr`

also add comment `# TODO #116 determine dimension based on datacube metadata` here please (I'm planning to work on that somewhere this week)

openeo/rest/datacube.py
```diff
@@ -711,6 +725,10 @@ def merge(self, other: 'DataCube', overlap_resolver: PGNode = None) -> 'DataCube
             'cube2': {'from_node': other._pg},
         }
         if overlap_resolver:
+            # TODO: for 1.0.0 support
```
soxofaan, line 728: Can you add `#125`  in this TODO line? makes it easier to find 1.0 related TODOs

soxofaan: or wait: as far as I understand the spec, there is no need for additional `process` level here. I think this is correct:
```
...
    "overlap_resolver": {
        "process_graph": {
            "max": { "process_id": "max" ....
```

openeo/rest/datacube.py
```diff
@@ -795,15 +813,24 @@ def _polygonal_timeseries(self, polygon: Union[Polygon, MultiPolygon, str], func
                 }
             }
 
+        reducer = {"process_graph": PGNode(
+                    process_id=func,
+                    arguments={"data": {"from_parameter": "data"}}
+                )}
+        # TODO: Might be necessary for 1.0.0rc2
```
soxofaan, line 820: same as above: from my understanding of spec this additional "process" level is not necessary

openeo/rest/job.py
```diff
@@ -79,24 +80,61 @@ def list_results(self, type=None):
         # GET /jobs/{job_id}/results
         pass
 
-    def download_results(self, target: Union[str, pathlib.Path]) -> pathlib.Path:
-        """ Download job results."""
+    def download_results(self, target: Union[str, pathlib.Path] = None) -> list:
+        """
+        Download job results to the target folder path.
```
soxofaan, line 85: This feels like a confusing API (especially the interplay between `target` and how many files there are in the result)

maybe we should provide instead:
- `download_result` to download a single result: fails when target is a folder or there are multiple results
- `download_results` is more generic: target should always be a folder

openeo/rest/job.py
```diff
@@ -79,24 +80,61 @@ def list_results(self, type=None):
...
-        with target.open('wb') as handle:
-            response = self.connection.get(file_url, stream=True)
-            for block in response.iter_content(1024):
-                if not block:
-                    break
-                handle.write(block)
-        return target
+
+        file_list = []
+
```
soxofaan, line 100: put a line  `target = Path(target or Path.cwd())` here, then you don't have to do `if target` anymore in the code below

openeo/rest/job.py
```diff
@@ -79,24 +80,61 @@ def list_results(self, type=None):
...
-        target = pathlib.Path(target)
-        with target.open('wb') as handle:
-            response = self.connection.get(file_url, stream=True)
-            for block in response.iter_content(1024):
-                if not block:
-                    break
-                handle.write(block)
-        return target
+
+        file_list = []
```
soxofaan, line 99: isn't it more logical to use a "to download" dict (mapping hrefs to local paths) instead of a list `file_list`?

soxofaan:
to address all "process"/"process_graph" changes I created dedicated ticket #129

openeo/imagecollection.py
bgoesswe: Yes, also "cube:dimensions" is now one layer up in the dictionary, I updated that in my last commit.

openeo/rest/datacube.py
bgoesswe: Yes true, done in latest commit

openeo/rest/datacube.py
bgoesswe: Yes you are right, I removed the comments

openeo/rest/job.py
bgoesswe: Now there are these two functions in the job class.

soxofaan:
Inspired by this PR, I worked out the "process"/"process_graph"  issue a bit more in PR #131 
Unfortunately this will conflict with this (PR #128 )

If you want @bgoesswe , I can merge #131  and  rebase #128   to get a view on the remaining issues

bgoesswe:
Yes sure that is a good idea.

soxofaan:
FYI: I just merged a rework of this PR in master

soxofaan:
Hi @bgoesswe   I think I merged now all your changes (with some finetuning here and there) into master (including commit `added TUW-UseCase to examples`).
Can you check if the your use cases work still on master as expected?

Because of the finetuning, this PR #128 is conflicting, and I think it is best to close it.

If you have more additions it is probably best to start with a new branch starting from master

soxofaan:
another thing (one of the finetunings I did), is simplification of this pattern:
```
job = datacube.send_job()
print(job.job_id)
print(job.start_job())
print(job.describe_job())
time.sleep(30)
print(job.download_results("/tmp/"))
```

you can now just do 
```
job = datacube.send_job()
res = job.start_and_wait().download_results("/tmp")
print(res)
```

instead of a hardcoded sleep, it will do a poll loop until job is finished

bgoesswe:
Yes works fine, thanks !

Issue #127, opened 2020-03-20
support for /process_graphs API


soxofaan:
add support for managing "user defined processes" with `/process_graphs` endpoints

related: 
- https://github.com/Open-EO/openeo-api/pull/261

PR #145, opened 2020-06-26
WIP: user-defined processes API

bossie:
#127

Connection supports a save_process_graph() method to persist a process graph and its metadata in the back end; it returns a (REST)ProcessGraph handle for future operations (update/delete/fetch details/validate).

DataCube also supports a save_process_graph() method to persist its process graph + metadata without evaluating it.

Not sure on the naming of the (REST)ProcessGraph handle and the way metadata is handled (it's a kwargs), hence the review.


openeo/rest/connection.py
```diff
@@ -525,6 +526,36 @@ def list_jobs(self) -> dict:
         # TODO: duplication with `user_jobs()` method
         return self.get('/jobs').json()["jobs"]
 
+    def save_process_graph(self, process_graph_id: str, process_graph: dict, **metadata) -> RESTProcessGraph:
```
soxofaan, line 529: I would use `user_defined_process` in these method names instead of `process_graph` to be closer to the terminology of the official docs.

Also, a user defined process is a bit more than a process graph, most importantly a declaration of the expected parameters. I would add an explicit argument `parameters` for that.

openeo/udp.py
```diff
@@ -0,0 +1,34 @@
+from abc import ABC, abstractmethod
+
+
+class ProcessGraph(ABC):
```
soxofaan, line 4: Are we going to use this base class in other places than only for `RESTProcessGraph`? If not, it seems unnecessary abstraction and duplication to maintain (e.g. doc strings)

openeo/rest/udp.py
```diff
@@ -0,0 +1,25 @@
...
+        return self._connection.save_process_graph(self.process_graph_id, process_graph, **metadata)
+
+    def describe(self) -> dict:
+        return self._connection.get(path="/process_graphs/{}".format(self.process_graph_id)).json()
+
+    def delete(self) -> None:
+        raise NotImplementedError
+
+    def validate(self) -> None:
+        raise NotImplementedError
```
soxofaan, line 25: As noted above: you could define a simple `Parameter` class here to simplify building parameter definitions and to allow a type hint  for `save_process_graph(..., parameters: List[Parameter]`...)

PR #146, opened 2020-07-09
Feature/ep 3389

jdries:


jdries:
@laxiwuka 
I think this is a good first step. We don't have actual interpretation of the result yet, like directly returning a parsed dataframe or numpy array, but that also could be a bit more tricky.
@soxofaan  Maybe you want to review as well? Otherwise this can be merged.

openeo/rest/connection.py
```diff
@@ -677,16 +677,20 @@ def _build_request_with_process_graph(self, process_graph: dict, **kwargs) -> di
     # TODO: Maybe rename to execute and merge with execute().
     def download(self, graph: dict, outputfile):
```
soxofaan, line 678: if `outputfile` can be string/Path or `None` I would add typehint and default here:
`outputfile: Union[Path, str, None] = None`

openeo/rest/datacube.py
```diff
@@ -990,7 +990,7 @@ def save_result(self, format: str = "GTIFF", options: dict = None):
             }
         )
 
-    def download(self, outputfile: str, format: str = "GTIFF", options: dict = None):
+    def download(self, outputfile: str = None, format: str = "GTIFF", options: dict = None):
```
soxofaan, line 993: change typehint here as well to `outputfile: Union[str, pathlib.Path]`

openeo/result.py
```diff
@@ -0,0 +1,76 @@
+from pathlib import Path
+from typing import Union, Dict
+
+from openeo.rest import OpenEoClientException
+from openeo.util import ensure_dir
+
+
+class Result:
+    def __init__(self, job):
```
soxofaan, line 9: I think it would be handy if this `Result` class can also be used for synchronous `executes` (so without a `job` instance). But maybe that is for another iteration.

openeo/result.py
```diff
@@ -0,0 +1,76 @@
...
+        assets = self._download_get_assets()
+        if len(assets) != 1:
+            raise OpenEoClientException(
+                "Expected one result file to download, but got {c}: {u!r}".format(c=len(assets), u=assets))
+        filename, metadata = next(iter(assets.items()))
+        url = metadata["href"]
+
+        return self.job.connection.get(url, stream=True)
+
+    def _download_get_assets(self) -> Dict[str, dict]:
```
soxofaan, line 68: minor: you can drop that `_download` prefix in the function name.
I used that under the `RESTJob` class to make clear that was about downloads, but here it is bit redundant

openeo/result.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def load_bytes(self):
+        return self._get_response_object().content
+
+    def _get_response_object(self):
+        assets = self._download_get_assets()
+        if len(assets) != 1:
+            raise OpenEoClientException(
+                "Expected one result file to download, but got {c}: {u!r}".format(c=len(assets), u=assets))
+        filename, metadata = next(iter(assets.items()))
```
soxofaan, line 63: FYI: in python you can also get the key-value tuple of a dict with one item with `k,v = d.popitem()`

openeo/result.py
```diff
@@ -0,0 +1,76 @@
...
+                handle.write(block)
+        return path
+
+    def load_json(self) -> dict:
+        return self._get_response_object().json()
+
+    def load_bytes(self):
+        return self._get_response_object().content
+
+    def _get_response_object(self):
```
soxofaan, line 58: maybe add retrun type hint for this method: `-> requests.Response`

openeo/result.py
```diff
@@ -0,0 +1,76 @@
...
+            for block in response.iter_content(1024):
+                if not block:
+                    break
+                handle.write(block)
+        return path
+
+    def load_json(self) -> dict:
+        return self._get_response_object().json()
+
+    def load_bytes(self):
```
soxofaan, line 55: add return type hint `bytes`

soxofaan:
@laxiwuka   I added a couple of minor review notes

openeo/result.py
```diff
@@ -0,0 +1,76 @@
+from pathlib import Path
```
soxofaan, line 1: I think this result.py file should be under `openeo.rest` instead of top level because it is rest specific

soxofaan:
thank, looks fine for merging to me

PR #149, opened 2020-08-26
EP-3521 propose 2 new ways of specifying callbacks

jdries:
This PR is about how we specify callbacks in the Python client. Untill now, we had:

- support for specific cases, like band math
- direct calls for process + callback, like reduce_temporal_udf, apply_dimension
- some use of internal API (PGNode) where users required some knowledge of openeo process specs

So basically not a generic solution that works in all cases.
This PR adds a processbuilder, which is similar to an ImageCollection/DataCube so you build a process graph by calling methods on it.
The difficult bit is how to link the result into the parent graph, with as little hassle as possible. This is context specific, as different processes use different parameter names etcetera.

Both of these solutions have been implemented, question is which one we prefer and want to push in our docs.

Solution 1: use a python function to define a callback (so yes, a real callback):

```
    collection = con100.load_collection("S2")

    def callback(data:ProcessBuilder):
        return data.absolute().run_udf(code='myfancycode', runtime='Python')

    neighbors = collection.apply_neighborhood(process=callback, size=[
        {'dimension': 'x', 'value': 128, 'unit': 'px'},
        {'dimension': 'y', 'value': 128, 'unit': 'px'}
    ], overlap=[
        {'dimension': 't', 'value': 'P10d'},
    ])
```

Solution 2: a builder pattern:
```
    callback = collection.apply_neighborhood( size=[
        {'dimension': 'x', 'value': 128, 'unit': 'px'},
        {'dimension': 'y', 'value': 128, 'unit': 'px'}
    ], overlap=[
        {'dimension': 't', 'value': 'P10d'},
    ])
    neighbors = callback.absolute().run_udf(code='myfancycode',runtime='Python').done()

```

So which solution is best?
- (1) maximally uses python language to specify an actual callback
- (2) uses a builder pattern, which could be well known among programmers, but not seen so often in Python API's
- my feeling is that (1) is a bit closer to the openEO API, so perhaps makes most sense?

### DIFF
--- a/docs/basics.rst
+++ b/docs/basics.rst
@@ -157,3 +157,30 @@ It can easily be converted into a pandas dataframe:
 
 The same method also works for multiple polygons, or GeoJSON or SHP files that are
 accessible by the backend. This allows computing aggregated values over very large areas.
+
+Some openEO process allow you to specify a process to be invoked on a subset of the datacube. This library allows this by specifying
+a 'callback' function.
+
+
+.. _callbackfunctions:
+
+Callback functions
+------------------
+
+A callback function is created by defining an actual Python function, or a lambda expression. This function is then passed on
+to processes such as :py:meth:`openeo.rest.datacube.DataCube.apply`, :py:meth:`openeo.rest.datacube.DataCube.apply_dimension`, :py:meth:`openeo.rest.datacube.DataCube.apply_neighborhood`,
+and :py:meth:`openeo.rest.datacube.DataCube.reduce_dimension`.
+
+This is an example:
+
+.. code-block:: python
+
+    datacube_absolute = cube.apply(process=lambda data:absolute(data))
+
+For more complex operations, you can define a function:
+
+.. code-block:: python
+
+   def callback(data):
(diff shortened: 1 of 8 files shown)


jdries:
Solution number 1 gets even better when you add lambda's. It allows us to write code like this:

`bandsum = collection.apply(process=lambda data:array_element(data,index=1) + array_element(data,index=2))`

This is very close to how openEO processes are defined and work, but also quite 'Pythonic' if you ask me.

jdries:
Just noticed that both the Javascript and R clients have a solution similar to number 1. That makes it an easy call, will merge.
In followup work, we'll need to add functions for all processes, possibly generating them.

bgoesswe:
Just to note: I m not sure if this still works, but some time ago I created a generation tool for the processes in 
https://github.com/Open-EO/openeo-python-client/pull/112

bgoesswe:
More specifically : https://github.com/Open-EO/openeo-python-client/blob/process_generation/openeo/internal/processes_gen.py

PR #152, opened 2020-09-10
EP-3521 fix parameter name "x" in "apply" callback

soxofaan:
follow-up of #149: @jdries   you changed a `from_parameter` in an `apply` callback from `x` to `data`:
https://github.com/Open-EO/openeo-python-client/blob/9e56beab772c3e91e51b905659cea17a6265a946/tests/data/1.0.0/apply_absolute.json#L10-L22

but from my understanding of the definition of the `apply` process (https://openeo.org/documentation/1.0/processes.html#apply),  the callback does get its input through parameter `x` (while the parent process indeed has the whole cube under `data`). So I think it should be like this:

      "apply1": {
        "process_id": "apply",
        "arguments": {
          "data": { "from_node": "loadcollection1"},
          "process": {
            "process_graph": {
              "absolute1": {
                "process_id": "absolute",
                "arguments": {
                  "x": { "from_parameter": "x"

@m-mohr   can you confirm which one is correct?

### DIFF
--- a/openeo/rest/datacube.py
+++ b/openeo/rest/datacube.py
@@ -683,7 +683,7 @@ class DataCube(ImageCollection):
             # Simple single string process specification
             process = PGNode(
                 process_id=process,
-                arguments={data_argument: {"from_parameter": "data"}}
+                arguments={data_argument: {"from_parameter": "x"}}
             )
         result_cube = self.process_with_node(PGNode(
             process_id='apply',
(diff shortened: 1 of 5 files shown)


m-mohr:
`x` is correct. data = array, x = single value.

soxofaan:
closing this one: fix was included in PR #153

PR #153, opened 2020-09-14
EP-3555 add missing processes for callbacks

soxofaan:
Follow up of EP-3521/#149: add missing processes for callbacks (EP-3555)

- adds a script `generator.py` to generate python source code that provides functions and methods corresponding to all processes from https://github.com/Open-EO/openeo-processes. (based on #112, which was still 0.4-style)
- unifies callback handling logic in `DataCube` methods that support i: `apply_dimension`, `reduce_dimension`, `apply_neighborhood`, `apply`, `merge_cubes`, `aggregate_spatial`,  `load_collection`
- unit tests


docs/processes.rst
```diff
@@ -240,3 +240,201 @@ The parameter listing of the example above could be written like this::
...
+You can also specify the callback as a "callable":
+a Python object that can be called (e.g. a function without parenthesis).
+
+The openEO Python Client Library defines the
+official processes in the :py:mod:`openeo.process.processes` module,
+which can be used directly:
+
+.. code-block:: python
+
+    from openeo.processes.processes import absolute, max
```
jdries, line 314: Can we avoid the double 'processes'? For instance by importing everything from processes.py in the __init__.py?

openeo/processes/generator.py
```diff
@@ -0,0 +1,163 @@
...
+import keyword
+import sys
+import textwrap
+from pathlib import Path
+from typing import Union, List, Iterator
+
+from openeo.processes.parse import Process, parse_all_from_dir
+
+
+class PythonRenderer:
```
jdries, line 11: Minor: this utility could be kept out of the source tree, to avoid packaging it in the wheel? Just to avoid confusing users.

jdries (changes requested):
Reviewed, main comment is about the import of processes!

docs/processes.rst
soxofaan: I will restructure code a bit to have imports like `from openeo.processes import max`

openeo/processes/generator.py
soxofaan: will move this to `openeo.internal`

soxofaan:
pushed extra commit addressing the comments (restructuring files)
let's see what the test runners think

soxofaan:
both Jenkins and Travis are happy
let's merge

PR #157, opened 2020-10-19
correct hand over of job options in send_job

claxn:
In `send_job`, job options like `title`, `budget`, `description`, ... where not used and send as `None` values to the back-end. This PR should fix this.


soxofaan:
I'll close this one as we merged #158  already

PR #163, opened 2020-12-04
small corrections

jonathom:
- grammatical error
- I think this was meant in the my_bandmath function, but I could be wrong
Don't know if master is correct to merge to, let me know.

### DIFF
--- a/docs/processes.rst
+++ b/docs/processes.rst
@@ -344,8 +344,8 @@ or normal Python functions:
     from openeo.processes import array_element
 
     def my_bandmath(data):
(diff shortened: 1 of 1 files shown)


jonathom:
I tested this on VITO and ran into an error since array elements are labeled __0__, 1, 2 and not 1, 2, 3. Is this backend-dependent or a general thing? One could also change this in the documentation then.

m-mohr:
The documentation of the process is pretty clear:

> The zero-based index of the element to retrieve.

soxofaan:
thanks for the fixes!

PR #167, opened 2020-12-28
Basic Usage Documentation

jonathom:
Minor adjustments so that the example now runs smoothly.

* the use of `filter_daterange` is deprecated and the recommended replacement `filter_temporal` doesn't seem to exist, so instead the parameter `temporal_extent` works as well
* the use of `crs="EPSG:4326"` leads to this error: `in spatial_extent_union raise ValueError("Different CRS's: {c}".format(c=crss)) ValueError: Different CRS's: {'EPSG:4326', 4326}`, so `crs=4326` is used instead

Furthermore I want to propose some ideas about this documentation, see below.


jonathom:
Proposed changes to this documentation:

1. Under "Managing jobs in openEO":
> Usually, you first need to get hold of your job, this can be done based on a job id:
>\>\>\> my_job = connection.job("0915ed2c-44a0-4519-8949-c58176ed2859")

Here, no example is given for obtaining the needed job id. How can this be done?

2. Under "Example: Retrieving aggregated timeseries":
> \>\>\> .polygonal_mean_timeseries(polygon)

is used. I needed some time to figure out how to pass `polygon`. The [documentation on the matter](https://open-eo.github.io/openeo-python-client/api.html#openeo.rest.datacube.DataCube.polygonal_mean_timeseries) didn't really help me. A geojson feature and also some solution using `shapely` seems to work. Will people just know how to pass polygons? I stumbled here, shortly. Maybe an addition to the linked API doc makes even more sense, plus a small comment here.

3. Also in this example, the last code block doesn't work right out of the box. I, as a beginner, did not know right away which imports to make:
```
import pandas as pd
import matplotlib
import matplotlib.pyplot as plt
```
and to also call `plt.show()` in the end. If this is something other people working in python will just know, than this can be ignored. Otherwise a complete example makes more sense to me.

If you think this is suitable, I can commit these additions to the pull request.

soxofaan:
Thanks for that fix. There is indeed more room for improvement, but I already merged your patch to get that out of the way

soxofaan:
building on your patch, I did a quick code (style) cleanup to improve consistency and readability.

soxofaan:
> filter_temporal doesn't seem to exist,

it does exist, but because of legacy reasons it's bit hidden in weird inheritance flow :smile:

soxofaan:
I addressed some of your other remarks too

> import matplotlib
> and to also call plt.show() in the end.

this depends quite a bit on the user workflow and environment, so I think it's better to leave this out here.


Feel free to add more suggestions, or close otherwise

jonathom:
Alright, thank you! Looks very nice to me :+1:

PR #170, opened 2021-01-20
Jupyter integration: UI components for data discovery

m-mohr:
This is the PR to show beautiful UI elements in Jupyter environments. (marketing mode off)

Currently, it supports Capabilties, Collections, Processes, File Formats, Service Types and UDF Runtimes. [I'll likely try to implement Logs (`GET .../logs`) next.](https://github.com/Open-EO/openeo-vue-components/issues/29)

This is still a bit experimental because I've re-implemented the technical basis to hopefully avoid issues between React and Vue, but [has another issue out of my scope](https://github.com/vuejs/vue-cli/issues/6225).

I'd appreciate a review from you guys as my Python programming skilly is still beginner-level. Please let me know anything that you think would be worth changing. Why does the build fail? I can't access your Jenkins...

Example:
image


jdries:
The code itself looks fine. I'll have a look at failing tests.

jdries:
@m-mohr  the problem seems to be that some methods return a List[dict], not a dict, for instance: list_processes, list_collections
So maybe this is a matter of creating also a VisualList?

m-mohr:
Ah, I see. That's a good point. I'll look into it.

m-mohr:
Is there a way to run the tests myself? I couldn't find details in the docs.

m-mohr:
@jdries I've added a VisualList. Multi inheritance doesn't seem to be the best entry point for a Python noob, but it seems to work... ;-)

Is there a way to run the tests locally? I couldn't find details in the docs.

soxofaan:
> Is there a way to run the tests locally? I couldn't find details in the docs.


to run the test you need the dev dependencies (from https://github.com/Open-EO/openeo-python-client/blob/master/.travis.yml):

    pip install .[dev]  --extra-index-url https://artifactory.vgt.vito.be/api/pypi/python-openeo/simple

and then you should be able to run the test suite with:

    pytest

We indeed should document this somewhere better

openeo/internal/jupyter.py
```diff
@@ -0,0 +1,70 @@
...
+# These classes are proxies to visualize openEO responses nicely in Jupyter
+# To show the actual list or dict in Jupyter, use repr() or print()
+
+SCRIPT_URL = 'https://cdn.jsdelivr.net/npm/@openeo/vue-components@2.0.0-beta.1/assets/openeo.js'
+
+class JupyterIntegration():
+
+	# Class instance to store whether the script has been inserted into the DOM
+	# TODO: This fails if a page is reloaded in Jupyter. The variable is still True, but the script doesn't exist in the page any longer.
+	script_sent = False
```
soxofaan, line 14: you seem to be using tabs here for indentation. The general recommendation/consensus is to use 4 spaces (https://www.python.org/dev/peps/pep-0008/#indentation)

m-mohr:
@soxofaan Thanks. Tests succeed now and I've created an issue to document the test procedure. So from my side this can be reviewed and merged.

We'll likely need to update the script URL again in some days/weeks, but that's only until we left beta/rc version numbers in vue-components again.

openeo/internal/jupyter.py
```diff
@@ -0,0 +1,70 @@
+from string import Template
+from collections import Mapping
+import json
+
+# These classes are proxies to visualize openEO responses nicely in Jupyter
+# To show the actual list or dict in Jupyter, use repr() or print()
+
+SCRIPT_URL = 'https://cdn.jsdelivr.net/npm/@openeo/vue-components@2.0.0-beta.1/assets/openeo.js'
+
+class JupyterIntegration():
```
soxofaan, line 10: If your class doesn't have parents, you don't need the empty parentheses.

openeo/internal/jupyter.py
```diff
@@ -0,0 +1,70 @@
...
+
+	# The first entry of the dict is expected to always be the actual dict with the data
+	def __init__(self, component: str, parameters: dict = {}):
+		JupyterIntegration.__init__(self, component, parameters)
+
+		data = {}
+		if len(parameters) > 0:
+			i = next(iter(parameters))
+			if isinstance(parameters[i], Mapping):
+				data = parameters[i]
```
soxofaan, line 51: why are you only checking/using the first item of the `parameters` dictionary? Note that dictionary ordering is not guaranteed in older python versions

soxofaan: From the usage below I see you you pass always a single item dictionary to the constructor here, and you are ignoring its key (`i` above) in the end. And there also seems to be redundancy between this key and `component`.

soxofaan: I would also make what's called `parameters` here as first argument, because that is the main argument of the parent class (`dict` in this case)

openeo/internal/jupyter.py
```diff
@@ -0,0 +1,70 @@
...
+		</openeo-$component>
+		""")
+		return template.substitute(
+			script = script,
+			component = self.component,
+			props = json.dumps(self.parameters)
+		)
+
+
+class VisualDict(dict, JupyterIntegration):
```
soxofaan, line 41: While multiple inheritance is possible in Python it is generally recommended to avoid it as much as possible, because it's easier to get inheritance chains wrong than to get it right. I'm not sure yet what other alternative you could use here though.

openeo/rest/connection.py
```diff
@@ -575,7 +589,8 @@ def list_services(self) -> dict:
         :return: data_dict: Dict All available service types
         """
         # TODO return parsed service objects
-        return self.get('/services').json()
+        data = self.get('/services').json()
+        return data
```
soxofaan, line 593: is this preparation, or you forgot VisualList?

openeo/internal/jupyter.py
m-mohr: > why are you only checking/using the first item of the `parameters` dictionary?

The parameters are the parameters that are passed to the Vue components (e.g. parameters `capabilities` and `url` in Capabilities component). To figure out which is the actual dict or list to proxy in the VisualDict/List, I need to have a convention which parameter it is (the name is always different). Thus I thought the convention is it's always the first...

> Note that dictionary ordering is not guaranteed in older python versions

From what I found on the web, order is guaranteed for Python >= 3.6 and the client readme says "Requirement: a Python 3.6 (or higher) environment" so I think it's fine?!

> From the usage below I see you you pass always a single item dictionary

No, there are cases with more than one (Capabilities) and there will be more in the future.

> and you are ignoring its key (i above) in the end

No, it's used in the json.dumps, which transmits the keys and values to the component, which is important.

> I would also make what's called parameters here as first argument, 

I need to know the key to pass it to the Vue component. So either we go `__init__(component : str, key : str, data : dict/list, parameters :dict = {})` then or leave it as it is.

openeo/internal/jupyter.py
m-mohr: Not sure what to do here?!

openeo/rest/connection.py
m-mohr: Could be preparation for the future, but to be honest, it was just a too greedy me copy/pasting things... reverted.

openeo/internal/jupyter.py
m-mohr: Okay, removed.

openeo/internal/jupyter.py
m-mohr: Oh, okay. My editor is defaulting to tabs for new files, it seems. Changed.

openeo/internal/jupyter.py
```diff
@@ -0,0 +1,70 @@
...
+	script_sent = False
+
+	def __init__(self, component: str, parameters: dict = {}):
+		self.component = component
+		self.parameters = parameters
+
+	def _repr_html_(self):
+		# Output script tag only once, otherwise JS throws errors
+		script = ''
+		if JupyterIntegration.script_sent == False:
```
soxofaan: Can't this check whether the script has already been included already be moved to JS space?
The JS frontend and Python backend work separately and I think you can get into situation where Python space `script_sent`  is out of sync with the reality of the JS-app side. For example: if you already included javascript in the JS app this way, and then restart your python kernel, the python space `script_sent`  would be reset to False, which is wrong.

openeo/internal/jupyter.py
m-mohr: I also experimented with **kwargs, but can't remember why this did not work... 🤔

openeo/internal/jupyter.py
```diff
@@ -0,0 +1,70 @@
...
+			script = '<script src="{}"></script>'.format(SCRIPT_URL)
+			JupyterIntegration.script_sent = True
+
+		# Construct HTML
+		template = Template("""
```
soxofaan: FYI: usage of this `string.Template` pattern is a bit overkill here. The standard `str.format` with named `{}` placeholders is powerful enough for what you want to do here, e.g.

    return """
        {script}
        <openeo-{component}> .... {props} ... </openeo-{component}>
    """.format(script= ..., component=....)

openeo/internal/jupyter.py
m-mohr: I'd love to, but I'm not sure how I can access the Jupyter DOM from the Python client. There are definitely cases where this could fail as documented in the code, especially a page reload. Kernel restart is less of a problem as it still works although some errors are logged to the JS console. With page reload it doesn't display anything if the JS is missing, although I could try to at least print a message that says what to do.

openeo/internal/jupyter.py
m-mohr: Oh, great. Wasn't aware of that. Fixed.

openeo/internal/jupyter.py
m-mohr: Maybe I can somehow build a JS wrapper script that is executed each time...

openeo/internal/jupyter.py
soxofaan: I think item order it's only guaranteed from py 3.7 and just a undocumented side-effect of the implementation in 3.6. But regardless of that, it looks bad style to depend on dict item order to my python eyes.

Wouldn't it be easier to just do something like this:
```python

class JupyterIntegration:
    def __init__(self, component, data, **kwargs):
        self.component = component
        self.data = data
        self.parameters = kwargs

    def _repr_html_(self):
        ...
        return " ....".format(
             ....
             props = json.dump({"data": self.data, "parameters": self.parameters})

class VisualDict(dict, JupyterIntegration):
    def __init__(self, data, component, **kwargs):
        dict.__init__(data)
        JupyterIntegration.__init__(component=component, data=data, **kwargs)

```

m-mohr: As I said, "data" is not a fixed term. It can be "formats", "collections", ... so I need to pass the key name, too.

openeo/internal/jupyter.py
soxofaan: yeah, I don't have a better alternative ready. So let's keep it as is for now

openeo/internal/jupyter.py
m-mohr: No, there's not. Either I need to provide it in the constructor or define a mapping with a couple of if/elifs.

openeo/internal/jupyter.py
m-mohr: @soxofaan 


Do you have a clue why this 
```py

        return """
        <script type="text/javascript">
        if (!window.customElements.get('openeo-{component}')) {
            var el = document.createElement('script');
            el.src = "{script}";
            document.appendChild(el);
        }
        </script>
        <openeo-{component}>
            <script type="application/json">{props}</script>
        </openeo-{component}>
        """.format(
            script = SCRIPT_URL,
            component = self.component,
            props = json.dumps(self.parameters)
        )
```
gives a "Invalid format string" error? If I remove the "{script}" it works though, but how to fill it then?

soxofaan: because it trips over the curly braces of the `if () {... } `  

you can escape these with double braces: 
    
    >>> "{x} {{y}}".format(x="foo", y="bar")
    'foo {y}'


But if you indeed want to use more advanced javascript in that string, it might be easier to return to that `string.Template` :rofl:

soxofaan: > how I can access the Jupyter DOM from the Python client. 

> a JS wrapper script that is executed each time..

yes, I meant something like a quick JS-side checking wrapper

openeo/internal/jupyter.py
soxofaan: Can't you move that logic to the JS side? then you have a smaller API surface between python and JS: just `component` instead of `component` + `key`?

m-mohr: No, unfortunately not. Actually, this is HTML here.

openeo/internal/jupyter.py
m-mohr: Wrote a small JS script that checks the existence of the required HTML tags and loads the script only if it's missing. It seems to work in all cases I tested.

openeo/internal/jupyter.py
m-mohr: I've moved the "key" handling into the Jupyter code now, it at least doesn't show up in the "normal" functions now... Is that okay? @soxofaan

m-mohr:
> but [has another issue out of my scope](https://github.com/vuejs/vue-cli/issues/6225).

I've implemented a workaround. This should be ready to go (but it seems TravisCI is acting up - I guess we need to migrate openEO CI at some point to GH Actions...)

openeo/internal/jupyter.py
```diff
@@ -0,0 +1,61 @@
+from collections import Mapping
+import json
+
+# These classes are proxies to visualize openEO responses nicely in Jupyter
+# To show the actual list or dict in Jupyter, use repr() or print()
+
+SCRIPT_URL = 'https://cdn.jsdelivr.net/npm/@openeo/vue-components@2.0.0-rc.1/assets/openeo.js'
```
soxofaan, line 7: maybe add a todo note here

     TODO: get this script url dynamically e.g. from a https://editor.openeo.org/ quert or from a local config file, with hardcoded fallback?

m-mohr: I'm not sure. We could add it to a config file, but we can't and shouldn't get it through other means dynamically. The implementation is bound to the specific version and once we have a stable version 2.0.0 (in one or two weeks) it will be `SCRIPT_URL = 'https://cdn.jsdelivr.net/npm/@openeo/vue-components@2/assets/openeo.js'`, which will cover always the latest version of the 2.x branch and then we won't change it any time soon again as 3.x would be breaking anyway.

m-mohr: We can keep this PR open until we have a 2.0.0 release.

openeo/metadata.py
```diff
@@ -392,3 +393,6 @@ def add_dimension(self, name: str, label: Union[str, float], type: str = None) -
         else:
             dim = Dimension(type=type or "other", name=name)
         return self._clone_and_update(dimensions=self._dimensions + [dim])
+    
+    def _repr_html_(self):
+        return VisualDict('collection', data = self._orig_metadata)._repr_html_()
```
soxofaan, line 398: I think it's cleaner to skip the `VisualDict` layer and just do `JupyterIntegration(...)._repr_html_()` directly.

Or one step further: move the logic of `JupiterIntegration._repr_html_` out of a class and make it a simple function e.g. `openeo_repr_html(component, data, parameters)`, which you call here directly, Then you decouple the jupyter-specific rendering from the object building, which might help with other kind of jupyter integrations later.

openeo/rest/rest_capabilities.py
```diff
@@ -32,3 +34,6 @@ def currency(self):
     def list_plans(self):
         """ List all billing plans."""
         return self.capabilities.get('billing', {}).get('plans')
+
+    def _repr_html_(self):
+        return VisualDict("capabilities", data = self.capabilities, parameters = {"url": self.url})._repr_html_()
```
soxofaan, line 39: see my comment about `MetadataException`

openeo/metadata.py
m-mohr: I did the first suggestion. The latter is more complicated due to the steps in the constructor.

soxofaan: an another step further: this might even eliminate the need for the `JupiterIntegration` layer and the related multiple inheritance. 

e.g. VisualDict could become something like this:

```python
class VisualDict(dict):
    def __init__(self, component: str, data : dict):
        super().__init__(data)
        self.component = component
   def _repr_html(self):
        return openeo_repr_html(component=self.component, data=self)
```

m-mohr: That doesn't fully capture it. The parameter mapping is not included and the additional parameters neither. I mean I can just duplicate the JupyterIntegration code in __init__, of course... is that desirable?

soxofaan: > the additional parameters neither

that's intentional: you are not using parameters in the places where you would be using this new VisualDict or VisualList

> parameter mapping is not included

you can still do that inside that openeo_repr_html function, I don't think it is necessary to do that in the "init" phase

m-mohr: Not yet, but it will come up in the future for sure

m-mohr: I've modified the code to be similar to your proposal.

openeo/internal/jupyter.py
jdries: Or merge now? This will not be released immediately anyway, but it will make it available to other people testing on master?

m-mohr: @jdries Sure, I'm fine with that. Just released another RC for vue-components so that all recent changes are in.

m-mohr: > This will not be released immediately anyway

@jdries But five hours later 😂 👍

PR #179, opened 2021-02-18
Further UI components

m-mohr:
- Add UI component for Batch Job Results and tables for jobs, services and files


m-mohr:
I've also added tables for list_services, list_jobs and list_files.

Example:
image

m-mohr:
@soxofaan Do you have an idea why this happens?

image

m-mohr:
Additionally implemented functionality that allows iterating through collection items (e.g. pre-computed ARD).

@soxofaan Could you have a first look at whether this is all good from a Python/library perspective or whether you'd like me to change things? The SCRIPT_URL will get valid in the next days. 

Example:
image

m-mohr:
We would also need to download assets from those API endpoints, but I think I'd aim for a more generic implementation that can also be re-used in downloadResults etc. - see #184

openeo/rest/connection.py
```diff
@@ -588,6 +617,30 @@ def describe_collection(self, name) -> dict:
         data = self.get('/collections/{}'.format(name)).json()
         return VisualDict("collection", data = data)
 
+    def collection_items(self, name, spatial_extent: Optional[Dict[str, float]] = None, temporal_extent: Optional[List[Union[str, datetime.datetime]]] = None) -> Paginate:
+        """
+        Loads items for a specific image collection.
+        May not be available for all collections.
```
jdries, line 623: It's actually not available for most collections, should we add a clean error in that case?

m-mohr: I don't know, I think I followed the route of the rest of the python client where it does not do any "feature-checking" against e.g. the endpoints in "/". But happy to adopt whethever you prefer.

openeo/rest/connection.py
```diff
@@ -55,6 +55,34 @@ def __init__(self, http_status_code: int = None,
...
+        return self
+
+    def __next__(self):
+        if self.nextUrl is None:
+            raise StopIteration
+        
+        response = self.con.get(self.nextUrl).json()
+        if isinstance(response['links'], list):
+            nextLink = next((link for link in response['links'] if link['rel'] == 'next'), None)
+            self.nextUrl = nextLink['href']
```
soxofaan, line 79: will fail if nextLink is None (there was no next link)

openeo/rest/connection.py
```diff
@@ -55,6 +55,34 @@ def __init__(self, http_status_code: int = None,
         self.url = url
         super().__init__("[{s}] {c}: {m}".format(s=self.http_status_code, c=self.code, m=self.message))
 
+class Paginate:
```
soxofaan, line 58: I think this can be implemented more succinctly  with a generator (`yield`) instead of implementing the iterator protocol.

Something like this (untested):

```python

def paginate(con: "Connection", url: str, callback: Callback):
    page = 0
    while True:
        resp = con.get(url).json()
        yield callback(resp, page)
        next_links = [link for link in resp.get("links", []) if link["rel"] == "next] 
        if not next_links:
             break
        url = next_links[0]
        page += 1
        


```

openeo/rest/connection.py
```diff
@@ -588,6 +617,30 @@ def describe_collection(self, name) -> dict:
...
+        q = []
+        if spatial_extent is not None:
+            bbox = map(lambda c: str(c), spatial_extent)
+            q.append('bbox={}'.format(','.join(bbox)))
+        
+        if temporal_extent is not None:
+            datetimes = map(lambda t: '..' if t is None else rfc3339.normalize(t), temporal_extent)
+            q.append('datetime={}'.format('/'.join(datetimes)))
+        
+        url = '/collections/{cid}/items?{q}'.format(cid = name, q = '&'.join(q))
```
soxofaan, line 641: I think we can avoid all this query parameter serialization, the `requests` library will take care of that for you. e.g

```python
params = {}
if spatial_extent:
    params["bbox"] = ",".join(str(c) for c in spatial_extent)
if temporal_extent:
    params["datetime"] = "/".join(".." if t is None else rfc3339.normalize(t) for t in temporal_extent)

con.get(url, params=params)
```

soxofaan (commented):
added some initial review notes

openeo/rest/connection.py
```diff
@@ -588,6 +617,30 @@ def describe_collection(self, name) -> dict:
         data = self.get('/collections/{}'.format(name)).json()
         return VisualDict("collection", data = data)
 
+    def collection_items(self, name, spatial_extent: Optional[Dict[str, float]] = None, temporal_extent: Optional[List[Union[str, datetime.datetime]]] = None) -> Paginate:
```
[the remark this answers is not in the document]
m-mohr: What is the return type? I can't figure it out with my limited Python knowledge

openeo/rest/connection.py
```diff
@@ -55,6 +55,34 @@ def __init__(self, http_status_code: int = None,
         self.url = url
         super().__init__("[{s}] {c}: {m}".format(s=self.http_status_code, c=self.code, m=self.message))
 
+class Paginate:
+
+    def __init__(self, con, url: str = None, callback: Callable = None):
```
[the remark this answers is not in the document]
m-mohr: I'll remove the default for url and make it so that it doesn't require a callback (is there a better term in py?)

openeo/rest/connection.py
m-mohr: Does that return? I guess I need to learn what yield is before...

openeo/rest/connection.py
m-mohr: Great, didn't know...

openeo/rest/connection.py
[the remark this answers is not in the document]
m-mohr: How would the method be used then? Is it the same as before?

openeo/rest/connection.py
m-mohr: How would the method be used then? Is it the same as before?

soxofaan: yes both are basically equivalent,
 `yield` allows you to create an (lazy) iterator in a more to-the-point way than the iterator protocol (`__iter__` and `_next__`)

openeo/rest/connection.py
jdries: Good point ;-)
But you are allowed to do feature checking, or otherwise to catch exceptions. In this case, I guess the server will return a 404 if item metadata is not available, so we could catch that, and translate it into a message that explains to the user that the feature is optional.

openeo/rest/connection.py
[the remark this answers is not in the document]
soxofaan: In this case it will be an iterator of `VisualDict` if I understand correctly, so `typing.Iterator[VisualDict]`

openeo/rest/connection.py
m-mohr: Fixed

openeo/rest/connection.py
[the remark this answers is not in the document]
m-mohr: Took "Iterator" for now...

openeo/internal/jupyter.py
```diff
@@ -1,16 +1,89 @@
...
 
-SCRIPT_URL = 'https://cdn.jsdelivr.net/npm/@openeo/vue-components@2.0.0-rc.2/assets/openeo.min.js'
+SCRIPT_URL = 'https://cdn.jsdelivr.net/npm/@openeo/vue-components@2/assets/openeo.min.js'
 COMPONENT_MAP = {
+    'collection': 'data',
+    'data-table': 'data',
     'file-format': 'format',
     'file-formats': 'formats',
+    'item': 'data',
+    'items': 'items',
```
soxofaan, line 10: "item"/"items" sounds very generic, can't we make it a bit more concrete and self-describing, like "stac item" or "collection item"? Or is it really generic/abstract at this point?

m-mohr: This is a mapping of component and parameter for the Vue component. This directly translates into code, we can't change anything here.

openeo/rest/connection.py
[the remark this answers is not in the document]
m-mohr: With recent changes it will be more generally Iterator[dict], thanks.

openeo/rest/connection.py
m-mohr: The request is only done in the (hopefully generic) Paginate iterator so I can't really make that specific unfortunately. There's no request being sent in collection_items. So I could only check against the endpoints in /. But that is something that we could do for all endpoints in general, that doesn't seem too specific for this PR.

openeo/rest/connection.py
m-mohr: Implemented, works, thanks.

m-mohr:
Except for the "feature checking" (not sure yet how to handle it), I think I've addressed all feedback. Could you have a look again, please? @soxofaan

openeo/internal/jupyter.py
soxofaan: > No user will see it.

I'm not sure, if they want to work with `collection_items` results outside of a jupyter context, they;'ll will have to understand what  `VisualDict("items"` is about

openeo/rest/connection.py
```diff
@@ -55,6 +55,21 @@ def __init__(self, http_status_code: int = None,
         self.url = url
         super().__init__("[{s}] {c}: {m}".format(s=self.http_status_code, c=self.code, m=self.message))
 
+def paginate(con, url: str, params: dict = {}, callback: Callable = None):
+    page = 1
+    while True:
+        response = con.get(url, params = params).json()
+        if callback:
```
soxofaan, line 62: even easier than `if callback ... else` : give callback this default value: `lambda resp, page: resp`

The has added value that the `callback` argument is more self-documenting

openeo/rest/connection.py
```diff
@@ -55,6 +55,21 @@ def __init__(self, http_status_code: int = None,
...
+            yield callback(response, page)
+        else:
+            yield response
+        next_link = next((link for link in response['links'] if link['rel'] == 'next'), None)
+        if next_link is None:
+             break
+        url = next_link['href']
+        page += 1
+        params = {}
```
soxofaan, line 71: Are you intentionally clearing the `params`?  Won't that change the result set , and consequently the paging?

openeo/rest/job.py
```diff
@@ -264,6 +265,10 @@ def __init__(self, job: RESTJob):
     def __repr__(self):
         return "<JobResults for job {j!r}>".format(j=self._job.job_id)
 
+    def _repr_html_(self):
+        response = self.get_metadata()
```
soxofaan, line 269: This is tricky:  `get_metadata()` will (currently) fail when the job is not finished yet (`requests to this endpoint MUST be rejected with openEO error JobNotFinished`) , you don't want this behavious in `_repr_html_`

openeo/internal/jupyter.py
m-mohr: Maybe, but as I said, that's the API of the vue-components and there's nothing we can change about it (except adding a comment behind the line). On the other hand, VisualDict is just a dict?!

openeo/rest/connection.py
m-mohr: Yes, the params must be in the url we get back two lines above.

openeo/rest/connection.py
m-mohr: Sure, changed.

openeo/rest/job.py
m-mohr: So I guess I should just catch the error and print something else? Or what would be better?

soxofaan: I think I can improve the exception that will be thrown there, so that you can check HTTP status code and openeo Error code, then you can indeed render some kind of "job not finished yet" message

soxofaan: aha probably no change required: you should already get a `OpenEoApiError` exception there with `.http_status_code` >= 400 and `.code` == "JobNotFinished"

openeo/internal/jupyter.py
m-mohr: I just realized I can remove the line completely.

openeo/rest/job.py
m-mohr: I've tried it, but I can't import OpenEoApiError in Job, seems like a recursive dependency. How to solve that?

m-mohr: Shall I just check for OpenEoClientException instead?

m-mohr: ...but then I can't access the message

m-mohr: Implemented, that was more complicated than I thought it would be...

soxofaan:
Can you also add a CHANGELOG entry?

openeo/rest/job.py
```diff
@@ -264,6 +265,13 @@ def __init__(self, job: RESTJob):
     def __repr__(self):
         return "<JobResults for job {j!r}>".format(j=self._job.job_id)
 
+    def _repr_html_(self):
+        try:
+            response = self.get_metadata()
+            return render_component("batch-job-result", data = response)
+        except OpenEoApiError as error:
+            return render_error(error, "Job not finished yet")
```
soxofaan, line 273: There could also be other errors, I would only catch the `JobNotFinished` ones, e.g.:

```python
try:
    response = self.get_metadata()
except OpenEoApiError as error:
    if error.code == "JobNotFinished":
        return render_error(error, "Job not finished yet")
    raise
else:
    return render_component("batch-job-result", data = response)


(the empty `raise` reraises the existing exception)

Comment (m-mohr) [reply openeo/rest/job.py:273]: That is handled, usually, it should just show the message. The message I'm passing there is just a fallback (maybe we don't even need the fallback?)

Comment (m-mohr) [reply openeo/rest/job.py:273]: Yeah, I think I'll just remove the fallback.

**Comment by m-mohr** [comment]
I've added changelog entries.

File: CHANGELOG.md [hunk 12]
```diff
@@ -7,13 +7,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- `collection_items` to request collection items through a STAC API
```
soxofaan, line 10: hmm, maybe you should rebase on latest master (or merge master), there should be already "added" entries here

m-mohr: Done

openeo/rest/connection.py
```diff
@@ -588,6 +587,30 @@ def describe_collection(self, name) -> dict:
...
+        :param temporal_extent: Limits the items to the specified temporal interval.
+        :param limit: The amount of items per request/page. If None, the back-end decides.
+        The interval has to be specified as an array with exactly two elements (start, end).
+        Also supports open intervals by setting one of the boundaries to None, but never both.
+        :return: data_list: List A list of items
+        """
+        url = '/collections/{}/items'.format(name)
+        params = {}
+        if spatial_extent:
+            params["bbox"] = ",".join(str(c) for c in spatial_extent)
```
soxofaan, line 606: `spatial_extent` is typed as `Dict[str, float]` (so a bounding box dictionary I guess), but `str(c) for c in spatial_extent` will iterate over the  keys ("west", "north", "south", "east" in some undetermined order), which is not what you want I guess

openeo/rest/connection.py
```diff
@@ -41,20 +41,18 @@ def url_join(root_url: str, path: str):
...
-    def __init__(self, http_status_code: int = None,
-                 code: str = 'unknown', message: str = 'unknown error', id: str = None, url: str = None):
-        self.http_status_code = http_status_code
-        self.code = code
-        self.message = message
-        self.id = id
-        self.url = url
-        super().__init__("[{s}] {c}: {m}".format(s=self.http_status_code, c=self.code, m=self.message))
-
+def paginate(con, url: str, params: dict = {}, callback: Callable = lambda resp, page: resp):
```
soxofaan, line 44: do you have experience with unit tests and pytest @m-mohr ? I think this one deserves a bit of coverage.
If not, I can have a look after merging

openeo/rest/connection.py
m-mohr: I copy-pasted that from load_collection. I guess we can also just allow float here? Otherwise, same order as in load_collection. So if there's an easy way to handle that as in load_collection, we could also just do that. But I was a bit confused about how load_collection did it...

openeo/rest/connection.py
m-mohr: Unit Tests yes, pytest no.

openeo/rest/job.py
soxofaan: my point is that you should, I think, only catch the "JobNotFinished" error, and let other issues propagate normally

m-mohr: Why not show the others nicely? It's just errors that really come from the API, usually with a message, right? I'm not catching all errors, I think.

openeo/rest/connection.py
soxofaan: we use a http mock library (requests_mock) for fake url serving, so we don't need a real "online" endpoint

m-mohr: Good.
To get things in so that others can test things for the SRR, I'd rather postpone to later. If you don't have the time, I can give it a try, maybe we can just open an issue so that we don't forget.

openeo/rest/connection.py
soxofaan: in `load_collection` and `filter_bbox` we want a bbox dictionary in the end, 
but if I understand `/collections/{}/items` correctly, bbox should be an array of length `2*n`, which is totally different

m-mohr: Indeed. I committed a new version (list instead dict + removing str + document the order)

openeo/rest/job.py
soxofaan: ok, true. It's a bit a gray zone: the user does not call `_repr_html_` knowingly, so maybe it's fine to intercept exceptions a bit more greedily.

openeo/rest/connection.py
soxofaan: deserves some unit tests too, I'd say :)

m-mohr: https://github.com/Open-EO/openeo-python-client/issues/186

openeo/rest/connection.py
m-mohr: https://github.com/Open-EO/openeo-python-client/issues/186

soxofaan:
Overall, I think I nitpicked enough :stuck_out_tongue:  and we can wrap this up. Disclaimer: I didn't test any of this for real, I just reviewed the source code.

One general remark is that we are adding yet another method (`collection_items`) method on `Connection`. I think we should avoid putting "everything but the kitchen sink" in `Connection` as this can be annoying for documentation overview and long term maintenance. It's cleaner to split off subsystems to dedicated classes (like with batch job results). However, in the case of collections we don't have a dedicated class for that, so I don't have a good alternative for how we do it in the current PR. If you agree with this: we could put a note "this is experimental API and might be subject to change" in the doc block.

clausmichele:
@soxofaan when do you plan to merge this? It is necessary to have it to conclude the metadata part of the notebooks, let me know if I can test it somehow.

soxofaan:
I can merge today, if ok for @jdries

jdries:
Ok for merge, go ahead!

m-mohr:
@soxofaan Yes, I agree. The next step would be a `download_collection_items` or something similar and - at the moment - would also go into Connection. Thus having a collection class would really be a good idea. There's already the Collection Metadata class though?! I'll add the experimental text.

m-mohr:
Done, ready to be merged.

soxofaan:
@clausmichele  it has been merged

PR #188, opened 2021-03-09
band name aliases

jdries:


jdries:
In this proposal, which indeed seems better, would users of the old band names get an error? (Because they try to use a band name that is not in the metadata?)

soxofaan:
Ah ok, so it is actually also about client side resolving of band names.
So it is intentional to depend on a non-standard metadata field "internal_name"?
(because of the "internal" I assumed it was something internally for geopyspark driver).

jdries:
@soxofaan  Ok to merge? Seems fine for me, and I need the aliases in the backend.

soxofaan:
yes look a lot better and more generic now

might need some test coverage though (not necessarily merge blocker)

jdries:
Ok, will merge already!

PR #202, opened 2021-04-19
Remove non-standard `srs` argument from `mask_polygon`

soxofaan:
geojson as used in `mask_polygon` is EPSG:4326/WGS84 only and openEO API currently does not provide way to extend this.

Current implementation in DataCube (and ImageCollectionClient) provides an additional `srs` argument to specify an alternative reference system, which is non-standard, not supported by a backend as far as I know .
Also the coverage of `srs` is incomplete: only the code path that handles a shapely geometry uses `srs`.
Other code paths do not. While it could be possible to add it to the `read_vector` and `dict` code paths, it is not possible for the `Parameter` code path without extending the API

I would just remove the `srs` argument.

(related EP-3747, Open-EO/openeo-processes#53)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Removed
 
+- Remove non-standard `srs` argument from `DataCube.mask_polygon` hmethod (EP-3747 related)
+
 ### Fixed
 
 - `Connection`: don't send default auth headers to non-backend domains ([#201](https://github.com/Open-EO/openeo-python-client/issues/201))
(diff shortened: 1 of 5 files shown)


soxofaan:
an alternative could be to let the client do the reprojection, but that still has problems: that would introduce new dependency on (e.g.) `pyproj` which is a non-trivial package to install . The `Parameter` use case would also still be a problem.

soxofaan:
Apparently this geojson extension (from 2008 spec, e.g. `"crs":{"type":"name","properties":{"name": ...`) is being depended on by some (internal) use cases, so we can not just drop it. On the contrary, it would be better for usability to standardize it in some way ideally at level of openEO API itself.

anyway closing this PR

soxofaan:
also see discussion about how to tackle this (ideally) at openEO API level: Open-EO/openeo-processes#235

Issue #115, opened 2020-02-10
Clean up deprecated/outdated/unused modules and classes


soxofaan:
While going through code and issues related to the 1.0 version of the API, I found some top-level modules/classes in the openeo python client that do not seem to be used (anymore?). It think we should remove these to streamline the code structure.

Some examples:

- `openeo/service.py`, last edit 15 months ago by @bgoesswe , is an abstract base class, but my IDE can't find concrete subclasses
- `openeo/processgraph.py` and `openeo/rest/rest_processgraph.py`: last edit 15 months ago by @bgoesswe  (except for minor import tweak by myself), my editor can't find usage of these classes
- `openeo/error_summary.py`: is only used in openeo-python-driver and openeo-geopyspark-driver: I think this should be moved to openeo-python-driver
- `openeo/catalog.py` with class `EOProduct`: is abstract base class, but I can't find concrete usage of it, last edit oct 2017 by @jdries

bgoesswe:
`openeo/processgraph.py`:
I think we can remove this, I just added them, because I felt like it is more compliant with the client development guidelines (back then). But I think a Python dictionary is good enough to represent a process graph. (maybe keeping [this API section](https://openeo.org/documentation/1.0/developers/api/reference.html#section/Processes/Defining-Processes) in mind).
`openeo/service.py`:
This was thought to be a class like jobs, but for services, so this might be interesting in the future (since it could handle [these endpoints](https://openeo.org/documentation/1.0/developers/api/reference.html#tag/Secondary-Services)), so if we want to return a list of Job objects with connection.list_jobs (#43), we could also do that for services. But having a separate class for that might be up for discussion.

m-mohr:
Potential implementation for service.py is in PR #206

soxofaan:
just removed some old proof of concept leftovers around file API. Implementation was 0.4-only, didn't work, and the interface was not ideal (returned http status codes instead of throwing python exceptions). Better just start over fresh when we need file API support (#122 )

soxofaan:
I think it's time to close this ticket now, most outdated/unused modules are cleaned up now

PR #206, opened 2021-05-04
More Jupyter visualizations + Services

m-mohr:
Adds support for additional Jupyter visualizations:

- UDP list
- UDP details
- Job details
- Job estimates
- Job logs (need help here)
- Show capabilities for Connection class

Not included, although available as components:
- Logs could also be made available for sync. processing, but seems to be unsupported by the Python client yet?
- Service details + logs could also be implemented, but seems to be unsupported by the Python client yet (service.py)?

I still need to deploy to npm, waiting to solve the remaining issues here (see comment) so that I can fully test the components locally afterward.

ToDos:
- [ ] Add CHANGELOG entries
- [x] Release Vue Components 2.1 and wait for CDN to update

### DIFF
--- a/openeo/api/logs.py
+++ b/openeo/api/logs.py
@@ -6,7 +6,7 @@ class LogEntry:
     """
 
     def __init__(self, entry):
-        self.log_id = entry['id']
+        self.id = entry['id']
         self.code = entry.get("code", 0)
         self.level = entry['level']
         self.message = entry['message']
@@ -15,6 +15,3 @@ class LogEntry:
         self.links = entry.get("links", [])
         if 'data' in entry:
             self.data = entry['data']
-
-    def toJson(self):
-        return json.dumps(self, default = lambda o: o.__dict__) # doesn't work
--- a/openeo/api/logs.py
+++ b/openeo/api/logs.py
@@ -1,0 +1,20 @@
+import json
+
+class LogEntry:
+    """
+    A log entry.
+    """
+
+    def __init__(self, entry):
+        self.log_id = entry['id']
+        self.code = entry.get("code", 0)
+        self.level = entry['level']
+        self.message = entry['message']
+        self.time =  entry.get("time", None) # todo: native date/time object?
+        self.path = entry.get("path", [])
+        self.links = entry.get("links", [])
+        if 'data' in entry:
+            self.data = entry['data']
+
+    def toJson(self):
+        return json.dumps(self, default = lambda o: o.__dict__) # doesn't work
--- a/openeo/api/logs.py
+++ b/openeo/api/logs.py
@@ -6,12 +6,37 @@ class LogEntry:
     """
 
     def __init__(self, entry):
+        # Unique ID for the log, string, REQUIRED
         self.id = entry['id']
-        self.code = entry.get("code", 0)
+
+        # Error code, string, optional
+        self.code = entry.get("code", "")
+
+        # Severity level, string (error, warning, info or debug), REQUIRED
         self.level = entry['level']
+
+        # Error message, string, REQUIRED
         self.message = entry['message']
-        self.time =  entry.get("time", None) # todo: native date/time object?
+
+        # Date and time of the error event as RFC3339 date-time, string, available since API 1.1.0
+        # todo: Use native date/time object?
+        self.time =  entry.get("time", None)
+
+        # A "stack trace" for the process, array of dicts
         self.path = entry.get("path", [])
+
+        # Related links, array of dicts
         self.links = entry.get("links", [])
+
+        # Usage metrics, dict, available since API 1.1.0
+        # May contain the following metrics: cpu, memory, duration, network, disk, storage and other custom ones
+        # Each of the metrics is also a dict with the following parts: value (numeric) and unit (string)
+        self.usage = entry.get("usage", {})
+        
+        # Arbritrary data the user wants to "log" for debugging purposes.
+        # Please note that this property may not exist as there's a difference
+        # between None and non-existing. None for example refers to no-data in
+        # many cases while the absence of the property means that the user did 
+        # not provide any data for debugging.
         if 'data' in entry:
             self.data = entry['data']
--- a/openeo/rest/connection.py
+++ b/openeo/rest/connection.py
@@ -229,6 +229,9 @@ class Connection(RestApiConnection):
         self._auth_config = auth_config
         self._refresh_token_store = refresh_token_store
 
+    def _repr_html_(self):
+        return self.capabilities()._repr_html_()
+
     @classmethod
     def version_discovery(cls, url: str, session: requests.Session = None) -> str:
         """
--- a/openeo/rest/connection.py
+++ b/openeo/rest/connection.py
@@ -937,7 +937,7 @@ class Connection(RestApiConnection):
         :param job_id: the service id of an existing secondary web service
         :return: A service object.
         """
-        return Service(service_id, self)
+        return Service(service_id, connection=self)
 
     def load_disk_collection(self, format: str, glob_pattern: str, options: dict = {}) -> ImageCollectionClient:
         """
@@ -1016,4 +1016,3 @@ def paginate(con: Connection, url: str, params: dict = None, callback: Callable 
         url = next_links[0]["href"]
         page += 1
         params = {}
-
--- a/openeo/rest/connection.py
+++ b/openeo/rest/connection.py
@@ -31,8 +31,8 @@ from openeo.rest.datacube import DataCube
 from openeo.rest.imagecollectionclient import ImageCollectionClient
 from openeo.rest.job import RESTJob
 from openeo.rest.rest_capabilities import RESTCapabilities
+from openeo.rest.service import Service
 from openeo.rest.udp import RESTUserDefinedProcess, Parameter
-from openeo.service import Service
 from openeo.util import ensure_list, legacy_alias, dict_no_none, rfc3339
 
 _log = logging.getLogger(__name__)
@@ -799,7 +799,7 @@ class Connection(RestApiConnection):
         service_id = response.headers.get("OpenEO-Identifier")
         return Service(service_id, self)
 
-    @deprecated("Use :py:meth:`openeo.service.Service.delete_service` instead.", version="0.7.0")
+    @deprecated("Use :py:meth:`openeo.rest.service.Service.delete_service` instead.", version="0.7.1")
     def remove_service(self, service_id: str):
         """
         Stop and remove a secondary web service.
--- a/openeo/rest/connection.py
+++ b/openeo/rest/connection.py
@@ -32,6 +32,7 @@ from openeo.rest.imagecollectionclient import ImageCollectionClient
 from openeo.rest.job import RESTJob
 from openeo.rest.rest_capabilities import RESTCapabilities
 from openeo.rest.udp import RESTUserDefinedProcess, Parameter
+from openeo.service import Service
 from openeo.util import ensure_list, legacy_alias, dict_no_none, rfc3339
 
 _log = logging.getLogger(__name__)
@@ -791,17 +792,14 @@ class Connection(RestApiConnection):
 
     imagecollection = legacy_alias(load_collection, name="imagecollection")
 
-    def create_service(self, graph: dict, type: str, **kwargs) -> dict:
+    def create_service(self, graph: dict, type: str, **kwargs) -> Service:
         # TODO: type hint for graph: is it a nested or a flat one?
         req = self._build_request_with_process_graph(process_graph=graph, type=type, **kwargs)
         response = self.post(path="/services", json=req, expected_status=201)
-        # TODO: "location" is url of the service metadata, not (base) url of service (https://github.com/Open-EO/openeo-api/issues/269)
-        # TODO: fetch this metadata and return a full metadata object instead?
-        return {
-            'url': response.headers.get('Location'),
-            'service_id': response.headers.get("OpenEO-Identifier"),
-        }
+        service_id = response.headers.get("OpenEO-Identifier")
+        return Service(service_id, self)
 
+    @deprecated("Use :py:meth:`openeo.service.Service.delete_service` instead.", version="0.7.0")
     def remove_service(self, service_id: str):
         """
         Stop and remove a secondary web service.
@@ -809,7 +807,7 @@ class Connection(RestApiConnection):
         :param service_id: service identifier
         :return:
         """
-        response = self.delete('/services/' + service_id)
+        Service(service_id, self).delete_service()
 
     @deprecated("Use :py:meth:`openeo.rest.job.RESTJob.get_results` instead.", version="0.4.10")
     def job_results(self, job_id) -> dict:
@@ -926,6 +924,17 @@ class Connection(RestApiConnection):
         :return: A job object.
         """
         return RESTJob(job_id, self)
+
+    def service(self, service_id: str) -> Service:
+        """
+        Get the secondary web service based on the id. The service with the given id should already exist.
+        
+        Use :py:meth:`openeo.rest.connection.Connection.create_service` to create new services
+
+        :param job_id: the service id of an existing secondary web service
+        :return: A service object.
+        """
+        return Service(service_id, self)
 
     def load_disk_collection(self, format: str, glob_pattern: str, options: dict = {}) -> ImageCollectionClient:
         """
--- a/openeo/rest/connection.py
+++ b/openeo/rest/connection.py
@@ -722,7 +722,8 @@ class Connection(RestApiConnection):
         """
         Lists all user-defined processes of the authenticated user.
         """
-        return self.get("/process_graphs").json()["processes"]
+        data = self.get("/process_graphs").json()["processes"]
+        return VisualList("processes", data = data)
 
     def user_defined_process(self, user_defined_process_id: str) -> RESTUserDefinedProcess:
         """
--- a/openeo/rest/job.py
+++ b/openeo/rest/job.py
@@ -28,12 +28,12 @@ class JobLogEntry:
 
     def __init__(self, entry):
         self.log_id = entry['id']
-        self.code = entry['code'] if 'code' in entry else 0
+        self.code = entry.get("code", 0)
         self.level = entry['level']
         self.message = entry['message']
-        self.time =  entry['time'] if 'time' in entry else None # todo: native date/time object?
-        self.path = entry['path'] if 'path' in entry else []
-        self.links = entry['links'] if 'links' in entry else []
+        self.time =  entry.get("time", None) # todo: native date/time object?
+        self.path = entry.get("path", [])
+        self.links = entry.get("links", [])
         if 'data' in entry:
             self.data = entry['data']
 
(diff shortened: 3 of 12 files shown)


openeo/rest/job.py
```diff
@@ -25,10 +26,19 @@ class JobLogEntry:
...
+        self.level = entry['level']
+        self.message = entry['message']
+        self.time =  entry['time'] if 'time' in entry else None # todo: native date/time object?
+        self.path = entry['path'] if 'path' in entry else []
+        self.links = entry['links'] if 'links' in entry else []
+        if 'data' in entry:
+            self.data = entry['data']
+
+    def toJson(self):
+        return json.dumps(self, default = lambda o: o.__dict__) # doesn't work
```
m-mohr, line 41: This class is not JSON serializable. I followed some blog posts, but couldn't get it working. Tried toJson and toJSON and multiple variants of json.dumps. Could you help, please? @soxofaan

soxofaan: what kind of failures do you get?

m-mohr: Simply that the class is not serializable. Basically the same as without the method, i.e. adding it did not change anything.

openeo/rest/job.py
```diff
@@ -25,10 +26,19 @@ class JobLogEntry:
     A Job log entry.
     """
 
-    def __init__(self, log_id: str, level: str, message: str):
-        self.log_id = log_id
-        self.level = level
-        self.message = message
+    def __init__(self, entry):
+        self.log_id = entry['id']
+        self.code = entry['code'] if 'code' in entry else 0
```
soxofaan, line 31: in Python you can write this more succinctly:

    self.code = entry.get("code", 0)

second argument is fallback value if the key is not in the dict

m-mohr: Sounds good, fixed.

m-mohr:
Also did some work on secondary web service support :-)

soxofaan:
can you add a CHANGELOG entry too?

m-mohr:
Yes, sure, will add it once I'm sure what we actually add through this PR (Services?).

openeo/service.py
```diff
@@ -1,24 +1,44 @@
 from abc import ABC
-
+from typing import List
```
soxofaan, line 2: re: https://github.com/Open-EO/openeo-python-client/issues/115#issuecomment-832515073

I think it's better to remove openeo/service.py
and implement this new stuff under openeo/rest/service.py, because it depends on a working rest Connection

soxofaan: I just removed openeo/service.py in 6b9318d87c0

openeo/api/logs.py
```diff
@@ -0,0 +1,20 @@
+import json
+
+class LogEntry:
+    """
+    A log entry.
+    """
+
+    def __init__(self, entry):
```
soxofaan, line 8: This constructor with just a generic `entry` argument API doesn't look very Pythonic or like other classes we have.
I think it is cleaner to have all fields as separate arguments to __init__, and have a classmethod `from_dict` to construct a LogEntry from a dict.

for example: https://github.com/Open-EO/openeo-python-client/blob/6b9318d87c0707cd606c0249c83dff6dcc058a02/openeo/rest/auth/oidc.py#L230-L260

openeo/api/logs.py
```diff
@@ -0,0 +1,20 @@
...
+        self.code = entry.get("code", 0)
+        self.level = entry['level']
+        self.message = entry['message']
+        self.time =  entry.get("time", None) # todo: native date/time object?
+        self.path = entry.get("path", [])
+        self.links = entry.get("links", [])
+        if 'data' in entry:
+            self.data = entry['data']
+
+    def toJson(self):
```
soxofaan, line 19: `to_json` is prefered

openeo/api/logs.py
```diff
@@ -0,0 +1,20 @@
...
+        return json.dumps(self, default = lambda o: o.__dict__) # doesn't work
```
soxofaan, line 20: does this work?

     json.dumps(self.__dict__)?

openeo/rest/connection.py
```diff
@@ -790,25 +792,22 @@ def load_collection(
...
-        # TODO: "location" is url of the service metadata, not (base) url of service (https://github.com/Open-EO/openeo-api/issues/269)
-        # TODO: fetch this metadata and return a full metadata object instead?
-        return {
-            'url': response.headers.get('Location'),
-            'service_id': response.headers.get("OpenEO-Identifier"),
-        }
+        service_id = response.headers.get("OpenEO-Identifier")
+        return Service(service_id, self)
 
+    @deprecated("Use :py:meth:`openeo.service.Service.delete_service` instead.", version="0.7.0")
```
soxofaan, line 802: next version will be at least 0.7.1, so that should be the deprecation threshold

openeo/rest/job.py
```diff
@@ -20,17 +21,6 @@
 logger = logging.getLogger(__name__)
 
 
-class JobLogEntry:
```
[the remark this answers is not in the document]
soxofaan: never mind, I just noticed you are reusing it for secondary services as well. Makes sense indeed.

soxofaan:
I notice you use this pattern in RESTJob, RESTUserDefinedProcess and Service:

```python
    def _repr_html_(self):
        return self.describe()._repr_html_()
    
    def describe(self):
        return VisualDict(....
```

In my opinion we should only do jupyter support in one place (so just the `_repr_html_`) and leave the `describe` as-is (just return a dict).

openeo/rest/job.py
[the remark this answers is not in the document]
m-mohr: Because it is not specific to jobs, it is also used in services and potentially sync processing in the future. So I removed the prefix Job and moved it to a neutral place.

openeo/rest/connection.py
m-mohr: Fixed.

openeo/api/logs.py
m-mohr: Isn't that pretty annoying to pass so many parameters individually? Also, I'm not sure how to not pass "data". There's a difference between None and not specified.

Do we actually need this wrapper? It seems pretty useless right now...

openeo/service.py
m-mohr: I've moved it to rest, but should it also get the "REST" prefix in filename and class name?

openeo/rest/job.py
m-mohr: I made it work...

openeo/api/logs.py
m-mohr: I had to fix the render_component and was able to remove it here (for now). So this is fixed. ;-)

openeo/api/logs.py
m-mohr: Had to add the default in render_component, not here. Removed it completely.

m-mohr:
> In my opinion we should only do jupyter support in one place (so just the `_repr_html_`) and leave the `describe` as-is (just return a dict).

I'm not sure I agree. Why should describe_job not display it nicely? 
Also, it seemed like in https://github.com/openEOPlatform/architecture-docs/issues/47 @jdries wants me to implement it for describe_job (see screenshot)?

openeo/service.py
soxofaan: I wouldn't add that, I'm not a fan of that REST prefix

jdries:
I had a look, but maybe it's indeed sufficient if the job object itself is already nicely rendered. The 'describe_job' seems to be triggered implicitly by that, so basically, users in jupyterlab will no longer really need 'describe_job' in most cases.
This is also the least intrusive, so lets go for leaving 'describe_job' as-is, and just have the repr_html. 
(Apologies if this contradicts with my own earlier suggestions, but I didn't gave the actual details that much thought when writing it.)

soxofaan:
> I'm not sure I agree. Why should describe_job not display it nicely?

The thing is that the openEO Python client library should be usable for non-Jupyter usage as well, so the Jupyter integration should be as non-intrusive as possible. You see that in other projects (e.g. pandas) as well: they just implement the _repr_html_ method on core objects. 
It is strange that you return these VisualDict and VisualList objects in other places, like "describe" methods. It adds a confusing level of indirection for a user that doesn't work in a jupyter context.

For example: when working on openeo-aggregator yesterday I wanted to list the available collections of a backend, and the `Connection.list_collections` method gives me a `VisualList`. It's strange to have to handle UI level objects when I just want a simple list of things.

m-mohr:
Okay, I'll change the code wrt to describe functions.

Wrt to Collections, that's, of course, a separate tasks to wrap this into Wrapper classes so that we don't need to expose those Visual things to externals and can just implement _repr_html_ on the wrapper. But did you actually had an issue in the Aggregator? I thought we inherit from dict to actually avoid this?

m-mohr:
I'm wondering how to document this best. For example in basics.rst it is:

```
    my_job = connection.job("0915ed2c-44a0-4519-8949-c58176ed2859")
    my_job.describe_job()
```

Should that remove describe_job? But that would show not so nice things in non Jupyter envs. So should __repr__ also call describe job? I could see this all getting a bit confusing when you have to hit different methods in different envs and get pretty different results.

m-mohr:
This is ready for another review. There are two open discussions above, which I'd like to get feedback on.
Other than that all feedback has been incorporated and the Vue components have been published, too.
I'll add the changelog after the next round of reviews.

openeo/api/logs.py
soxofaan: > Isn't that pretty annoying to pass so many parameters individually?

I don't see it like that, with explicit arguments, you can easily add type hints, defaults (and documentation), which is not the case with a generic dict. You indeed have to pass the parameters individually, but the end user will not have to do that. It's probably just going to be used internally in one or two places.

> Also, I'm not sure how to not pass "data" in case it's not given. There's a difference between None (JSON: null) and not specified for data.

True, but in practice I don't think a user is going to wonder about why a field is `None`: because it was not specified by backend, or because the backend provided `null` explicitly? The practical result for the user is the same: "nothing to see here, please move along.".  The same can actually be said about `code`, `path` and `links`: now you do give them a default value if they are missing, e.g. a missing `path` will give empty list.

> Do we actually need this wrapper? It seems pretty useless right now...

The added value of a  wrapper like this are I think:
- making sure missing fields get sensible default values so the end user does not have to repeat all the related boilerplate 
- end user can access fields through attributes instead of dict keys, which is safer (because IDE can detect typos and provide code completion)
- in jupyter context you can provide "pretty" rendering

openeo/api/logs.py
```diff
@@ -0,0 +1,17 @@
+import json
+
+class LogEntry:
+    """
+    A log entry.
+    """
+
+    def __init__(self, entry):
+        self.id = entry['id']
```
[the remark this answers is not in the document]
soxofaan: I don't think it's a problem.
However I don't understand what you mean with serialization here, there is no method (yet) for serialization, right?

openeo/api/logs.py
```diff
@@ -0,0 +1,17 @@
...
+        self.code = entry.get("code", 0)
+        self.level = entry['level']
+        self.message = entry['message']
+        self.time =  entry.get("time", None) # todo: native date/time object?
+        self.path = entry.get("path", [])
+        self.links = entry.get("links", [])
+        if 'data' in entry:
+            self.data = entry['data']
```
soxofaan, line 17: as noted above, I don't think it is user friendly to make the presence of the `data` property  unpredictable.
I would just give value None when it is missing

openeo/api/logs.py
```diff
@@ -0,0 +1,17 @@
...
+    """
+    A log entry.
+    """
+
+    def __init__(self, entry):
+        self.id = entry['id']
+        self.code = entry.get("code", 0)
+        self.level = entry['level']
+        self.message = entry['message']
+        self.time =  entry.get("time", None) # todo: native date/time object?
```
soxofaan, line 13: Is the "time" an official log field?

openeo/rest/connection.py
```diff
@@ -926,6 +928,17 @@ def job(self, job_id: str):
...
+    def service(self, service_id: str) -> Service:
+        """
+        Get the secondary web service based on the id. The service with the given id should already exist.
+        
+        Use :py:meth:`openeo.rest.connection.Connection.create_service` to create new services
+
+        :param job_id: the service id of an existing secondary web service
+        :return: A service object.
+        """
+        return Service(service_id, self)
```
soxofaan, line 940: I would write `Service(service_id, connection=self)` here, otherwise the `self` as second argument  looks like a typo (because `self` is usually first argument to a method definition).

openeo/rest/service.py
```diff
@@ -0,0 +1,42 @@
+from typing import List
+from openeo.internal.jupyter import VisualDict, VisualList
+from openeo.api.logs import LogEntry
+
+class Service:
+    """Represents a secondary web service in openeo."""
+
+    def __init__(self, service_id: str, connection: 'Connection'):
+        self.service_id = service_id
+        """Unique identifier of the secondary web service (string)."""
```
soxofaan, line 10: a docstring here as comment about `self.service_id` looks weird. Just use a normal `#` comment above `self.service_id`

tests/rest/datacube/test_datacube.py
```diff
@@ -411,7 +412,7 @@ def check_request(request):
     )
 
     res = s2cube.tiled_viewing_service(type="WMTS", title="S2 Foo", description="Nice!", custom_param=45)
-    assert res == {"url": API_URL + "/services/sf00", 'service_id': 'sf00'}
+    assert res.service_id == Service('sf00', connection).service_id
```
soxofaan, line 415: `Service('sf00', connection).service_id` ?
why not just `'sf00'`?

soxofaan (commented):
Sorry for the delay. Just found some time for another review pass

openeo/api/logs.py
soxofaan: side note: if we could drop python 3.5 support (#210 )  creating a LogEntry container would be very low on boilerplate when using [typing.NamedTuple](https://docs.python.org/3.6/library/typing.html#typing.NamedTuple)

soxofaan:
> I'm wondering how to document this best. For example in basics.rst it is:
>
>       my_job = connection.job("0915ed2c-44a0-4519-8949-c58176ed2859")
>       my_job.describe_job()
>
> Should that remove describe_job? But that would show not so nice things in non Jupyter envs. So should repr also call describe job? I could see this all getting a bit confusing when you have to hit different methods in different envs and get pretty different results.


What I expect as a reasonable experienced python user:
- when working in a non-jupyter environment: `__repr__` (or `__str__`) gives a compact text based representation of the object. Because the `__repr__` method is called implicitly most of the times by your runtime environment, It should be as fail-safe as possible, involve as little fancy logic as possibly, just use local state and certainly *not* do HTTP requests or JSON parsing. With your batch job example, I would expect `__repr__` to return something like `<batch job "23412423423">` and if I want  more extensive info (based on a HTTP request and JSON parsing) I want to do it explicitly with a method, e.g. `job.describe()`. 
- when working in a jupyter, it's more acceptable that `__repr__html__` involves fancy logic, including HTTP requests and JSON parsing. 

So I think the general pattern should be:

```python
class OpeneoThing:
    def __init__(self, id: str, connection: 'Connection'):
        self.id = id
        self._connection = connection

    def __repr__(self):
        return "<OpeneoThing({i})>".format(i=id)

    def describe(self) -> dict:
         return self._connection.get(...).json()

    def _repr_html_(self):
        return render_component('thing', data = self.describe())
```


Concerning the documentation: even though a lot of users work in Jupyter-like environment nowadays, I think they are fine with docs that describe things in plain, non-Jupyter Python. 
In that sense, your example doc snippet is a bit mixed in nature: 

       my_job = connection.job("0915ed2c-44a0-4519-8949-c58176ed2859")
       my_job.describe_job()

It's not clear what the side effect of the last line is. I think there are (at least) two better approaches:

- add explicit `print` (and optionally show expected result as comment), e.g. :

       my_job = connection.job("0915ed2c-44a0-4519-8949-c58176ed2859")
       print(my_job.describe_job())
       # {"id": "0915ed2c-44a0-4519-8949-c58176ed2859", ....

- suggest an interactive (but non-jupiter) Python/IPython session with `>>>`  prompts and inline result rendering:

       >>> my_job = connection.job("0915ed2c-44a0-4519-8949-c58176ed2859")
       >>> my_job.describe_job()
       {"id": "0915ed2c-44a0-4519-8949-c58176ed2859", ....

Explanation of the Jupyter features could be covered in a dedicated page of the documentation. And it's probably also more useful to cover it in example notebooks or video/webinar format.

soxofaan:
@m-mohr  on a more general note: how do you see proceeding with this ticket: 
- we keep iterating here in #206
- or I already cherry pick some parts that don't need need more discussion, to get them out of the way. This will most certainly cause merge/rebase conflicts however

openeo/api/logs.py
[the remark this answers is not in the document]
m-mohr: I meant serialization to JSON for the Vue Components...

openeo/api/logs.py
m-mohr: The README says:

> Requirement: a Python 3.6 (or higher) environment

So I'm confused.

> True, but in practice I don't think a user is going to wonder about why a field is `None`: because it was not specified by backend, or because the backend provided `null` explicitly? The practical result for the user is the same: "nothing to see here, please move along.". 

This is not correct. null (None) can actually have a specific meaning (usually `no-data`) here! This field is used for debugging and users can pass intermediate results here, in which one of the allowed values is `null` for no-data. This can be different from "doesn't exist" or "was not given/provided" by the process.

openeo/api/logs.py
m-mohr: Yes, since 1.1.0.

Actually, I also have to add the metrics here.

openeo/api/logs.py
m-mohr: See discussion above.

openeo/rest/connection.py
m-mohr:
```suggestion
        return Service(service_id, connection=self)
```

tests/rest/datacube/test_datacube.py
m-mohr: No clue, can't remember.

```suggestion
    assert res.service_id == 'sf00'
```

m-mohr:
Hmm, I think getting this PR a bit smaller would make sense if we can't touch ground on these easily. I think most things are fine except for the discussion around logs and services.

openeo/rest/service.py
m-mohr: Done.

openeo/api/logs.py
m-mohr: For now: I've added a lot more documentation about each individual property and added the metrics property, too.

openeo/api/logs.py
m-mohr: Done, see also https://github.com/Open-EO/openeo-python-client/pull/206/files#r652520824

m-mohr:
I'll split this into 3 PRs.

soxofaan:
@m-mohr I think it's best to remove this `jupyter` branch, to avoid that we continue working from it.
Unless you want to keep it for some reason?
It will be highly annoying to work with anyway because you will run in to conflicts very easily.

PR #211, opened 2021-06-10
Moved service class, integrate first round of review feedback

soxofaan:
This is an extract from PR #206:  the `Service` implementation 

because #206 was becoming quite large, I tried to extract a part of to simplify review


soxofaan:
@m-mohr   as mentioned, I extracted the `Service` part from your PR #206  in this PR 
and added some additional finetuning


This will obviously create conflict with #206, so we have to balance if it is actually worth it to split PR #206

soxofaan:
Reconsidering this PR, it's going to make things more difficult than necessary

PR #213, opened 2021-06-16
Jupyter integration for Connection, Job, Job Estimate, User-Defined Processes

m-mohr:
Based on #206, which is split into three PRs.

Adds rich HTML rendering for more backend metadata: Connections, Jobs, Job Estimates, User-Defined Processes

This should be ready for a merge, there were no open comments in #206 regarding those changes, I think.

### DIFF
--- a/openeo/rest/connection.py
+++ b/openeo/rest/connection.py
@@ -228,6 +228,9 @@ class Connection(RestApiConnection):
         self._auth_config = auth_config
         self._refresh_token_store = refresh_token_store
 
+    def _repr_html_(self):
+        return self.capabilities()._repr_html_()
+
     @classmethod
     def version_discovery(cls, url: str, session: requests.Session = None) -> str:
         """
@@ -724,7 +727,8 @@ class Connection(RestApiConnection):
         """
         Lists all user-defined processes of the authenticated user.
         """
-        return self.get("/process_graphs").json()["processes"]
+        data = self.get("/process_graphs").json()["processes"]
+        return VisualList("processes", data = data)
 
     def user_defined_process(self, user_defined_process_id: str) -> RESTUserDefinedProcess:
         """
--- a/openeo/rest/connection.py
+++ b/openeo/rest/connection.py
@@ -202,7 +202,7 @@ class Connection(RestApiConnection):
     oidc_auth_user_id_token_as_bearer = False
 
     def __init__(
-            self, url, auth: AuthBase = None, session: requests.Session = None, default_timeout: int = None,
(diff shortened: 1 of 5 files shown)


openeo/rest/connection.py
```diff
@@ -228,6 +228,9 @@ def __init__(
         self._auth_config = auth_config
         self._refresh_token_store = refresh_token_store
 
+    def _repr_html_(self):
```
soxofaan, line 231: As a user If find this `Connection._repr_html_` pretty intrusive.

I normally start my notebooks with a cell like this:

```
con = openeo.connect("https://openeo.vito.be")
con.authenticate_oidc()
```

Normally I just get a simple output cell with `<Connection to 'https://openeo.vito.be/openeo/1.0/' with BearerAuth>`, which gives me the essential information: what backend URL am I using and am I logged in?

But with this PR I suddenly get a large rich HTML element:
- it only shows the root url, but not the backend URL after version discovery
- it doesn't show if the connection is authenticated
-  I have to scroll almost a full window height to get to the next cell

soxofaan:
I did a code style cleanup and already merged this (without the `Connection._repr_html` part)

openeo/rest/connection.py
m-mohr: I wasn't aware that the authentication processes return a Connection object. It was meant that only printing the connection object itself gives this overview. Is there a way to make this happen?

I'm personally found the `<Connection to 'https://openeo.vito.be/openeo/1.0/' with BearerAuth>` always pretty strange. It gives a URL that I did not specify with some strange Auth thing that I (as a normal user) don't know anything about. I would expect something like `<Connection established to https://openeo.vito.be, version {x.x}, authenticated via {IP}>`.

m-mohr:
I must admit I don't understand the last comment...

openeo/rest/connection.py
m-mohr: Moved to #216

soxofaan:
> I must admit I don't understand the last comment(s)...

I just pushed some commits on the `jupyter-udp-jobs` branch to update this PR

soxofaan:
but now with #216  we can close this PR I think

m-mohr:
@soxofaan I don't understand. There are more changes in this PR than just the Connection object, which should still be merged.

m-mohr:
Okay, it seems the changes appeared to master without the PR being merged. I'm confused, but okay...

soxofaan:
> it seems the changes appeared on master without the PR being merged.

yeah I rebased the branch before merging in master to get rid of the merge commits with conflicts

the "remaining changes" are the parts of this PR that weren't merged in master

PR #214, opened 2021-06-16
Jupyter integration for Logs, renamed JobLogEntry

m-mohr:
Based on #206, which is split into three PRs.

Adds rich HTML rendering for more backend metadata: Logs

Class `JobLogEntry` got replaced with a more feature-complete and re-usable `LogEntry` class.

Here were some open comments regarding the LogEntry class.

### DIFF
--- a/openeo/api/logs.py
+++ b/openeo/api/logs.py
@@ -30,10 +30,6 @@ class LogEntry(dict):
         if missing:
             raise ValueError("Missing required fields: {m}".format(m=sorted(missing)))
 
-        # todo: Use native date/time object?
-        self.setdefault("time", None)
-        self.setdefault("links", [])
-
     @property
     def id(self):
         return self["id"]
@@ -48,3 +44,5 @@ class LogEntry(dict):
     @property
     def level(self):
         return self["level"]
+
+    # TODO: add properties for "code", "time", "path", "links" and "data" with sensible defaults?
--- a/openeo/api/logs.py
+++ b/openeo/api/logs.py
@@ -1,5 +1,3 @@
-import json
-
 class LogEntry:
     """
     A log entry.
@@ -20,7 +18,7 @@ class LogEntry:
 
         # Date and time of the error event as RFC3339 date-time, string, available since API 1.1.0
         # todo: Use native date/time object?
-        self.time =  entry.get("time", None)
+        self.time = entry.get("time", None)
 
         # A "stack trace" for the process, array of dicts
         self.path = entry.get("path", [])
@@ -32,7 +30,7 @@ class LogEntry:
         # May contain the following metrics: cpu, memory, duration, network, disk, storage and other custom ones
         # Each of the metrics is also a dict with the following parts: value (numeric) and unit (string)
         self.usage = entry.get("usage", {})
-        
+
         # Arbritrary data the user wants to "log" for debugging purposes.
         # Please note that this property may not exist as there's a difference
         # between None and non-existing. None for example refers to no-data in
--- a/openeo/api/logs.py
+++ b/openeo/api/logs.py
@@ -1,40 +1,34 @@
-class LogEntry:
-    """
-    A log entry.
-    """
+"""
+LogEntry - log messages for jobs and services
 
-    def __init__(self, entry):
-        # Unique ID for the log, string, REQUIRED
-        self.id = entry['id']
+Properties:
+- id: Unique ID for the log, string, REQUIRED
+- code: Error code, string, optional
+- level: Severity level, string (error, warning, info or debug), REQUIRED
+- message: Error message, string, REQUIRED
+- time: Date and time of the error event as RFC3339 date-time, string, available since API 1.1.0
+- path: A "stack trace" for the process, array of dicts
+- links: Related links, array of dicts
+- usage: Usage metrics available as property 'usage', dict, available since API 1.1.0
+  May contain the following metrics: cpu, memory, duration, network, disk, storage and other custom ones
+  Each of the metrics is also a dict with the following parts: value (numeric) and unit (string)
+- data: Arbritrary data the user wants to "log" for debugging purposes.
+  Please note that this property may not exist as there's a difference
+  between None and non-existing. None for example refers to no-data in
+  many cases while the absence of the property means that the user did 
+  not provide any data for debugging.
+"""
 
-        # Error code, string, optional
-        self.code = entry.get("code", "")
+class LogEntry(dict):
 
-        # Severity level, string (error, warning, info or debug), REQUIRED
-        self.level = entry['level']
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
 
-        # Error message, string, REQUIRED
-        self.message = entry['message']
-
-        # Date and time of the error event as RFC3339 date-time, string, available since API 1.1.0
+        # Check required fields
+        missing = self._required.difference(self.keys())
+        if missing:
+            raise ValueError("Missing required fields: {m}".format(m=sorted(missing)))
+        
         # todo: Use native date/time object?
-        self.time = entry.get("time", None)
-
-        # A "stack trace" for the process, array of dicts
-        self.path = entry.get("path", [])
-
-        # Related links, array of dicts
-        self.links = entry.get("links", [])
-
-        # Usage metrics, dict, available since API 1.1.0
-        # May contain the following metrics: cpu, memory, duration, network, disk, storage and other custom ones
-        # Each of the metrics is also a dict with the following parts: value (numeric) and unit (string)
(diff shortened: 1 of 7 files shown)


openeo/api/logs.py
```diff
@@ -0,0 +1,40 @@
...
+        # Each of the metrics is also a dict with the following parts: value (numeric) and unit (string)
+        self.usage = entry.get("usage", {})
+
+        # Arbritrary data the user wants to "log" for debugging purposes.
+        # Please note that this property may not exist as there's a difference
+        # between None and non-existing. None for example refers to no-data in
+        # many cases while the absence of the property means that the user did 
+        # not provide any data for debugging.
+        if 'data' in entry:
+            self.data = entry['data']
```
soxofaan, line 40: I don't like this conditional `data` property: 
it forces the end user to always add an additional check (like `hasattr(log, "data")` ) before they can inspect the value. I guess in most cases the user is not interested in the subtlety of `None` versus "not specified", they just want to inspect the value.

Note that you are assigning default values for other fields like code, time, path, links, usage, so why not for data as well?

openeo/internal/jupyter.py
```diff
@@ -105,7 +105,7 @@ def render_component(component: str, data = None, parameters: dict = {}):
     """.format(
         script = SCRIPT_URL,
         component = component,
-        props = json.dumps(parameters)
+        props = json.dumps(parameters, default = lambda o: o.__dict__) # default parameter added to serialize LogEntry
```
soxofaan, line 108: This `default = lambda o: o.__dict__` basically means "export everything" and is pretty liberal. I'm a bit worried whether this will cause accidental exposure of things you don't want to expose (not now, but in the longer term when Jupyter support will expand).

As far as I understand this is just to make LogEntry items JSON-seriazable, right?

openeo/api/logs.py
m-mohr: > Note that you are assigning default values for other fields like code, time, path, links, usage, so why not for data as well?

Because there's no difference between None and not existing for the others.

Can we assign any value as default that is not part of JSON for data? The first thought is NaN, but that's maybe not ideal.

openeo/internal/jupyter.py
m-mohr: Yes.

soxofaan: I pushed an alternative implementation that only does the `o.__dict__` on `LogEntry`  instances

openeo/api/logs.py
soxofaan: > Because there's no difference between None and not existing for the others.

Yes, but my point is that you still are changing the value you receive from the backend. 
For example: if a log you receive from backend does not have a `links` field, you still are assigning `log.links = []` and you are exporting `"links": []`  when you generate the HTML/JS with render_component.

m-mohr: Yes, but except for thr data property it doesn't make any difference in the UI.

m-mohr: I'm only assigning the defaults here so that it's easier to use in Python. For the UI components there's no difference whether they have defaults or not.

soxofaan: An alternative solution is to go back to a dictionary based structure, because dictionaries by design allow having undefined fields. Then we don't have to reinvent the hot water.

If you define `LogEntry` as a dictionary subclass you could  apply default values where that would make sense.

What could also help is this `dict_item` utility we have in the openeo_driver project: https://github.com/Open-EO/openeo-python-driver/blob/407f83bb231713cacd81e863de34fc3f02f19ea9/openeo_driver/utils.py#L250-L287
It allows accessing dictionary fields through an attribute  (`log.message`)  in addition to classic dict keys (`log["message").

soxofaan: Proof of concept alternative:
```

class LogEntry(dict):

    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        # Check required fields
        for required in ["id", "level", "message"]:
            if required not in self:
                raise ValueError("Field {r!r} is required".format(r=required))
        # Set some default fields
        self.setdefault("time", None)
        self.setdefault("links", [])
```

m-mohr: That makes a lot of sense to me. Will look into it.

soxofaan: or to show all missing fields at once:

```python
class LogEntry(dict):
    _required = {"id", "level", "message"}

    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        # Check required fields
        missing = self._required.difference(self.keys())
        if missing:
            raise ValueError("Missing required fields: {m}".format(m=sorted(missing)))
        # Set some default fields
        self.setdefault("time", None)
        self.setdefault("links", [])


log = LogEntry(id="log1", level="info", message="hello")
print(log)
# {'id': 'log1', 'level': 'info', 'message': 'hello', 'time': None, 'links': []}

log = LogEntry(id="log2")
# ValueError: Missing fields: ['level', 'message']
```

m-mohr: Integrated, works for me.

m-mohr: Reverted the changes regarding json serialization again, we neither need my initial proposal nor your improvement anymore, just plain json.dumps.

openeo/api/logs.py
```diff
@@ -1,40 +1,34 @@
...
-        self.code = entry.get("code", "")
-
-        # Severity level, string (error, warning, info or debug), REQUIRED
-        self.level = entry['level']
-
-        # Error message, string, REQUIRED
-        self.message = entry['message']
-
-        # Date and time of the error event as RFC3339 date-time, string, available since API 1.1.0
+"""
```
soxofaan, line 1: this "doc block" (triple quoted string) should come after `class LogEntry` line

openeo/api/logs.py
```diff
@@ -1,40 +1,34 @@
...
+  not provide any data for debugging.
+"""
+
+class LogEntry(dict):
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        # Check required fields
+        missing = self._required.difference(self.keys())
```
soxofaan, line 28: `_required` is not defined, right?

openeo/api/logs.py
soxofaan: > Reverted the changes regarding json serialization again, 

yes that's a nice side effect

openeo/api/logs.py
m-mohr: Fixed.

openeo/api/logs.py
m-mohr: Yes, forgot to copy, fixed.

soxofaan: that's why the tests fail FYI

m-mohr: Yes, forgot to copy, fixed.

m-mohr: Still failing, but don't have access to the Jenkins...

soxofaan: latest failures re due to github connection issue, the tests themselves didn't even run. Not sure if it just a temporary glitch or something serious

m-mohr: Figured it out, should work now.

openeo/api/logs.py
```diff
@@ -0,0 +1,35 @@
...
+        if missing:
+            raise ValueError("Missing required fields: {m}".format(m=sorted(missing)))
+        
+        # todo: Use native date/time object?
+        self.setdefault("time", None)
+        self.setdefault("links", [])
```
soxofaan, line 35: FYI: I just used default for only "time" and "links" as an example,   we might  want defauls for  "path", "usage" too. Or no default at all. I'm not sure what is the most useful/sensible at the moment

m-mohr: I've no real preference here. We can give an empty array for path and an empty dict for usage, but it also works well without. Whatever you prefer.

openeo/api/logs.py
soxofaan: tests run for me locally as well, so that's no blocker

Our s-ops team is looking at broken jenkins builds (it's a more serious problem apparently), I have no idea how long that will take

soxofaan: oh sorry I spoke to soon, there are two actual test failures I didn't notice. 
I'll try to fix them

openeo/api/logs.py
soxofaan: I'd drop the defaults for now then

openeo/api/logs.py
soxofaan: fixed now, all tests pass on my setup

soxofaan:
I rebased this PR, cleaned up the merge conflicts and merged in master in 5850514

PR #220, opened 2021-07-07
Support for visualized process graphs in HTML components

m-mohr:
Support for visualized process graphs (model builder) in HTML components. 

Note: The JS gets more ugly, but I need to load FontAwesome icons to show the Model Builder in Vue Components v2.3+ nicely and I have to load the font here as I can't load it programmatically in Web Components. Browsers currently don't load fonts for them, that's a known limitation.


m-mohr:
Example for mask_scl_dilation:

image

openeo/rest/connection.py
```diff
@@ -686,7 +686,7 @@ def list_processes(self) -> List[dict]:
         :return: processes_dict: Dict All available processes of the back end.
         """
         data = self.get('/processes').json()["processes"]
-        return VisualList("processes", data = data)
+        return VisualList("processes", data = data, parameters = {'show-graph': True})
```
soxofaan, line 689: will this trigger the visualization of the process graphs of all process (where available)?

m-mohr: Yes, of course only once expanded.

soxofaan:
Can you provide a notebook where it works, because I can't get the process graph visualization to work with my initial testing against VITO backend. I just get an empty space under "processing Instructions":

Screenshot from 2021-07-13 16-43-14

m-mohr:
Any errors in the console?

soxofaan:
no errors in console, no. Also no HTTP issues in network tab.

(using Firefox 89 on linux FYI)

m-mohr:
This notebook works for me: [ui-elements.zip](https://github.com/Open-EO/openeo-python-client/files/6809832/ui-elements.zip)

m-mohr:
image

m-mohr:
(I should probably disable the download JSON thing in Jupyter)

soxofaan:
I got it working now on my setup. Looks nice.

You are planning to do more finetuning @m-mohr ?

For example, the scroll behavior could be annoying: you are scrolling through your notebook, at one point your mouse pointer hovers one of the process graph visualization and suddenly you are zooming in/out on that visualization. I'm not sure what could be a good solution.

m-mohr:
Indeed, it could be nice to implement something so that it only zooms e.g. when STRG is pressed. I think that's what I've seen for some Maps. Let me check whether I can easily implement this. It will likely not be released in the next couple of days, but I can already set the parameter here so that it just adapts the changes once released.

m-mohr:
@soxofaan I have implemented it in the Vue components. This PR can be merged as it is, it doesn't require any changes. The improved Zoom behavior will get available automatically once released.

m-mohr:
@soxofaan The improvements for the Model Builder have been deployed/published.

soxofaan:
can you also add a CHANGELOG item?

m-mohr:
@soxofaan Added the changelog entry

PR #230, opened 2021-09-06
add proper child callback support for fit_curve and predict_curve

### LINKED ISSUE #229: Proper callable support in fit_curve and related
from https://github.com/openEOPlatform/documentation/issues/10

```python
curve_fitting = l2a_bands.fit_curve(
    parameters=[1,1,1], # Initial guess of the parameters
    dimension="t",      # Fit the function along the temporal dimension
    function=l2a_bands._get_callback(fit_function, parent_parameters=["data","parameters"])
)
```

it's unfortunate that you have to call a private method `l2a_bands._get_callback` like that, we should make sure that you can just do `function=fit_function` directly

soxofaan:
fixes #229 

related to  openEOPlatform/documentation#10

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ([#225](https://github.com/Open-EO/openeo-python-client/issues/225), [openeo-api#410](https://github.com/Open-EO/openeo-api/issues/410))
 - Add `DataCube.dimension_labels()` (EP-4008)
 - Add `Connection.load_result()` (EP-4008)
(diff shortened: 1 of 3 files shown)


m-mohr:
Is there a way to enable this globally? So e.g. also for the .process(...) function?

In JS we are checking the parameters and when they are of type function then handle them as callback. Would something like that also work in Python? It would make custom functions with callbacks easier to use (in general) and generate (in the Web Editor).

soxofaan:
It's probably possible to support this generally

but the fix per case is just a simple straightforward one-liner, 
and I'm not sure that some generic way will be able to shave off a lot or make things simpler

m-mohr:
I've opened a separate issue for it so that we can consider this later and don't block this PR: #232

PR #231, opened 2021-09-06
Add ProcessBuilder `array_element(data, n)` through `data[n]` syntax

### LINKED ISSUE #228: ProcessBuilder __getitem__ support
from https://github.com/openEOPlatform/documentation/issues/10

```python
def fit_function(x:ProcessBuilder, parameters):
    pi = math.pi
    a0 = array_element(parameters, 0)
    a1 = array_element(parameters, 1)
    a2 = array_element(parameters, 2)
    return a0 + a1*cos(2*pi/31557600*x) + a2*sin(2*pi/31557600*x) # 31557600 are the seconds in one year
```

It would be nice to allow writing `parameters[0]` instead of `array_element(parameters, 0)`

soxofaan:
Fixes #228

related to  openEOPlatform/documentation#10

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add `DataCube.dimension_labels()` (EP-4008)
 - Add `Connection.load_result()` (EP-4008)
 - Add proper support for child callbacks in `fit_curve` and `predict_curve` ([#229](https://github.com/Open-EO/openeo-python-client/issues/229))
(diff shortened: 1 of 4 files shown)


m-mohr:
I like that this is supporting both indices and labels.

I think it would make sense to document that "integer" labels are not supported. (I have the same issue in the  JS client).

soxofaan:
> I think it would make sense to document that "integer" labels are not supported. (

Indeed, a user however can still use integer labels by directly using `array_element`,  
this `__getitem__` trick is just syntactic sugar

m-mohr:
Yes, I only meant this special case.

PR #233, opened 2021-09-07
Remove the invalid default CRS

m-mohr:
Remove `crs = "EPSG:4326"` from examples due to the following reasons:

1. EPSG:4326 is not a valid value according to the process specification. So specifying it reduces portability.
2. EPSG 4326 (given as `4326`) is the default CRS anyway, so it is redundant.

Also, the Python client should not send the crs as null.

I have not updated the Notebooks.


soxofaan:
Getting rid of invalid `"EPSG:4326"` is fine for me

But I'm not sure we should remove it completely instead of replacing it with valid int `4326`: even tough it is the default value, it has documentation value to keep it (at least) in the examples.

m-mohr:
I've added a non 4326 example to the filter_bbox documentation.

If all examples use 4326 and we want some non-default examples, we should probably replace some with non 4326 bboxes as examples instead.

soxofaan:
To simplify review, I already cherry-picked a large part of the original PR in 75e4a58b9c94400800dc3b652b6df5c7566e1232

docs/basics.rst
```diff
@@ -131,7 +131,7 @@ pixels to 0 and all other pixels to 1 using a simple comparison::
     s2_sceneclassification = (
         connection.load_collection("TERRASCOPE_S2_TOC_V2", bands=["SCENECLASSIFICATION_20M"])
         .filter_temporal(extent=["2016-01-01", "2016-03-10"])
-        .filter_bbox(west=5.1518, east=5.1533,south=51.1819,north=51.1846, crs=4326)
+        .filter_bbox(west=5.1518, east=5.1533,south=51.1819,north=51.1846)
```
soxofaan, line 134: I would keep these two cases of `crs 4326` because of its documentation value

openeo/rest/imagecollectionclient.py
```diff
@@ -122,9 +122,10 @@ def _filter_temporal(self, start: str, end: str) -> 'ImageCollection':
 
     def filter_bbox(self, west, east, north, south, crs=None, base=None, height=None) -> 'ImageCollection':
         extent = {
-            'west': west, 'east': east, 'north': north, 'south': south,
-            'crs': crs,
+            'west': west, 'east': east, 'north': north, 'south': south
         }
+        if crs is not None:
+            extent.update(crs=crs)
```
soxofaan, line 128: There should be a bit of test coverage of this (and on datacube.py too)

m-mohr:
I'll need to get back to this after my vacation, so this will stay open and unresolved for about 3 weeks.

soxofaan:
No problem, I kept some things open because I wanted to improve unit test coverage of them first
I can resolve the remaining bits myself

docs/basics.rst
m-mohr: I guess it would make sense to describe somewhere that 4326 is the default and can be omitted - or use another CRS for the example.

soxofaan: yeah that line (and other lines)  "need work" anyway: you have to scroll horizontally to see everything on https://open-eo.github.io/openeo-python-client/basics.html#example-applying-a-mask

soxofaan:
The remaining bits are now also incorporated in (1fe248f)

(closes this PR)

Issue #238, opened 2021-09-13
DataCube.download: autodetect format


soxofaan:
```
cube.download("foo.tiff")  # -> autodetect to use format "GTiff"
cube.download("foo.nc")  # -> autodetect to use format "NetCDF"

```

jonathom:
see [pull request #244](https://github.com/Open-EO/openeo-python-client/pull/244).

PR #244, opened 2021-10-04
download: autodetect format (see #238)

jonathom:
Are all cases covered so far?


openeo/rest/imagecollectionclient.py
```diff
@@ -970,8 +970,29 @@ def save_result(self, format: str = "GTIFF", options: dict = None):
             }
         )
 
-    def download(self, outputfile: str = None, format: str = "GTIFF", options: dict = None):
+    def download(self, outputfile: str = None, format: str = None, options: dict = None):
         """Download image collection, e.g. as GeoTIFF."""
+        if format == None:
+            dots = [i for i, ltr in enumerate(outputfile) if ltr == "."]
+            format_string = outputfile[dots[len(dots)-1]+1:len(outputfile)].upper()
```
soxofaan, line 977: these two lines written can be a lot easier:
```python
extension = outputfile.rsplit(".", 1)[-1].upper()
```

openeo/rest/imagecollectionclient.py
```diff
@@ -970,8 +970,29 @@ def save_result(self, format: str = "GTIFF", options: dict = None):
...
+            elif format_string == "NC":
+                format = "NetCDF"
+            elif format_string == "CSV":
+                format = "CSV"
+            elif format_string == "PNG":
+                format = "PNG"
+            elif format_string == "COVJSON":
+                format = "COVJSON"
+            else:
+                raise AttributeError("A correct format string should be provided, as the format couldn't be detected from the outputfile.")
```
soxofaan, line 994: `AttributeError` is not the correct exception to raise here (it's for when you have an object and request an attribute that doesn't exist).  A `ValueError` or `OpenEoClientException` would be better.

soxofaan: Also, instead of raising an exception, we could also just use the file extension directly as format. If that doesn't make sense, the back-end will complain anyway

openeo/rest/imagecollectionclient.py
```diff
@@ -970,8 +970,29 @@ def save_result(self, format: str = "GTIFF", options: dict = None):
...
         )
 
-    def download(self, outputfile: str = None, format: str = "GTIFF", options: dict = None):
+    def download(self, outputfile: str = None, format: str = None, options: dict = None):
         """Download image collection, e.g. as GeoTIFF."""
+        if format == None:
+            dots = [i for i, ltr in enumerate(outputfile) if ltr == "."]
+            format_string = outputfile[dots[len(dots)-1]+1:len(outputfile)].upper()
+
+            if format_string == "TIFF":
```
soxofaan, line 979: code tip: instead of this large if-elif-construct you can also define a dictionary mapping file extensions to formats, e.g.
```python
format_map = {
    "GTIFF": "GTIFF", "GEOTIFF": "GTIFF", "TIFF": "GTIFF", "TIF": "GTIFF",
    "JSON": "JSON",
    "NC": "NetCDF", "NETCDF": "NetCDF",
...
format = format_map[extension]
```

that's more compact and easier to maintain

soxofaan:
thanks for this PR!

This format guessing should be added to `DataCube.download` too, so instead of putting it directly in the `download` methods, I would add a function `guess_format` to the `util.py` module

Also, if you are familiar with unit testing, I would recommend to add some tests too

Also, you can also add a item to CHANGELOG.md about this.

jonathom:
Thanks for all the help @soxofaan! I went for not throwing an exception and passing everything through. This actually makes the dict much shorter since png, csv, json and covjson just need to be converted to uppercase to be the right format.
I'm not very familiar with unit tests but they seem to have passed :) Should there be tests for the use of the helper function in the `download` functions as well?

openeo/rest/datacube.py
```diff
@@ -1469,17 +1469,20 @@ def save_result(self, format: str = "GTiff", options: dict = None) -> 'DataCube'
...
         If outputfile is provided, the result is stored on disk locally, otherwise, a bytes object is returned.
         The bytes object can be passed on to a suitable decoder for decoding.
 
         :param outputfile: Optional, an output file if the result needs to be stored on disk.
-        :param format: Optional, defaults to "GTIFF", an output format supported by the backend.
+        :param format: Optional, an output format supported by the backend.
         :param options: Optional, file format options
         :return: None if the result is stored to disk, or a bytes object returned by the backend.
         """
+        if format == None:
```
jonathom, line 1483: would we need a check here that ensures that the `outputfile` string is there?

openeo/rest/datacube.py
```diff
@@ -1469,17 +1469,20 @@ def save_result(self, format: str = "GTiff", options: dict = None) -> 'DataCube'
...
+            format = guess_format(outputfile = outputfile)
```
soxofaan, line 1484: some python tips/nitpicking:
- checking for `None`  should be done with `x is None` instead of `x == None` because `None` is an object and not a value (roughly speaking)
- a common pattern to simplify this kind of default handling `if` construct is to write it as `format = format or guess_format(outputfile)`

openeo/util.py
```diff
@@ -375,6 +375,21 @@ def deep_set(data: dict, *keys, value):
         raise ValueError("No keys given")
 
 
+def guess_format(outputfile: str = None):
```
soxofaan, line 378: In this context, I would just use `filename` as argument name, it's not necessary to assume it is an _output_ file

also, the default value `None` does not make sense (and won't work  anyway)

openeo/util.py
```diff
@@ -375,6 +375,21 @@ def deep_set(data: dict, *keys, value):
...
+    """
+    Guess the output format from a given filename and return the corrected format.
+    Any names not in the dict get passed through in upper case.
+    """
+    format_string = outputfile.rsplit(".", 1)[-1].upper()
+
+    format_map = {
+        "GTIFF": "GTIFF", "GEOTIFF": "GTIFF", "GEOTIF": "GTIFF", "TIFF": "GTIFF", "TIF": "GTIFF",
+        "NC": "NetCDF", "NETCDF": "NetCDF",
+    }
```
soxofaan, line 388: some more nitpicking :smile:  :

- according to https://gdal.org/drivers/raster/index.html  it's `netCDF`, not `NetCDF`
- if you use lowercase for `format_string` (which I would rename to `extension` for clarity), the dict will be more readable: keys in lowercase, values in upper case (mostly)

openeo/util.py
```diff
@@ -375,6 +375,21 @@ def deep_set(data: dict, *keys, value):
...
+    
+    return(format_map.get(format_string, format_string))
```
soxofaan, line 390: return is not a function: no need for parantheses

openeo/rest/datacube.py
soxofaan: hmm yeah there is indeed a conflict with "If outputfile is provided..., otherwise, a bytes object is returned."

openeo/util.py
soxofaan: and apparently it's also `GTiff` instead of `GTIFF`

openeo/rest/datacube.py
soxofaan: I wasn't aware of this "return as bytes" feature

the behavior could be:
- `download(outputfile=None, format=None)` -> GTiff format, return as bytes
- `download(outputfile=None, format=X)`   format `X`, return as bytes
- `download(outputfile="foo.tiff", format=None)` ->  GTiff format, download to file
- `download(outputfile="result", format=X)`   -> format `X`, download to file

openeo/util.py
jonathom: Should I add GTIFF as a default or no default at all?

soxofaan:
> should there be tests for the use of the helper function in the download functions as well?

At minimum the helper function should be covered, as you already did


Given the "download as bytes" feature discussed above, it would be good to also cover these different download use cases. This is less trivial (because you have to mock API requests). I'll see if I can find a good example you can work from

openeo/util.py
soxofaan: I meant that  argument `outputfile`  should have no default value

openeo/rest/datacube.py
jonathom: Does `imagcollectionclient.py` need the same behaviour? It's not documented for that function, but there is no default now either.

openeo/rest/datacube.py
```diff
@@ -1480,8 +1480,10 @@ def download(self, outputfile: Union[str, pathlib.Path, None] = None, format: st
         :param options: Optional, file format options
         :return: None if the result is stored to disk, or a bytes object returned by the backend.
         """
-        if format == None:
-            format = guess_format(outputfile = outputfile)
+        if outputfile is None and format is None:
+            format = "GTiff"
+        else:
+            format = format or format = guess_format(outputfile)
```
soxofaan, line 1486: that's not what I meant syntax-wise, it's just
```python
format = format or guess_format(outputfile)
```

soxofaan: but due to this `outputfile=None`  case I din't think of, it's probably easier to do the whole thing with:
```python
if not format:
```

jonathom: I have trouble understanding your code example.
Right now I changed it to
```python
        if outputfile is None and format is None:
            format = "GTiff"
        elif format is None:
            format = guess_format(outputfile)
```
to cover the behaviour you described above (default to GTiff as bytes when on outputfile is provided).

However only putting `format = format or guess_format(outputfile)` crashes tests in imagecollectionclient. should there be an `if` in front of it?

jonathom: > but due to this outputfile=None case I din't think of, it's probably easier to do the whole thing with:

Saw that only now, thanks! ignore the comment above then.

soxofaan:
in d476d73 I added some tests about `DataCube.download()` as inspiration for testing the different download use cases

jonathom:
Alright there are some more tests now covering all cases. However I'm not sure if they do what we want. E.g.
```
def test_download_bytes_no_format(connection, requests_mock):
    requests_mock.get(API_URL + "/collections/S2", json={})
    requests_mock.post(API_URL + '/result', content=b"tiffdata")
    result = connection.load_collection("S2").download(None)
    assert result == b"tiffdata"
```
Here, the format should default to `GTiff`, but how can we see that it does? Or is all that's important that the download returned the correct payload?

soxofaan:
This is a rewrite of test_download_path_str that show how you can check the `format` that is sent to the back-end

```python
def test_download_path_str(connection, requests_mock, tmp_path, api_version):
    requests_mock.get(API_URL + "/collections/S2", json={})

    def result_callback(request, context):
        post_data = request.json()
        pg = (post_data["process"] if api_version >= ComparableVersion("1.0.0") else post_data)["process_graph"]
        assert pg["saveresult1"]["arguments"]["format"] == "GTiff"
        return b"tiffdata"

    requests_mock.post(API_URL + '/result', content=result_callback)
    path = tmp_path / "tmp.tiff"
    connection.load_collection("S2").download(str(path), format="GTiff")
    assert path.read_bytes() == b"tiffdata"
```

jonathom:
Thanks, I added this now. Only thing is that I can't test for other extensions except GTiff related ones. From printing I know that also `.nc` etc. gets passed through correctly, but it seems that another function is only allowing GTiff output.

Screenshot from 2021-10-05 15-52-05

I think the changes in this PR are sufficiently tested though. If it guesses correctly from `.tiff`, then the whole thing works right I'd say.

soxofaan:
indeed, the fake backend in those tests only claims to support GTiff :

https://github.com/Open-EO/openeo-python-client/blob/d476d7362ec6b2501471f457a54968f38d332b72/tests/rest/datacube/conftest.py#L48-L52

you could override that `requests_mock.get(API_URL + "/file_formats",` call directly in your test function to also support netcdf, json, csv, ...

I think it would be good to at least test for netcdf too

soxofaan:
or easier: just add netcdf to https://github.com/Open-EO/openeo-python-client/blob/d476d7362ec6b2501471f457a54968f38d332b72/tests/rest/datacube/conftest.py#L48-L52

jonathom:
Done :+1:

soxofaan:
Thanks for all the iterations :)

I've squashed all your commits to one (5f8bba9),
did some finetuning of the tests (8718329)
and merged in master.

this issue can be closed.

soxofaan:
minor note @jonathom : I noticed that you worked for this PR on the master branch in your personal fork (jonathom:master).
your master branch will now have conflicts with the master of the  official repo.
I think it's best to reset your master branches (locally and on your github fork), if you're not familiar with that, I can help you out.

In any case, it's recommended to start a fresh feature branch for each PR you are planning to do

jonathom:
Thanks!
> minor note @jonathom

Thank you for this, however it was only temporary. Now that I have access to the repo I will work on branches (per feature).

PR #245, opened 2021-10-05
rename id in connection.load_collection

jonathom:
Should the parameter also be changed in `datacube.load_collection()` and `imagecollectionclient.load_collection()`?

### DIFF
--- a/openeo/rest/connection.py
+++ b/openeo/rest/connection.py
@@ -823,7 +823,7 @@ class Connection(RestApiConnection):
 
     def load_collection(
             self,
-            collection_id: str,
+            id: str,
             spatial_extent: Optional[Dict[str, float]] = None,
             temporal_extent: Optional[List[Union[str, datetime.datetime, datetime.date]]] = None,
             bands: Optional[List[str]] = None,
@@ -833,7 +833,7 @@ class Connection(RestApiConnection):
         """
         Load a DataCube by collection id.
 
-        :param collection_id: image collection identifier
+        :param id: image collection identifier
         :param spatial_extent: limit data to specified bounding box or polygons
         :param temporal_extent: limit data to specified temporal interval
         :param bands: only add the specified bands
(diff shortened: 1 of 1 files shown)


openeo/rest/connection.py
```diff
@@ -833,7 +833,7 @@ def load_collection(
         """
         Load a DataCube by collection id.
 
-        :param collection_id: image collection identifier
```
m-mohr, line 836: Should this be deprecated instead of being removed?

jonathom: Would I just put `@deprecated("Use 'id' instead") above the `:param collection_id` line?

m-mohr: Sounds good to me, but @soxofaan should have the final vote.

openeo/rest/connection.py
```diff
@@ -833,7 +833,7 @@ def load_collection(
...
+        :param id: image collection identifier
```
jonathom, line 836:
```suggestion
```suggestion
        @deprecated("Use 'id' instead!")
        :param collection_id: image collection identifier
        :param id: image collection identifier
```

openeo/rest/connection.py
soxofaan: I don't think we can make a breaking change like this.
Existing code that uses keyword argument `load_collection(collection_id="S2")` will break.

There should be a fallback mechanism:
- give `id` default value None
- add a `**kwargs` and check if `collection_id` is set (and raise deprecation warning if that is the case)

openeo/rest/connection.py
```diff
@@ -842,13 +842,13 @@ def load_collection(
         """
         if self._api_version.at_least("1.0.0"):
             return DataCube.load_collection(
-                collection_id=collection_id, connection=self,
+                collection_id=id, connection=self,
```
soxofaan, line 845: I think the `load_collection` methods of DataCube and ImageCollectionClient should also be addressed

but here it's fine to do it in a backwards incompatible way because these are practically "internal" methods nobody should be using

soxofaan:
also to do:

- test coverage for old style and new style
- Changelog entry because this is somewhat "breaking" (with fallback mechanism)

openeo/rest/connection.py
m-mohr: Yes, that's what I meant to say :-D

jonathom: @soxofaan
```python
def load_collection(
    self,
    id: str = None,
[...]
    if "collection_id" in kwargs:
        id = kwargs["collection_id"]
        warnings.warn("The use of `collection_id` is deprecated, use `id` instead.", DeprecationWarning)
```
like so?

jonathom:
> test coverage for old style and new style

What do you mean by this?

soxofaan:
test that both `load_collection(collection_id="Sentinel2")` and `load_collection(id="Sentinel2")` work

jonathom:
Alright, did that. Didn't add tests for `imagecollectionclient` and `datacube` yet, but I can if you want @soxofaan.

PR #248, opened 2021-10-11
UDP storing: categories, links etc.

jonathom:
Things I couldn't figure out so far:

* How to test for the added parameters
* This part of `openeo/rest/udp.py`: 
```    
if categories is not None:
        process["categories"] = [
            (c if isinstance(c, str) else str(c))
            for c in categories
        ]
    if examples is not None:
        process["examples"] = [
            (e if isinstance(e, dict) else e.to_dict())
            for e in examples
        ]
    if links is not None:
        process["links"] = [
            (l if isinstance(l, str) else l.to_dict())
            for l in links
        ]
```
Really unsure about these `else *.to_dict()` conversions.

### DIFF
--- a/openeo/rest/connection.py
+++ b/openeo/rest/connection.py
@@ -717,7 +717,9 @@ class Connection(RestApiConnection):
             self, user_defined_process_id: str,
             process_graph: Union[dict, ProcessBuilderBase],
             parameters: List[Union[dict, Parameter]] = None,
-            public: bool = False, summary: str = None, description: str = None
+            public: bool = False, summary: str = None, description: str = None,
+            returns: Optional[dict] = None, categories: Optional[List[str]] = None, 
+            examples: Optional[List[dict]] = None, links: Optional[List[dict]] = None
     ) -> RESTUserDefinedProcess:
         """
         Store a process graph and its metadata on the backend as a user-defined process for the authenticated user.
@@ -736,7 +738,8 @@ class Connection(RestApiConnection):
         if not parameters:
             warnings.warn("Defining user-defined process {u!r} without parameters".format(u=user_defined_process_id))
         udp = RESTUserDefinedProcess(user_defined_process_id=user_defined_process_id, connection=self)
-        udp.store(process_graph=process_graph, parameters=parameters, public=public,summary=summary,description=description)
+        udp.store(process_graph=process_graph, parameters=parameters, public=public,summary=summary,description=description,
+        returns=returns, categories=categories, examples=examples, links=links)
         return udp
 
     def list_user_defined_processes(self) -> List[dict]:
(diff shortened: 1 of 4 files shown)


openeo/rest/udp.py
```diff
@@ -31,20 +34,38 @@ def build_process_dict(
...
+    if examples is not None:
+        process["examples"] = [
+            (e if isinstance(e, dict) else e.to_dict())
+            for e in examples
+        ]
+    if links is not None:
+        process["links"] = [
+            (l if isinstance(l, dict) else l.to_dict())
+            for l in links
+        ]
```
soxofaan, line 68: Why is all this `... if isinstance ... to_dict ... for x in ...` necessary? 
The type hint already documents that e.g. `categories` is a list of strings, so why the need to for extra conversion?

soxofaan: note the `to_dict` stuff is necessary for "parameters", because these can be `Parameter` objects which are not JSON-serializable. But for categories, examples, links there is no need for explicit conversion here because lists/dicts are JSON-serializable out of the box

jonathom: Alright, I thought it had something to do with the `Lists`. I removed it now.

openeo/rest/connection.py
```diff
@@ -736,7 +738,8 @@ def save_user_defined_process(
         if not parameters:
             warnings.warn("Defining user-defined process {u!r} without parameters".format(u=user_defined_process_id))
         udp = RESTUserDefinedProcess(user_defined_process_id=user_defined_process_id, connection=self)
-        udp.store(process_graph=process_graph, parameters=parameters, public=public,summary=summary,description=description)
+        udp.store(process_graph=process_graph, parameters=parameters, public=public,summary=summary,description=description,
+        returns=returns, categories=categories, examples=examples, links=links)
```
soxofaan, line 742: if you make this a multiline function call, preferably indent as follows:
```python
        udp.store(
            process_graph=process_graph, parameters=parameters, public=public,summary=summary,description=description,
            returns=returns, categories=categories, examples=examples, links=links
        )
```

tests/rest/test_udp.py
```diff
@@ -21,7 +21,7 @@ def test_describe(con100, requests_mock):
 
     udp = con100.user_defined_process(user_defined_process_id='evi')
     details = udp.describe()
-
+
```
soxofaan, line 24: Can you add a new test too (I guess you wanted to do that here)

for example copy https://github.com/Open-EO/openeo-python-client/blob/bde2d0f992bd52fc244c8bfeceac4e58d6b12c2d/tests/rest/test_udp.py#L28-L51   and create a test `test_store_full` that uses and checks for all these added properties

soxofaan:
did a bit of final fine-tuning and merged in master

thanks for the PR!

PR #253, opened 2021-10-18
Issue154 Refer to openeo.org in auth doc

jonathom:
So far this is what could be safely removed (with keeping some duplicates so users won't have to constantly switch tabs). If you wanna exclude auth, client and resource code completely, let me know.


soxofaan:
This is a pretty old PR and I think it's best to close it.

I also think that duplication of this kind of documentation is actually not as bad as it is with source code in general. Avoiding  redirection to other docs on other websites avoids confusion and loss of focus. 

In any case, I also created #608 for further fine-tuning of the auth docs

Issue #246, opened 2021-10-05
simplify specifying collection properties


jdries:
Currently, in load_collection, property filtering only seems to work with callbacks:

```
    s2 = connection.load_collection(s["SENTINEL2"],                                 
                                temporal_extent=temp_ext,
                                spatial_extent=spat_ext,
                                bands=["B03","B04","B05","B06","B07","B08","B11","B12","SCL"],
                                properties={"cloudcover":lambda x: x.eq(90)}
```

This increase the learning curve somewhat, and a lot of property filters are based on equality. So something like:

```    
    properties={"cloudcover":90}
```
could be translated into a correct callback by the client?

m-mohr:
What would `properties={"cloudcover":90, "platform": "S1a"}` translate to? Are they connected with "and" or "or"?

Long-term it might make sense to adopt something like CQL text as a simplified version instead of objects, which are more human-readable and flexible (e.g. "cloudcover<90 and platform = 's1a'").

PS: cloud cover equality is probably not the best example...

m-mohr:
CQL2 text would be exactly that, which then you can also re-use for STAC API queries etc.

soxofaan:
> or introduce a minimal (SQL-like) query language

such a mini language could even be useful in other places, e.g.:

```python
cube.reduce_dimension("bands", reducer=" (B08 - B05) / (B08 - B05)")

mask = cube.apply("x == 4")
```

that's actually a generalization of what's under construction under  #255

soxofaan:
In #331  I experimented with this approach:

```python
from openeo import collection_property

cube = con100.load_collection(
    "S2",
    properties=[
        collection_property("eo:cloud_cover") <= 75,
        collection_property("platform") == "Sentinel-2B",
    ]
)
```

m-mohr:
I like that, although could "collection_property" be a bit shorter? :-)

soxofaan:
https://github.com/Open-EO/openeo-python-client/pull/331 was already merged (Nov 2023)

But it only works with `load_collection`. 
Should also be done for `load_stac` (stumbled on that today)

soxofaan:
737c90d adds support for `collection_property` in `load_stac`

soxofaan:
I think we can close this ticket now.
There are some other ideas here, but I guess there is little incentive to work on these at the moment, giving the existing solutions.
Feel free to reopen or create new tickets.

PR #255, opened 2021-11-03
calculate indices in client using one-liner

Bart92:
Die testen moet ik echt nog aanpassen, maar ik weet niet zo goed hoe.
Ik heb gemerkt dat met die equals in de test_get_expression_map ik eender welke waarde in kan geven voor array_create en in alle gevallen gaat de test groen markeren, wat te maken zou moeten hebben met de implementatie van de __eq__ (die eq aanroept, die er in de client ok uitziet).
Heb jij nog ideeën hoe ik die testen zou kunnen verbeteren?

### DIFF
--- a/openeo/extra/spectral_indices/__init__.py
+++ b/openeo/extra/spectral_indices/__init__.py
@@ -1,5 +1,13 @@
 """
-Calculate spectral indices (vegetation indices, water indices, etc.) in one line
+Easily calculate spectral indices (vegetation, water, urban etc.).
+
+Supports the indices defined in the
+`Awesome Spectral Indices <https://awesome-ee-spectral-indices.readthedocs.io/>`_ project
+by `David Montero Loaiza <https://github.com/davemlz>`_.
+
+
+.. versionadded:: 0.9.1
+
 """
 
 from openeo.extra.spectral_indices.spectral_indices import *
--- a/openeo/extra/spectral_indices/spectral_indices.py
+++ b/openeo/extra/spectral_indices/spectral_indices.py
@@ -144,12 +144,13 @@ def compute_and_rescale_indices(datacube: DataCube, index_dict: dict, append=Fal
                 }
             }
 
+        If you don't want to rescale your data, you can fill the input-, index- and output-range with ``None``.
+
         See `list_indices()` for supported indices.
 
-    .. note:: this "rescaled" index version uses an experimental API (e.g. `index_dict` argument) that is subject to change.
-
-    If you don't want to rescale your data, you can fill the input-, index- and output range with None.
     :return: the datacube with the indices attached as bands
+
+    .. warning:: this "rescaled" index helper uses an experimental API (e.g. `index_dict` argument) that is subject to change.
     """
     index_specs = load_indices()
 
@@ -185,9 +186,9 @@ def append_and_rescale_indices(datacube: DataCube, index_dict: dict) -> DataCube
 
         See `list_indices()` for supported indices.
 
-    .. note:: this "rescaled" index version uses an experimental API (e.g. `index_dict` argument) that is subject to change.
-
     :return: data cube with appended indices
+
+    .. warning:: this "rescaled" index helper uses an experimental API (e.g. `index_dict` argument) that is subject to change.
     """
     return compute_and_rescale_indices(datacube=datacube, index_dict=index_dict, append=True)
 
@@ -233,6 +234,7 @@ def compute_index(datacube: DataCube, index: str) -> DataCube:
     :param index: name of the index to compute. See `list_indices()` for supported indices.
     :return: data cube containing the index as band
     """
+    # TODO: option to compute the index with `reduce_dimension` instead of `apply_dimension`?
     return compute_indices(datacube=datacube, indices=[index], append=False)
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add `options` argument to `DataCube.atmospheric_correction` ([Open-EO/openeo-python-driver#91](https://github.com/Open-EO/openeo-python-driver/issues/91))
 - Add `atmospheric_correction_options` and `cloud_detection_options` arguments to `DataCube.ard_surface_reflectance` ([Open-EO/openeo-python-driver#91](https://github.com/Open-EO/openeo-python-driver/issues/91))
 - UDP storing: add support for "returns", "categories", "examples" and "links" properties ([#242](https://github.com/Open-EO/openeo-python-client/issues/242))
+- Add `openeo.extra.spectral_indices`: experimental API to easily compute spectral indices (vegetation, water, urban, ...)
+  on a `DataCube`, using the index definitions from [Awesome Spectral Indices](https://awesome-ee-spectral-indices.readthedocs.io/en/latest/)
 
 
 ### Changed
--- a/docs/cookbook/index.rst
+++ b/docs/cookbook/index.rst
@@ -8,3 +8,4 @@ OpenEO CookBook
    ard
    sampling
    udp_sharing
+   spectral_indices
--- a/docs/cookbook/spectral_indices.rst
+++ b/docs/cookbook/spectral_indices.rst
@@ -1,0 +1,9 @@
+====================================
+Spectral Indices
+====================================
+
+.. warning::
+    This is a new experimental API, subject to change.
+
+.. automodule:: openeo.extra.spectral_indices
+    :members: list_indices, compute_and_rescale_indices, append_and_rescale_indices, compute_indices, append_indices, compute_index, append_index


openeo/extra/spectral_indices/__init__.py
```diff
@@ -0,0 +1,15 @@
...
+and further supplemented with indices that were necessary for use cases
+"""
+
+class BaseOpenEoException(Exception):
+    pass
+
+from openeo._version import __version__
+
+def client_version() -> str:
+    return __version__
```
soxofaan, line 15: Why is this exception and version stuff defined here? seems unused

openeo/extra/spectral_indices/spectral_indices.py
```diff
@@ -0,0 +1,99 @@
...
+        "LANDSAT4-5_TM_L1": lookupL457,
+        "LANDSAT4-5_TM_L2": lookupL457,
+        "LANDSAT8_L1": lookupL8,
+        "LANDSAT8_L2": lookupL8,
+        "MODIS": lookupModis,
+        "PROBAV_L3_S5_TOC_100M": lookupProbav,
+        "PROBAV_L3_S10_TOC_333M": lookupProbav,
+        "SENTINEL2_L1C_SENTINELHUB": lookupS2,
+        "SENTINEL2_L2A_SENTINELHUB": lookupS2,
+        "TERRASCOPE_S2_TOC_V2": lookupS2,
```
soxofaan, line 53: I think we should avoid hardcoding names like SENTINELHUB and TERRASCOPE here

openeo/extra/spectral_indices/spectral_indices.py
```diff
@@ -0,0 +1,99 @@
+from openeo.processes import ProcessBuilder, array_modify
+from openeo.rest.datacube import DataCube
+import numpy as np
+
+def _get_expression_map(cube: DataCube, x: ProcessBuilder):
+    collection_id = cube.graph["loadcollection1"]["arguments"]["id"]
```
soxofaan, line 6: `loadcollection1` is not a very robust way to get collection id. It can easily be `loadcollection2` or something else.

I think it's more reliable to look into the datacube's metadata attribyte

openeo/extra/spectral_indices/spectral_indices.py
```diff
@@ -0,0 +1,99 @@
...
+    bands = [band.replace("0", "").upper() for band in cube.metadata.band_names]
+
+    def check_validity():
+        if not all(band in orig_bands for band in bands):
+            raise ValueError(
+                "The bands in your cube {} are not all Sentinel-2 bands (the following are: {})".format(bands,orig_bands))
```
soxofaan, line 12: error mentions "Sentinel-2" but if I understand correctly there is also support for Landsat, probaV, Modis, ...

openeo/extra/spectral_indices/spectral_indices.py
```diff
@@ -0,0 +1,99 @@
+from openeo.processes import ProcessBuilder, array_modify
+from openeo.rest.datacube import DataCube
+import numpy as np
+
+def _get_expression_map(cube: DataCube, x: ProcessBuilder):
+    collection_id = cube.graph["loadcollection1"]["arguments"]["id"]
+    bands = [band.replace("0", "").upper() for band in cube.metadata.band_names]
+
+    def check_validity():
```
soxofaan, line 9: check_validity is not used as far as I understand

openeo/extra/spectral_indices/spectral_indices.py
```diff
@@ -0,0 +1,99 @@
...
+        "LANDSAT8_L2": lookupL8,
+        "MODIS": lookupModis,
+        "PROBAV_L3_S5_TOC_100M": lookupProbav,
+        "PROBAV_L3_S10_TOC_333M": lookupProbav,
+        "SENTINEL2_L1C_SENTINELHUB": lookupS2,
+        "SENTINEL2_L2A_SENTINELHUB": lookupS2,
+        "TERRASCOPE_S2_TOC_V2": lookupS2,
+    }
+
+    if collection_id not in list(lookupPlatform.keys()):
```
soxofaan, line 56: you can just do `if collection_id not in lookupPlatform`  (so without `keys()` and `list()`)

openeo/extra/spectral_indices/spectral_indices.py
```diff
@@ -0,0 +1,99 @@
...
+        raise Exception("Sorry, satellite platform not supported for index computation!")
+
+    orig_bands, norm_bands = lookupPlatform[collection_id]()
```
soxofaan, line 59: this whole setup looks bit overly complicated:
- define `lookupX()` functions that return static list
-  put `lookupX()` functions in dict
- get desired `lookupX()` from dict
- call `lookupX()` to get     `orig_bands, norm_bands`
- use `dict(zip(orig_bands, norm_bands))`  to convert to mapping


I think you can just define a dictionary that holds the orig-to-common band name mappings

openeo/extra/spectral_indices/resources/spectral-indices-dict.json
```diff
@@ -0,0 +1,1344 @@
...
+                "S1"
+             ],
+            "contributor": "vito",
+            "date_of_addition": "2021-10-27",
+            "formula": "exec('import numpy as np') or exec('from openeo.processes import clip') or np.arccos(clip((( np.sqrt( (0.8328 - 0.6646)**2 + (N - R)**2 )**2 + np.sqrt( (1.610 - 0.8328)**2 + (S1 - N)**2 )**2 - np.sqrt( (1.610 - 0.6646)**2 + (S1 - R)**2 )**2 ) / (2 * np.sqrt( (0.8328 - 0.6646)**2 + (N - R)**2 ) * np.sqrt( (1.610 - 0.8328)**2 + (S1 - N)**2 ))), -1,1)) * (1. / np.pi)",
+            "long_name": "Angle at Near InfraRed",
+            "reference": "",
+            "short_name": "ANIR",
+            "type": "vegetation",
+            "range": "(0,1)"
```
soxofaan, line 17: is this `range` a customization? I can't find it in original https://github.com/davemlz/eemont/blob/master/eemont/data/spectral-indices-dict.json

openeo/extra/spectral_indices/resources/spectral-indices-dict.json
```diff
@@ -0,0 +1,1344 @@
...
+                "R"
+            ],
+            "contributor": "https://github.com/davemlz",
+            "date_of_addition": "2021-04-07",
+            "formula": "(N - R)/(N + R)",
+            "long_name": "Normalized Difference Vegetation Index",
+            "reference": "https://doi.org/10.1016/0034-4257(79)90013-0",
+            "short_name": "NDVI",
+            "type": "vegetation",
+            "range": "(0,1)"
```
soxofaan, line 771: shouldn't range of NDVI be `-1, 1`?

openeo/extra/spectral_indices/spectral_indices.py
```diff
@@ -0,0 +1,99 @@
...
+    params = _get_expression_map(datacube, x)
+    for index_name in index_list:
+        if index_name not in index_specs.keys():
+            raise NotImplementedError("Index " + index_name + " has not been implemented.")
+        index_result = eval(index_specs[index_name]["formula"], params)
+        if scaling_factor is not None:
+            if "range" not in index_specs[index_name].keys():
+                raise ValueError(
+                    "You want to scale the " + index_name + ", however the range of this index has not been supplied in the indices json yet.")
+            index_result = index_result.linear_scale_range(*eval(index_specs[index_name]["range"]), 0, scaling_factor)
```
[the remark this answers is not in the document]
soxofaan: If I understand correctly, if I use `scale_factor=20` on on an index with range `-5,5`, the output will be range `0,20`  instead of `-100, 100`  which I would expect from a argument called "scale factor"

openeo/extra/spectral_indices/spectral_indices.py
```diff
@@ -0,0 +1,99 @@
...
+        index_values.append(index_result)
+    if scaling_factor is not None:
```
soxofaan, line 78: wait, you have again `if scaling_factor is not None`  with `linear_scale_range`? why is that?

openeo/extra/spectral_indices/spectral_indices.py
```diff
@@ -0,0 +1,99 @@
...
+        x_res = x_res.linear_scale_range(0, 8000, 0, scaling_factor)
+    return array_modify(data=x_res, values=index_values, index=len(datacube.metadata.band_names))
+
+
+def compute_indices(datacube: DataCube, index_list: list, scaling_factor: int = None) -> DataCube:
```
soxofaan, line 83: maybe also add a function to calculate a single index (instead of putting multiple in an array)?

tests/test_spectral_indices.py
```diff
@@ -0,0 +1,34 @@
+from unittest.mock import Mock
+from openeo.extra.spectral_indices.spectral_indices import _callback, _get_expression_map
+from processes import array_create
+
+x = array_create([9,4,6,1,3,8,2,5,7,11,7,2])
```
soxofaan, line 5: you should not define global variables like this for tests, you should recreate your input for each test separately

tests/test_spectral_indices.py
```diff
@@ -0,0 +1,34 @@
...
+cube = Mock()
```
soxofaan, line 6: it should not be necessary to mock a cube.


In test_datacube100.py you can find loads of examples where a datacube is constructed from a mock collection, but the datacube is a real datacube.

tests/test_spectral_indices.py
```diff
@@ -0,0 +1,34 @@
...
+    "NDRE5": {
+        "bands": ["RE1","RE3"],
+        "formula": "(RE3 - RE1) / (RE3 + RE1)",
+        "range": "(-1,1)"
+    }
+}
+
+
+def test_get_expression_map():
+    assert _get_expression_map(cube, x)["A"] == array_create([9])
```
soxofaan, line 29: I think you should create the process builder here with `x = ProcessBuilder({"from_parameter": "data"})`  instead of doing `x = array_create([9,4,6,1,3,8,2,5,7,11,7,2])`, which will probably not work

tests/test_spectral_indices.py
```diff
@@ -0,0 +1,34 @@
...
+
+def test_callback():
+    index_list = ["NDRE1","NDGI","NDRE5"]
+    scaling_factor = None
+    assert _callback(x, index_list, cube, scaling_factor, index_specs)
```
soxofaan, line 34: this assert is not doing anything at the moment (except checking that `_callback` doesn't return None)

soxofaan (commented):
first round of review notes :sweat_smile:

m-mohr:
I found this via #246, but I can't understand what it's about. Could you translate it, please? ;-)

soxofaan:
> Could you translate it, please? ;-)

The original post is not really explaining the PR, it's mainly wondering about how to write the unit tests. 


But in a nutshell, this PR adds a helper module to easily calculate the spectral indices defined in https://github.com/davemlz/eemont/blob/master/eemont/data/spectral-indices-dict.json:   
user specifies that they want  for example index "ARVI", 
the python client looks up the formula for that : 
```
"formula": "(N - (R - gamma * (R - B))) / (N + (R - gamma * (R - B)))",
```
and translates that into a process graph (based on band name mapping)

soxofaan:
maybe this is a better reference to this spectral indices repo: https://github.com/davemlz/awesome-spectral-indices

m-mohr:
Thanks, this is interesting. It seems that this is not a general "band math parser", right?

openeo/extra/spectral_indices/spectral_indices.py
soxofaan: I think the point is about just to detect if a collection is  Sentinel1, Sentinel2, landsat, modis, ...? The exact collection_id is not that relevant.

In most cases you can guess this pretty well from the collection name (e.g. `"S2" in collection_id or "SENTINEL2" in collection_id`),
but even better would be using explicit collection metadata like "constellation" or "platform"

m-mohr: All these lookups seem mostly unnecessary if we'd properly fill the metadata with the common names as defined by STAC and the indices repo. They look very similar and then you could simply read the bands from the metadata without any hardcoding, I think.

soxofaan: this is about building the mapping from  band names from collection metadata to the names used in the formulas in that "spectral indices json file" (which is bit different from common names), so you need some degree of hardcoding I'm afraid

m-mohr: Hmm, as far as I understoof https://github.com/davemlz/awesome-spectral-indices#expressions all expressions consist of the "Standard" values, which seem to be somewhat aligned with https://github.com/stac-extensions/eo#common-band-names - I think the common names would make this work (more or less) automatically, similarly to the NDVI process in openEO, which implements the same idea.

openeo/extra/spectral_indices/resources/spectral-indices-dict.json
Bart92: Yes, I need that for rescaling in compute_indices

openeo/extra/spectral_indices/spectral_indices.py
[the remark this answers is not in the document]
Bart92: Yes you understood that correctly, and I will change the param name to something else to make this clearer

openeo/extra/spectral_indices/spectral_indices.py
Bart92: this is for the normal bands , which have a different rescaling

openeo/extra/spectral_indices/spectral_indices.py
Bart92: a single index can already be calculated by just putting a single one in an array (e.g. ["NDVI"] ) or would you still prefer me to add one ?

openeo/extra/spectral_indices/__init__.py
```diff
@@ -0,0 +1,15 @@
+"""
+Calculate spectral indices (vegetation indices, water indices, etc.) in one line
+
```
soxofaan, line 3: FYI: you also need an (empty) __init__.py in the extra folder

openeo/extra/spectral_indices/spectral_indices.py
```diff
@@ -0,0 +1,76 @@
...
+    def get_params():
+        return {band_mapping[key]: x.array_element(i) for i, key in enumerate(bands)}
+
+    landsat457_mapping = {"B1":"B", "B2":"G", "B3":"R", "B4":"N", "B5":"S1", "B6":"T1", "B7":"S2"}
+    landsat8_mapping = {"B1":"A", "B2":"B", "B3":"G", "B4":"R", "B5":"N", "B6":"S1", "B7":"S2", "B10":"T1", "B11":"T2"}
+    modis_mapping = {"B3":"B", "B4":"G", "B1":"R", "B2":"N", "B5":np.nan, "B6":"S1", "B7":"S2"}
+    probav_mapping = {"BLUE":"B", "RED":"R", "NIR":"N", "SWIR":"S1"}
+    sentinel2_mapping = {"B1":"A", "B2":"B", "B3":"G", "B4":"R", "B5":"RE1", "B6":"RE2", "B7":"RE3", "B8":"N", "B8A":"RE4", "B9":"WV", "B11":"S1", "B12":"S2"}
+
+    if "LANDSAT8" in collection_id:
```
soxofaan, line 23: make these ifs case insensitive

openeo/extra/spectral_indices/spectral_indices.py
```diff
@@ -0,0 +1,76 @@
...
+
+    param datacube: an instance of openeo.rest.DataCube
+    param index_list: a list of indices. The indices that are implemented are derived from the eemont package created by davemlz:
+    https://github.com/davemlz/eemont/blob/master/eemont/data/spectral-indices-dict.json
+    param uplim_rescale: the upper range to which you want to scale the value (the result after rescaling will be [0,uplim_rescale])
+    return: the datacube with the indices attached as bands
+
+    """
+    with open("resources/spectral-indices-dict.json") as f:
+        index_specs = eval(f.read())["SpectralIndices"]
```
soxofaan, line 72: oh funky, this should just be a  `json.load` instead of `eval` :smile:

openeo/extra/spectral_indices/spectral_indices.py
```diff
@@ -0,0 +1,76 @@
...
+                raise ValueError(
+                    "You want to scale the " + index_name + ", however the range of this index has not been supplied in the indices json yet.")
+            index_result = index_result.linear_scale_range(*eval(index_specs[index_name]["range"]), 0, uplim_rescale)
+        index_values.append(index_result)
+    if uplim_rescale is not None:
+        x_res = x_res.linear_scale_range(0, 8000, 0, uplim_rescale)
+    return array_modify(data=x_res, values=index_values, index=len(datacube.metadata.band_names))
+
+
+def compute_indices(datacube: DataCube, index_list: list, uplim_rescale: int = None) -> DataCube:
```
soxofaan, line 65: this `compute_indices` works with  `apply_dimension`  where multiple indices are appended to the cube

I think user would also expect other approaches:
- compute multiple indices but disregard all original bands
- compute single index with `reduce_dimension`  (e.g. how all the original NDVI examples are done)

openeo/extra/spectral_indices/spectral_indices.py
soxofaan: see my other comment about the use case to calculate a single index with `reduce_dimension`, instead of appending a list to an existing cube with `apply_dimension`

openeo/extra/spectral_indices/__init__.py
soxofaan: done in d9ae22f45d9933105a9fbffeac85d16cd50bd8bd

tests/test_spectral_indices.py
```diff
@@ -0,0 +1,43 @@
+from unittest.mock import Mock, MagicMock
```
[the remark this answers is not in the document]
soxofaan: added new test file like that in d9ae22f45d9933105a9fbffeac85d16cd50bd8bd FYI

I think you can remove this `tests/test_spectral_indices.py` unless there is something to keep

openeo/extra/spectral_indices/spectral_indices.py
```diff
@@ -0,0 +1,82 @@
...
+            raise NotImplementedError("Index " + index_name + " has not been implemented.")
+        index_result = eval(index_specs[index_name]["formula"], params)
+        if uplim_rescale is not None:
+            if "range" not in index_specs[index_name].keys():
+                raise ValueError(
+                    "You want to scale the " + index_name + ", however the range of this index has not been supplied in the indices json yet.")
+            index_result = index_result.linear_scale_range(*eval(index_specs[index_name]["range"]), 0, uplim_rescale)
+        index_values.append(index_result)
+    if uplim_rescale is not None:
+        x_res = x_res.linear_scale_range(0, 8000, 0, uplim_rescale)
```
soxofaan, line 61: where does this 8000 come from?

soxofaan: Overall, I'm still confused by this uplim_rescale feature:
- why rescaling the original band values too? And why assuming the input range is 0-8000? 
- for what use case is this rescaling actually necessary? I guess it's for data preparation reasons in a machine learning context, but I'm wondering if there aren't other solutions that would avoid handling rescaling inside  `compute_indices`

openeo/extra/spectral_indices/resources/spectral-indices-dict.json
soxofaan: also:  avoid strint `"(0,1)"`, just specify an array with two elements

PR #257, opened 2021-11-18
EP-4067: add drop_dimension method to CollectionMetadata

laxiwuka:


openeo/metadata.py
```diff
@@ -413,6 +413,15 @@ def add_dimension(self, name: str, label: Union[str, float], type: str = None) -
...
             dim = Dimension(type=type or "other", name=name)
         return self._clone_and_update(dimensions=self._dimensions + [dim])
+
+    def drop_dimension(self, type: str = None) -> 'CollectionMetadata':
+        """Drop dimension of a certain type"""
+        if type == "bands":
+            return self._clone_and_update(dimensions=[d for d in self._dimensions if not isinstance(d, BandDimension)])
+        elif type == "temporal":
+            return self._clone_and_update(dimensions=[d for d in self._dimensions if not isinstance(d, TemporalDimension)])
+        else:
```
m-mohr, line 423: I don't think this implementation does what it should. The second parameter is a name, not a type in the process spec. So it should always just remove a single dimension, not all of the same type.

soxofaan: indeed, the first argument (after `self`) should be `name`.

(Dropping by type could still be an additional client side feature though)

soxofaan:
I think this PR should also add `drop_dimension`  to the `DataCube` class, not just the metadata

openeo/metadata.py
m-mohr: > (Dropping by type could still be an additional client side feature though)

This can be tricky and sometimes lead to unexpected results though. Also, some types are missing here.

soxofaan: also see https://processes.openeo.org/#drop_dimension and https://github.com/Open-EO/openeo-processes/issues/305: it's probably better to use argument name `name`

soxofaan:
thanks!

Issue #269, opened 2022-01-26
Add describe_process + Vue component


m-mohr:
Add a describe_process function to the Python client as we also have it in the other clients. It should render the process nicely in Jupyter.

PR #270, opened 2022-01-26
Add describe_process #269

m-mohr:
Implements #269

### DIFF
--- a/openeo/rest/connection.py
+++ b/openeo/rest/connection.py
@@ -719,6 +719,24 @@ class Connection(RestApiConnection):
             processes = self.get('/processes/' + namespace, expected_status=200).json()["processes"]
         return VisualList("processes", data=processes, parameters={'show-graph': True, 'provide-download': False})
 
+    def describe_process(self, id: str, namespace: str = None) -> dict:
+        """
+        Retuens a single process from the back end.
+
+        :param id: The id of the process.
+        :param namespace: The namespace of the process.
+
+        :return: process: Dict The process definition.
+        """
+
+        processes = self.list_processes(namespace)
+        for process in processes:
+            if process["id"] == id:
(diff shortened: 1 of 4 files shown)


clausmichele:
As soon as I have time I will test it! Thanks

openeo/rest/connection.py
```diff
@@ -719,6 +719,24 @@ def list_processes(self, namespace: str = None) -> List[dict]:
             processes = self.get('/processes/' + namespace, expected_status=200).json()["processes"]
         return VisualList("processes", data=processes, parameters={'show-graph': True, 'provide-download': False})
 
+    def describe_process(self, id: str, namespace: str = None) -> dict:
+        """
+        Retuens a single process from the back end.
```
soxofaan, line 724: typo "Retuens"

openeo/rest/connection.py
```diff
@@ -719,6 +719,24 @@ def list_processes(self, namespace: str = None) -> List[dict]:
...
+
+        :param id: The id of the process.
+        :param namespace: The namespace of the process.
+
+        :return: process: Dict The process definition.
```
soxofaan, line 729: I know you copied this from `list_processes` (were it is also wrong), but this `Dict` does not belong there, the doc generator tool gets that from the function return type hint. It should just be

> :return: The process definition.

soxofaan (commented):
looks great (just some small doc issues)

openeo/rest/connection.py
m-mohr:
```suggestion
        :return: The process definition.
```

openeo/rest/connection.py
m-mohr:
```suggestion
        Returns a single process from the back end.
```

m-mohr:
Thanks, @soxofaan. I've fixed the points you have raised.

Issue #276, opened 2022-02-21
minor: Connection.create_job vs DataCube.send_job


soxofaan:
`Connection.create_job` vs `DataCube.send_job`
having differently named methods for same thing in different context is bit confusing when depending on code completion feature.

Let's settle on `create_job` (because that is also the verb used in openEO api)?

soxofaan:
Issue is not so minor, it would be good if it would be resolved before improving VectorCube and MlModel support (#278, #279), so that these new classes don't have to drag in deprecated usage patterns from the start

soxofaan:
done

PR #283, opened 2022-03-15
replace send_job with create_job

soxofaan:
self-review PR for issue #276


soxofaan:
@jdries   are you ok with deprecating `DataCube.send_job()` in favor of `DataCube.create_job()`  (to be more consistent internally and with other libraries/documentation) ?
Legacy `send_job()`  calls will still work, but trigger a deprecation warning.

It would help to settle this before further finetuning the `VectorCube` and `MlModel` classes further

jdries:
Yes, agreed, fine to merge!

Issue #237, opened 2021-09-13
OIDC device flow: visual indication that notebook is waiting


soxofaan:
When doing the OIDC device flow in a notebook (or even as a regular script), the `con.authenticate_oidc()` call (or related) will print something like 
```
To authenticate: visit https://aai.egi.eu/oidc/device and enter the user code 'F0oB6r'.
```
and then do noting (except waiting).

It would be cool if there was a more visual indication that the client code is waiting for the user to do something: e.g. some kind of spinner, progress bar (based on waiting timeout), ...

soxofaan:
problem with current implementation: output  text widget in jupyter notebook is now refreshed every 1, which makes it hard/impossible to select the url or code for copy pasting, which is a usability regression

soxofaan:
Also to do: progress bar should be cleaned up (or a new line should be added), now it looks like this:
```
To authenticate: visit https://aai.egi.eu/oidc/device and enter the user code '.....'.
Authorized successfully.   ] Waiting for authentication.             
Authenticated using device code flow.
```

soxofaan:
implemented in #420

PR #297, opened 2022-05-04
visual waiting indicator for OIDC device code flow

soxofaan:
Initial POC for #237

### DIFF
--- a/openeo/rest/auth/oidc.py
+++ b/openeo/rest/auth/oidc.py
@@ -10,6 +10,7 @@ import hashlib
 import http.server
 import json
 import logging
+import math
 import random
 import string
 import threading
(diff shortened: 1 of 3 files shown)


soxofaan:
played a bit with `IPython.display` in jupyter and this seems to be a better POC :
```python
import time

import IPython.display

url = "https://example.com"
html = f"""<p>
    To authenticate: visit <a href="{url}">{url}</a> 
    <a href="#" onclick="navigator.clipboard.writeText('{url}');return false;" title="Copy to clipboard">&#128203;</a>
    </p>"""

IPython.display.display_html(html, raw=True)
progress_bar = IPython.display.display({"text/html":" "}, raw=True, display_id=True)
for i in range(50):
    progress_bar.update({"text/html": "=" * i}, raw=True)
    time.sleep(0.1)
```

- has clickable link
- has a "copy to clipboard" button
-  update of the progress bar does not ruin copy paste functionality of the url

soxofaan:
overruled by #420

Issue #298, opened 2022-05-04
OIDC: automatically use refresh token on access token expiry 


soxofaan:
after OIDC authentication,  the connection object should keep track of the expiry of the access token and get a new one through refresh token when necessary. Now the user has to explicitly re-authenticate, while this is not really necessary.

This should make long running job scenario's easier.

PR #301, opened 2022-05-06
Issue #298: try to refresh access token when expired

soxofaan:
self-review PR for #298


soxofaan:
still to do:
- CHANGELOG entry
- unit tests

Issue #316, opened 2022-06-22
deprecated `distutils.version.LooseVersion`


soxofaan:
Noticed this in test warnings
```
tests/test_views_execute.py: 24 warnings
  /../openeo/openeo-python-client/openeo/capabilities.py:85: DeprecationWarning: distutils Version classes are deprecated. Use packaging.version instead.
    self._version = LooseVersion(version)
```

soxofaan:
see https://github.com/pypa/distutils/pull/75

soxofaan:
migrating to this `packaging` package means adding a new dependency, which we want to avoid if not necessary.
In this case it's probably not hard to  replace it with a DIY implementation (split version string in tuple of ints and use standard tuple comparison functionality)

iamkaushal:
@soxofaan I have created a PR for this issue. Please review it, this is my first open source contribution.

soxofaan:
#321  was merged

PR #321, opened 2022-07-02
custom LooseVersion replaces deprecated distutils.version.LooseVersion

iamkaushal:
resolved issue #316 by creating a similar LooseVersion class for the deprecated distutils.version.LooseVersion.

### DIFF
--- a/openeo/capabilities.py
+++ b/openeo/capabilities.py
@@ -3,89 +3,6 @@ from abc import ABC
 import re
 from typing import Union
 
-# Is this base class (still) useful?
-
-
-class LooseVersion:
-    """A version number consists of a series of numbers,
-    separated by either periods or strings of letters. When comparing
-    version numbers, the numeric components will be compared
-    numerically, and the alphabetic components lexically. The following
-    are all valid version numbers, in no particular order:
-        1.5.1
-        1.5.2b2
-        161
-        3.10a
-        8.02
-        3.4j
-        1996.07.12
-        3.2.pl0
-        3.1.1.6
-        2g6
-        11g
-        0.960923
-        2.2beta29
-        1.13++
-        5.5.kw
-        2.0b1pl0
-    In fact, there is no such thing as an invalid version number under
-    this scheme; the rules for comparison are simple and predictable.
-    """
-
-    component_re = re.compile(r'(\d+ | [a-z]+ | \.)', re.VERBOSE)
-
-    def __init__(self, version_string=None):
-        if isinstance(version_string, str):
-            self.parse(version_string)
-
-        if isinstance(version_string, LooseVersion):
-            self.version = version_string.version
-
-    def parse(self, version_string):
-        self.vstring = version_string
-        components = [x for x in self.component_re.split(version_string)
-                      if x and x != '.']
-        for i, obj in enumerate(components):
-            with contextlib.suppress(ValueError):
-                components[i] = int(obj)
-        self.version = tuple(components)
-
-    def __str__(self):
-        return str(self.vstring)
-
-    def __repr__(self):
-        return "LooseVersion ('%s')" % str(self)
-
-    def _cmp(self, other):
-        if isinstance(other, str):
-            other = LooseVersion(other)
-        elif not isinstance(other, LooseVersion):
-            return NotImplemented
-        if self.version == other.version:
-            return 0
-        if self.version < other.version:
-            return -1
-        if self.version > other.version:
-            return 1
-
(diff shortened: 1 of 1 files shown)


openeo/capabilities.py
```diff
@@ -1,10 +1,92 @@
...
 from abc import ABC
-from distutils.version import LooseVersion
+import re
 from typing import Union
 
 # Is this base class (still) useful?
 
 
+class LooseVersion:
+    """A version number consists of a series of numbers,
```
soxofaan, line 10: If I understand correctly this is a straight copy from the original `distutils.version.LooseVersion`.
I think that should be noted in the docstring.

Moreover it now looks unnecessary and overkill to have both LooseVersion and ComparableVersion defined here. I think it makes more sense to merge both classes to a single class

openeo/capabilities.py
```diff
@@ -1,10 +1,92 @@
...
+    """
+
+    component_re = re.compile(r'(\d+ | [a-z]+ | \.)', re.VERBOSE)
+
+    def __init__(self, version_string=None):
+        if isinstance(version_string, str):
+            self.parse(version_string)
+
+        if isinstance(version_string, LooseVersion):
+            self.version = version_string.version
```
soxofaan, line 42: there is no `else` here: what if `version_string=None` (the default)?

openeo/capabilities.py
```diff
@@ -1,10 +1,92 @@
...
+
+    def parse(self, version_string):
+        self.vstring = version_string
```
soxofaan, line 45: this method introduces `self.vstring` which is not defined in `__init__`

soxofaan (commented):
Thanks for this PR!
Sorry for the delay in reviewing this (I had to work through my backlog after my summer holidays)

openeo/capabilities.py
iamkaushal: fixed

openeo/capabilities.py
iamkaushal: fixed

openeo/capabilities.py
iamkaushal: Makes sense. Fixed.

iamkaushal:
Thank you for the review, I have made the changes, please review again.

openeo/capabilities.py
```diff
@@ -1,9 +1,8 @@
+import contextlib
 from abc import ABC
-from distutils.version import LooseVersion
+import re
 from typing import Union
 
-# Is this base class (still) useful?
```
soxofaan, line 5: This comment shouldn't be removed (it applies to Capabilities class)

openeo/capabilities.py
```diff
@@ -52,7 +51,33 @@ def list_plans(self):
 
 class ComparableVersion:
     """
-    Helper to compare a version (e.g. API version) against another (threshold) version
+    This code is copied from distutils.version.LooseVersion
```
soxofaan, line 54: Now after being merged, I would not say "This code is copied", but "implementation is based on"

openeo/capabilities.py
```diff
@@ -78,11 +103,21 @@ class ComparableVersion:
         `a.accept_lower(b)`: b is lower than a
     """
 
+    component_re = re.compile(r'(\d+ | [a-z]+ | \.)', re.VERBOSE)
```
soxofaan, line 106: I would make this a private attribute

openeo/capabilities.py
```diff
@@ -78,11 +103,21 @@ class ComparableVersion:
...
+
     def __init__(self, version: Union[str, 'ComparableVersion']):
         if isinstance(version, ComparableVersion):
             self._version = version._version
         else:
-            self._version = LooseVersion(version)
+            self.parse(version)
+
+    def parse(self, version_string):
```
soxofaan, line 114: I think this should be a (private) class method, that doesn't modify `self`, but just returns a tuple

soxofaan (commented):
Thanks for the updates!

some more fine-tune suggestions

openeo/capabilities.py
iamkaushal: done

openeo/capabilities.py
iamkaushal: fixed

soxofaan:
Thanks!

I did some final fine-tuning and merged in 17cbe08

iamkaushal:
@soxofaan  Awesome, it was my first open source contribution, thank you very much for helping me throughout the process.

PR #325, opened 2022-09-09
.download() refinement

clausmichele:
The sync call .download is currently always adding a save_result node even if the datacube contains it already.
This is my proposal to check firstly if save_result it's already in the graph and add it only if not present.


openeo/rest/datacube.py
```diff
@@ -1610,7 +1610,11 @@ def download(
         if not format:
             format = guess_format(outputfile) if outputfile else "GTiff"
         # TODO: only add `save_result` node when there is none yet?
```
soxofaan, line 1612: you can remove this TODO note now :)

openeo/rest/datacube.py
```diff
@@ -1610,7 +1610,11 @@ def download(
...
-        cube = self.save_result(format=format, options=options)
+        for p in self.flat_graph().keys():
+            if "save_result" in self.flat_graph()[p]["process_id"] and self.flat_graph()[p]["result"]:
+                cube = self
+            else:
+                cube = self.save_result(format=format, options=options)
```
soxofaan, line 1617: I think this can just be

```python
        cube = self
        if cube.result_node().process_id != "save_result":
            cube = cube.save_result(format=format, options=options)
```

soxofaan:
if you are familiar with that: can you also add unit tests about his?

clausmichele:
Ok I'll try!

soxofaan:
I tackled this one already in 9e7763191654048867336ed7644809e30afce394

Issue #296, opened 2022-05-02
warn about newer version of python client


soxofaan:
This sometimes comes up in discussions on hosted jupyterlab contexts with outdated `openeo` (python package): is it possible to inform the user that a newer version of the client is available?

One possible solution inspired by how `pip`, which shows a warning when it detects a newer version is abailbale:
```
$ pip install ...
WARNING: You are using pip version 20.1.1; however, version 22.0.4 is available.
You should consider upgrading via the '..../bin/python3.7 -m pip install --upgrade pip' command.
```

clausmichele:
Hi @soxofaan, I've tried to create a PR to address this issue. It actually check if the current version is the same as the one available online, not if is >=.

soxofaan:
quick note: with https://github.com/Open-EO/openeo-python-client/issues/176 the conda package is now also distributed through conda, which makes this "version warning" feature a bit more complex: you don't want to push users to pip install if the package was originally installed with conda. I'm not sure if it is possible to properly detect the installation context and emit the correct warning

soxofaan:
Just saw that jupyterlab also has a "newer version is available" notification. Implementation could be inspiration: https://github.com/jupyterlab/jupyterlab/blob/97f1c6f957f272113dce8157318d74c4fb23c289/jupyterlab/handlers/announcements.py#L78

Issue #176, opened 2021-02-01
Provide openeo package on conda-forge


soxofaan:
We provide openeo python client library already on pypi.org, but it could be beneficial to our adoption rate and installation friendliness to also provide it through conda on conda-forge channel.

I'm not sure at the moment what has to happen to make it possible, how much additional maintenance it would require and if it is all worth the trouble. Just putting it out here to collect some feedback, if any.

enolfc:
We do manage most packages with conda in the [EGI Notebooks](https://notebooks.egi.eu) service and having packages in conda would simplify the installation of OpenEO.

sebastian-luna-valero:
Hi,

I am happy to help create a conda package for `openeo-python-client`. Here is the first attempt:
https://github.com/conda-forge/staged-recipes/pull/15717

@soxofaan could you please have a look and provide feedback? Would you like to be included as a conda package maintaner?

Best regards,
Sebastian

jdries:
Hi @sebastian-luna-valero , thanks this is very helpfull, as we don't have experience with this ourselves!
Stefaan is on holiday, so he may not reply for a few weeks.
I had a look at the recipe, looks fine for me. You can add me as a maintainer as well.

sebastian-luna-valero:
Thanks @jdries 

Great, I will add you as a maintainer as well.

I have got `pytest` working on Linux only. Could you have a look at the errors for OSX and Windows [here](https://dev.azure.com/conda-forge/84710dde-1620-425b-80d0-4cf5baca359d/_build/results?buildId=354886&view=logs&jobId=e35d4f76-8ff2-5536-d795-df91e63eb9f7) and let me know your thoughts, please?

Best regards,
Sebastian

jdries:
Hi Sebastian,
seems much needed that we also have some windows and OSX testing.
I'm afraid these point to some actual issues, and seem to remember that they were not super easy to solve.
Is it ok if my colleague Stefaan has a look at this in august? 
Otherwise, we may need to come up with some temporary workaround...

sebastian-luna-valero:
Hi Jeroen,

I am happy to wait until Stefaan can have a look.

It is also possible to just build and release the conda package for Linux, and forget about OSX and Windows for now.

Best regards,
Sebastian

sebastian-luna-valero:
Hi,

FYI: I will be on leave the last two weeks of August, so I will be available to help with the conda package either next week or after 31st of August.

Best regards,
Sebastian

[...]

PR #327, opened 2022-09-14
Version warning

clausmichele:
Trying to solve this issue: https://github.com/Open-EO/openeo-python-client/issues/296

### DIFF
--- a/openeo/__init__.py
+++ b/openeo/__init__.py
@@ -37,7 +37,7 @@ def check_if_latest_version():
         return
     installed_version = client_version()
     if latest_version != client_version():
-        warnings.warn(f'WARNING: You are using {package} version {installed_version}; however, version {latest_version} is available. You should consider upgrading via the \'pip install --upgrade {package}\' command.', Warning)
+        warnings.warn(f'You are using {package} version {installed_version}; however, version {latest_version} is available. You should consider upgrading via the \'pip install --upgrade {package}\' command.', Warning)
     return
 
 check_if_latest_version()
--- a/openeo/__init__.py
+++ b/openeo/__init__.py
@@ -17,7 +17,8 @@ from openeo.rest.datacube import DataCube
 from openeo.rest.connection import connect, session, Connection
 from openeo.rest.job import BatchJob, RESTJob
 from openeo.internal.graph_building import UDF
-
+import requests
+import warnings
 
 def client_version() -> str:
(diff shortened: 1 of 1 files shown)


openeo/__init__.py
```diff
@@ -17,11 +17,28 @@ class BaseOpenEoException(Exception):
...
     try:
         import importlib.metadata
         return importlib.metadata.version("openeo")
     except Exception:
         return __version__
+
+def check_if_latest_version():
+    try:
+        package = 'openeo'
+        response = requests.get(f'https://pypi.org/pypi/{package}/json')
```
soxofaan, line 33: I think we should put a timeout here of just a couple of seconds so that users are not blocked to do `import openeo` when pypi is slow

soxofaan: I also think we should do this check like once a day, or cache the pypi.org response

openeo/__init__.py
```diff
@@ -17,11 +17,28 @@ class BaseOpenEoException(Exception):
...
+        latest_version = response.json()['info']['version']
+    except:
+        # Probably no internet connection available, pass
+        return
+    installed_version = client_version()
+    if latest_version != client_version():
```
soxofaan, line 39: Why a `!=`  check instead of `>` ?

for developers the local version could be ahead of the latest published version and you don't want this warning then

soxofaan (commented):
cool, thanks for starting this PR

openeo/__init__.py
soxofaan: for example, the pip implementation is roughly contained here: https://github.com/pypa/pip/blob/main/src/pip/_internal/self_outdated_check.py 
apparently the check frequency is once a week there

openeo/__init__.py
clausmichele: I'm using the ComparableVersion class from openeo.capabilities now.

openeo/__init__.py
clausmichele: I've added the timeout. The pip implementation using the cache would require some more time to be ported over.

soxofaan:
quick note: with https://github.com/Open-EO/openeo-python-client/issues/176 the conda package is now also distributed through conda, which makes this "version warning" a bit more complex: you don't want to push users to `pip install`  if the package was originally installed with conda.

Issue #328, opened 2022-09-20
Simplify setting (max) cloud coverage in load_collection


soxofaan:
Setting max cloud cover can be quite important when loading a collection (e.g. to avoid loading unusable raw data).
It's currently however quite cryptic to do  that with the python client:

```python
cube = connection.load_collection(
    ... , 
    properties={"eo:cloud_cover": lambda v: v == 80}
)
```

There should be a user friendly shortcut (e.g. a dedicated `Connection.load_collection` argument to lower the bar to use this, and promote usage.

soxofaan:
a1a35c9 added  `max_cloud_cover` argument to load_collection:
```python
cube = connection.load_collection(
        "S2",
        max_cloud_cover=75,
    )
```

PR #331, opened 2022-09-30
Add `collection_property` helper to simplify building property filters

soxofaan:
While working on #328, I thought of this proof of concept to simplify specifying simple collection property filters.

Usage example:

```python
from openeo import collection_property

cube = con100.load_collection(
    "S2",
    properties=[
        collection_property("eo:cloud_cover") <= 75,
        collection_property("platform") == "Sentinel-2B",
    ]
)
```

note that `properties` is now a list (instead of a dictionary),
containing these `collection_property` operator tricks that are translated to the proper process graphs

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -139,8 +139,10 @@ openeo.processes
 Graph building
 ----------------
 
+Various utilities and helpers to simplify the construction of openEO process graphs.
+
 .. automodule:: openeo.rest.graph_building
-    :members: collection_property
+    :members: collection_property, CollectionProperty
 
 .. automodule:: openeo.internal.graph_building
     :members: PGNode, FlatGraphableMixin
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -136,8 +136,11 @@ openeo.processes
     :members: process
 
 
-openeo.internal
+Graph building
 ----------------
+
+.. automodule:: openeo.rest.graph_building
+    :members: collection_property
 
 .. automodule:: openeo.internal.graph_building
     :members: PGNode, FlatGraphableMixin
--- a/openeo/__init__.py
+++ b/openeo/__init__.py
@@ -14,6 +14,7 @@ class BaseOpenEoException(Exception):
 from openeo._version import __version__
 from openeo.rest.connection import Connection, connect, session
 from openeo.rest.datacube import UDF, DataCube
+from openeo.rest.graph_building import collection_property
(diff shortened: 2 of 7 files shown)


soxofaan:
could also be considered a solution for #246

m-mohr:
I like that, although could "collection_property" be a bit shorter? :-) prop, field, property, key or so.

soxofaan:
It's indeed longer, but only  in number of characters. 
It's simpler in terms of symbols or concepts: just a function call and a comparison, while with the existing approach you have to understand how `lambda` works (which isn't the case for most beginners)

`collection_property` is indeed a bit long, but making it shorted like prop, field, property or key is risky because the user is expected (or will be tempted) to import it globally (`from openeo import thing`) and you don't want `thing` to collide with  existing globals. `property` is for example a python builtin global, `field` and `key` are very popular variable names with high chance on collision. `prop` is too short or cryptic in my opinion.

soxofaan:
I also expect that users are going to copy-paste  from example snippets (or use AI driven code generators in the future :stuck_out_tongue: ), so I don't think that the extra typing effort for `collection_property` is that much of an issue

m-mohr:
But if they copy/use AI anyway, then you can simply stick with `"eo:cloud_cover": lambda v: v <= 75` :-P

Another reason not to use collection_filter: We also have other processes that use the property filtering, which are not related to collections, e.g. filter_labels. So I'd argue `property` or `filter` or if you wanna namespace if `openeo_filter` are better choices... :-)

soxofaan:
`property` and `filter` are a no-go as they are builtin functions, you don't want a user to accidentally overwrite these. Namespacing like `openeo_filter`  looks a bit weird compared to the rest of the python API. 
`value_filter` could be an option, because that's what it is: a  helper to build a `{"from_parameter": "value"}` filter process graph.

However, I personally still like the concreteness of something like
```
load_collection(...
    properties=[
        collection_property("eo:cloud_cover") <= 75,
```

Another alternative is using the terminology "metadata property" as used in the description of `load_collection`:
```
load_collection(...
    properties=[
        metadata_property("eo:cloud_cover") <= 75,
```

m-mohr:
I just checked it against and my claim that filter_label also uses this is wrong. In all other "logical expressions" (filter_labels, count, aray_filter) you have direct values to work against and don't need the special handling.

In the schema ths subtype is called `metadata-filter` by the way.

PR #338, opened 2022-10-28
Client side processing

clausmichele:
@soxofaan I'm starting this PR to discuss how to properly implement the functionality required for client side processing.


openeo/rest/connection.py
```diff
@@ -1285,17 +1285,23 @@ def _config_log(message):
...
+        auth_type = auth_type.lower() if isinstance(auth_type, str) else auth_type
+        if auth_type in {None, False, 'null', 'none'}:
+            pass
+        elif auth_type == "basic":
+            connection.authenticate_basic(**(auth_options or {}))
+        elif auth_type in {"oidc", "openid"}:
+            connection.authenticate_oidc(**(auth_options or {}))
+        else:
+            raise ValueError("Unknown auth type {a!r}".format(a=auth_type))
+    except Exception as e:
```
soxofaan, line 1300: Switching to `LocalConnection`  through a `try-except` doesn't look very clean.

In this experimental phase of this feature, I would avoid touching `rest/connection.py` and related and work in an isolated submodule

openeo/rest/localconnection.py
```diff
@@ -0,0 +1,314 @@
...
+        b_dimension = {}
+        if len(bands)>0:
+            b_dimension = {'bands': {'type': 'bands', 'values':bands}}
+
+        metadata['cube:dimensions'] = {**t_dimension,**x_dimension,**y_dimension,**b_dimension}
+
+        return metadata
+    
+    def get_netcdf_collections(self):
+        local_collections_netcdfs = glob(self.local_collections_path + '/*.nc')
```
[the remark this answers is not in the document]
clausmichele: This could be achieved already by initializing the LocalConnection object with a different path. Or do you mean something different?

openeo/rest/datacube.py
```diff
@@ -1610,7 +1610,11 @@ def download(
         if not format:
             format = guess_format(outputfile) if outputfile else "GTiff"
         # TODO: only add `save_result` node when there is none yet?
-        cube = self.save_result(format=format, options=options)
+        for p in self.flat_graph().keys():
```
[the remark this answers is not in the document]
clausmichele: Probably a leftover, I will fix this thanks.

openeo/rest/connection.py
clausmichele: Ok, I will move to a different module instead of having it there. From `openeo/rest/connection.py` to `openeo/local/connection.py`

openeo/rest/localconnection.py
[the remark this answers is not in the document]
soxofaan: I mean that I have for example local netcdf files at `/data/S1/2022/*.nc`, `/data/S2/2021/*.nc`, `/data/S2/2022/*.nc`, and I want to be able to load these collections from the same `LocalConnection` object

openeo/rest/localconnection.py
```diff
@@ -0,0 +1,314 @@
...
+
+    def __init__(self,local_collections_path):
+        """
+        Constructor of LocalConnection.
+
+        :param local_collections_path: String path to the folder with the local collections in netCDF or geoTIFF
+        """
+        self.local_collections_path = local_collections_path.split('file://')[-1]
+        
+    def get_temporal_dimension(self,dims):
```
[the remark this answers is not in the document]
clausmichele: Ok, now they are private functions:
https://github.com/Open-EO/openeo-python-client/pull/338/commits/c703edf21f6ff29bbe6c1ee617558160e68fb16e

openeo/rest/localconnection.py
```diff
@@ -0,0 +1,314 @@
...
+            b_dimension = {'bands': {'type': 'bands', 'values':bands}}
+
+        metadata['cube:dimensions'] = {**t_dimension,**x_dimension,**y_dimension,**b_dimension}
+
+        return metadata
+    
+    def get_netcdf_collections(self):
+        local_collections_netcdfs = glob(self.local_collections_path + '/*.nc')
+        local_collections_list = []
+        if len(local_collections_netcdfs)>0:
```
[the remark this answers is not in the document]
clausmichele: Solved with this commit https://github.com/Open-EO/openeo-python-client/pull/338/commits/32e49333fb38b627d88e0d00133c1ade61460cd0

openeo/rest/connection.py
clausmichele: Removed the new local parts from `openeo/rest/connection.py` and introduced `openeo/local/connection.py` in this commit: https://github.com/Open-EO/openeo-python-client/pull/338/commits/20647753e5fd0df25d26e3251c9243142d7a4470

openeo/rest/localconnection.py
```diff
@@ -0,0 +1,314 @@
+import datetime
```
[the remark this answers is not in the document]
clausmichele: Solved moving to `/local/connection.py`

clausmichele:
@soxofaan after the last commits this is the current syntax:

```python
import openeo
local_conn = openeo.local.connection.LocalConnection("./results/")
local_conn.list_collections()
```

I would still like to simplify it a bit similarly to the `rest` part, so with something like `openeo.connect_local()` maybe?

clausmichele:
Specific requirements added here https://github.com/Open-EO/openeo-python-client/pull/338/commits/b3067cd75e30c045d650d8959623df780614f136. Now it can be installed using `pip install .[processing]`

setup.py
```diff
@@ -26,6 +26,12 @@
     'geopandas',
 ]
 
+processing_require = [
+    'rioxarray',
+    'pyproj',
+    'pathlib'
```
soxofaan, line 32: pathlib is not necessary I guess, it's part of standard library, right? or is this something special?

setup.py
```diff
@@ -55,7 +61,8 @@
               "sphinx-autodoc-typehints",
               "flake8",
               "myst-parser",
-          ]
+          ],
+          "processing": processing_require
```
soxofaan, line 65: I think "local" should be in this "extra" name. Just "processing" seems to suggest that a default install without extras does not allow processing.
e.g. "localprocessing", or even just "local"

openeo/local/connection.py
```diff
@@ -0,0 +1,316 @@
...
+from pyproj import Transformer
+from typing import Dict, List, Tuple, Union, Callable, Optional, Any, Iterator
+
+from openeo.metadata import CollectionMetadata
+from openeo.internal.graph_building import PGNode
+from openeo.rest.datacube import DataCube
+from openeo.internal.jupyter import VisualDict, VisualList
+
+def _get_temporal_dimension(dims):
+    if 't' in dims:
```
soxofaan, line 16: fyi you can do this more compactly like 

```python
for field in ["t", "time", "temporal", "DATA"]:
    if field in dims:
        return field
```
(explicit `None` return at the end is optional)

openeo/__init__.py
```diff
@@ -16,7 +16,7 @@ class BaseOpenEoException(Exception):
 from openeo.rest.datacube import DataCube, UDF
 from openeo.rest.connection import connect, session, Connection
 from openeo.rest.job import BatchJob, RESTJob
-
```
soxofaan, line 19: I think we should only expose `LocalConnection`  here once this feature has some maturity, otherwise we risk  to "vendor-lock-in" ourselves  in a corner

openeo/local/connection.py
```diff
@@ -0,0 +1,316 @@
+import datetime
+import xarray as xr
+import numpy as np
+import rioxarray
+from glob import glob
```
soxofaan, line 5: tip: if you are already using `pathlib`, I would just use `pathlib.Path.glob` instead of the old style `glob`

setup.py
clausmichele: yes, I was also thinking about `localprocessing`, but then it seemed too long. But it makes more sense since it is self explanatory. I will rename it into `localprocessing`.

openeo/local/connection.py
clausmichele: You are totally right, thanks for the hint.

openeo/local/connection.py
clausmichele: Thanks, it's on my list!

openeo/local/connection.py
soxofaan: you can make it even generic, instead of a separate function for each dimension:

```python
def _get_dimension(dims: dict, candidates: List[str]):
    for name in candidates:
        if name in dims:
            return name
....

x_dim = _get_dimension(data.dims, ["x", "X", "lon", "longitude"])
...
```

soxofaan:
> I would still like to simplify it a bit similarly to the rest part, so with something like openeo.connect_local() maybe?

No, I think ultimately the local processing could also be covered from the central entry point `openeo.connect()`. But for now, while we are still figuring things out, I would stay away from baking those things too early in there yet.

If you want to make it a bit more compact you could have 
```
```
with `openeo/local/__init__.py` like:
```
from openeo.local.connection import LocalConnection
```
That doesn't look too bad to me


> Edit: does it make sense to have two files with the same name? connection.py I am not sure if this could create some confusion.

I think that's fine. It will just be an implementation detail and the end user would/should not be confronted with that.

clausmichele:
@soxofaan I updated the sample notebook including a sample workflow of local processing using the recently released repos https://github.com/Open-EO/openeo-processes-dask and https://github.com/Open-EO/openeo-pg-parser-networkx

after Christmas we should have a meeting discussing the next steps to integrate it better.

openeo/rest/localconnection.py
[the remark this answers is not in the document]
clausmichele: Fixed now using `rglob` instead of `glob`: https://github.com/Open-EO/openeo-python-client/pull/338/commits/9d452edb9aef396fae3b706e291622a8a2a47adf

openeo/local/connection.py
```diff
@@ -0,0 +1,290 @@
...
+
+from openeo.metadata import CollectionMetadata
+from openeo.internal.graph_building import PGNode
+from openeo.rest.datacube import DataCube
+from openeo.internal.jupyter import VisualDict, VisualList
+
+def _get_dimension(dims: dict, candidates: List[str]):
+    for name in candidates:
+        if name in dims:
+            return name
```
soxofaan, line 17: you probably want to raise exception too if no candidate matched?

openeo/local/connection.py
```diff
@@ -0,0 +1,290 @@
...
+        
+def _get_netcdf_zarr_metadata(file_path):
```
soxofaan, line 19: all this netcdf and geotiff handling should probably move outside of "connection.py" module because it is not directly related to the connection object.
move it e.g. to "collections.py"?

clausmichele:
@soxofaan from our last meeting:

- Added exceptions for missing dims
- Allow multiple folders to be scanned for local collections 
- Sample test added
- Moved the local collections metadata parts to `collections.py`
- Added `execute()` method for `LocalCollection` (created a new `processing.py` to contain specific local processing code)
- Updated notebook with examples using `execute()`

soxofaan:
I also see that there is (still/again) a merge conflict on `setup.py` . Can you resolve that?

soxofaan:
I also noticed this is dependency-wise quite a heavy feature.
A basic install of openeo package pulls in just these 15 deps:
```
certifi==2022.12.7
charset-normalizer==3.0.1
Deprecated==1.2.13
idna==3.4
numpy==1.24.2
packaging==23.0
pandas==1.5.3
python-dateutil==2.8.2
pytz==2022.7.1
requests==2.28.2
shapely==2.0.1
six==1.16.0
urllib3==1.26.14
wrapt==1.14.1
xarray==2023.1.0
```

Install of the `localprocessing` extra adds 70 deps to that!
```
affine==2.4.0
attrs==22.2.0
cachetools==5.3.0
cftime==1.6.2
click==8.1.3
click-plugins==1.1.1
cligj==0.7.2
cloudpickle==2.2.1
contourpy==1.0.7
cycler==0.11.0
dask==2023.1.1
dask-geopandas==0.3.0
dask-image==2022.9.0
datacube==1.8.11
distributed==2023.1.1
Fiona==1.9.0
fonttools==4.38.0
fsspec==2023.1.0
GeoAlchemy2==0.13.1
geojson-pydantic==0.5.0
geopandas==0.12.2
greenlet==2.0.2
HeapDict==1.0.1
imageio==2.25.0
Jinja2==3.1.2
jsonschema==4.17.3
kiwisolver==1.4.4
lark==1.1.5
locket==1.0.0
MarkupSafe==2.1.2
matplotlib==3.6.3
msgpack==1.0.4
munch==2.5.0
netCDF4==1.6.2
networkx==2.8.8
numexpr==2.8.4
odc-algo==0.2.3
odc-geo==0.3.3
openeo-pg-parser-networkx==2023.1.2
openeo-processes-dask==2023.1.2
partd==1.3.0
pendulum==2.1.2
Pillow==9.4.0
PIMS==0.6.1
psutil==5.9.4
psycopg2==2.9.5
pydantic==1.10.4
pyparsing==3.0.9
pyproj==3.4.1
pyrsistent==0.19.3
pytzdata==2020.1
PyWavelets==1.4.1
PyYAML==6.0
rasterio==1.3.5.post1
rioxarray==0.13.3
ruamel.yaml==0.17.21
ruamel.yaml.clib==0.2.7
scikit-image==0.19.3
scipy==1.10.0
slicerator==1.1.0
snuggs==1.4.7
sortedcontainers==2.4.0
SQLAlchemy==1.4.46
tblib==1.7.0
tifffile==2023.2.3
toolz==0.12.0
tornado==6.2
typing_extensions==4.4.0
xgboost==1.7.3
zict==2.2.0
```

That's a very heavy addition (not only in number of packages, but also in terms of volume of compiled wheels to download). It would be good if this could be trimmed down.

clausmichele:
> I also see that there is (still/again) a merge conflict on `setup.py` . Can you resolve that?

Solved!

clausmichele:
@soxofaan how did you get the list of all the dependencies required by installing the extra `localprocessing`?

soxofaan:
> how did you get the list of all the dependencies required by installing the extra localprocessing?

I just played in temporary virtual envs and compared the list of installed package

> Anyway, the large dependencies number is not something can be addressed easily. 

of course there are necessary packages for local processing, but I also see a lot related to web dev that do not make sense just for this purpose, e.g.:
```
click==8.1.3
click-plugins==1.1.1
fonttools==4.38.0
greenlet==2.0.2
Jinja2==3.1.2
MarkupSafe==2.1.2
psutil==5.9.4
psycopg2==2.9.5
pyparsing==3.0.9
PyYAML==6.0
SQLAlchemy==1.4.46
tornado==6.2
```
(and I guess these drag in a number of smaller secondary dependencies as well)

tests/local/conftest.py
```diff
@@ -0,0 +1,34 @@
+import pytest
+import xarray as xr
+import numpy as np
+import pandas as pd
+
+@pytest.fixture()
+def create_local_netcdf(tmp_path_factory):
```
soxofaan, line 7: I don't think this has to be a fixture. I guess for tests you want to produce several kinds of netcdf/geotiff files with varying dimensions, label ranges, etc
so you just need a helper function

openeo/local/collections.py
```diff
@@ -0,0 +1,250 @@
...
+    b_dimension = {}
+    if len(bands)>0:
+        b_dimension = {'bands': {'type': 'bands', 'values':bands}}
+
+    metadata['cube:dimensions'] = {**t_dimension,**x_dimension,**y_dimension,**b_dimension}
+
+    return metadata
+
+def _get_geotiff_metadata(file_path):
+    data = rioxarray.open_rasterio(file_path.as_posix(),chunks={})
```
soxofaan, line 114: it looks like there is a lot of overlap/duplication with `_get_netcdf_zarr_metadata`
Can't this be handled more generically?

openeo/local/collections.py
```diff
@@ -0,0 +1,250 @@
...
+            except Exception as e:
+                _log.error(e)
+                continue
+    
+    local_collections_dict = {'collections':local_collections_list}
+    
+    return local_collections_dict
+
+def _get_geotiff_collections(local_collections_path):
+    if isinstance(local_collections_path,str):
```
soxofaan, line 235: again, this looks very similar to `_get_netcdf_zarr_collections`
I guess this can be generalized

tests/local/conftest.py
clausmichele: Done, fixture removed and created a helper function

clausmichele: I would need `rioxarray` if I want to test also geoTIFFs. Does it make sense to include it in `tests_require` or not?  I wonder if these tests should run only if installing the `localprocessing` part?

soxofaan: In principle, in CI we want to run all tests, but with localprocessing being quite heavy in terms of dependencies (all have to be installed with each run), this will make our CI a lot slower, so I'm not sure we should do that.
As long as it is an experimental feature we should skip the localprocessing test during the standard CI run and only run them manually as health check while development.

So in short: for now: keep all localprocessing related dependencies (also the ones just for testing) out of the standard requirements, and just put them in the localprocessing extra

openeo/local/collections.py
clausmichele: Merged the two functions into one

soxofaan:
I just had a look again at the dependency problem now that `openeo_processes_dask 2023.3.0` is available and the situation is now much better :+1:  
the `localprocessing` extra now only adds these 18 additional deps:
```
> affine==2.4.0
> attrs==22.2.0
> click==8.1.3
> click-plugins==1.1.1
> cligj==0.7.2
> geojson-pydantic==0.5.0
> networkx==2.8.8
> openeo-pg-parser-networkx==2023.3.0
> openeo-processes-dask==2023.3.0
> pendulum==2.1.2
> pydantic==1.10.6
> pyparsing==3.0.9
> pyproj==3.4.1
> pytzdata==2020.1
> rasterio==1.3.6
> rioxarray==0.13.4
> snuggs==1.4.7
> typing_extensions==4.5.0
```

soxofaan:
I finetuned the PR's branch (e.g. so that no regular non-localprocessing code paths where touched),
and merged in 05b9f0667472326ce765a3226b556ed24a857ab8

Great work, thanks!

PR #339, opened 2022-10-29
Add Datacube preview using a XYZ service

Ardweaden:
@dthiex 

@soxofaan


openeo/rest/datacube.py
```diff
@@ -14,13 +14,14 @@
...
 from builtins import staticmethod
-from typing import List, Dict, Union, Tuple, Optional, Any
+from typing import List, Dict, Union, Tuple, Optional, Any, Iterable
 
 import numpy as np
 import requests
 import shapely.geometry
 import shapely.geometry.base
 from shapely.geometry import Polygon, MultiPolygon, mapping
+from ipyleaflet import Map, TileLayer, basemaps
```
soxofaan, line 24: see my comment about setup.py below

openeo/rest/datacube.py
```diff
@@ -1829,6 +1830,71 @@ def validate(self) -> List[dict]:
...
+                        [spatial_extent["south"], spatial_extent["west"]],
+                        [spatial_extent["north"], spatial_extent["east"]],
+                    ]
+                )
+
+        try:
+            # Check we are inside Jupyer Notebook
+            shell = get_ipython().__class__.__name__
+            if shell == "ZMQInteractiveShell":
+                from IPython.display import display
```
soxofaan, line 1888: This `IPython`  depdendency should also be declared I guess

soxofaan (commented):
quick first review (didn't try it out yet)

openeo/rest/datacube.py
```diff
@@ -1829,6 +1830,71 @@ def validate(self) -> List[dict]:
...
+        center: Union[Iterable, None] = None,
+        zoom: Union[int, None] = None,
+    ) -> Tuple[Map, Service]:
+        """
+        Creates a service with the process graph and displays a map widget. Only supports XYZ.
+
+        :param center: (optional) Map center. Default is (0,0).
+        :param zoom: (optional) Zoom level of the map. Default is 1.
+
+        :return: ipyleaflet Map object and the displayed Service
```
soxofaan, line 1857: I'm wondering if we can't just a single object here

e.g. a service object that also contains the ipyleaflet map (which) is rendered through the `_repr_html_` API

openeo/rest/datacube.py
```diff
@@ -1829,6 +1830,71 @@ def validate(self) -> List[dict]:
...
+            if spatial_extent is not None:
+                m.fit_bounds(
+                    [
+                        [spatial_extent["south"], spatial_extent["west"]],
+                        [spatial_extent["north"], spatial_extent["east"]],
+                    ]
+                )
+
+        try:
+            # Check we are inside Jupyer Notebook
```
soxofaan, line 1885: Maybe I'm missing some context, but is there another context (other than Jupyter notebooks), where this `preview()` method makes sense?  If you can not `display()`  the `Map`, is it useful to construct it in the first place?

setup.py
```diff
@@ -46,7 +46,8 @@
           'numpy>=1.17.0',
           'xarray>=0.12.3',
           'pandas>0.20.0',
-          'deprecated>=1.2.12'
+          'deprecated>=1.2.12',
+          'ipyleaflet'
```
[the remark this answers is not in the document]
Ardweaden: Added on-demand viewer as `ondemandpreview` extra, that installs `ipyleaflet` and `ipython` explicitly. 

Everything is now imported inside the `preview` method and exceptions are raised if the runtime is not inside a Jupyter notebook and not all required dependencies are installed.

openeo/rest/datacube.py
m-mohr: Could conflict with the Vue components, but could be combined somehow, I guess.

jdries:
@soxofaan  I guess this one cannot be merged yet right? Also something we'd like to use in trainings.

openeo/rest/datacube.py
```diff
@@ -1829,6 +1829,75 @@ def validate(self) -> List[dict]:
...
+        """
+
+        # Check we are inside Jupyer Notebook
+        try:
+            shell = get_ipython().__class__.__name__
+            if shell == "ZMQInteractiveShell":
+                try:
+                    from IPython.display import display
+                    from ipyleaflet import Map, TileLayer, basemaps
+                except ModuleNotFoundError:
```
soxofaan, line 1865: This can be just an `ImportError` I think

openeo/rest/datacube.py
```diff
@@ -1829,6 +1829,75 @@ def validate(self) -> List[dict]:
...
+                    raise Exception(
+                        "Additional modules must be installed for on demand viewer. Run `pip install openeo[ondemandpreview]` or refer to the documentation."
+                    )
+            else:
+                raise Exception("On demand preview only supported in Jupyter notebooks!")
```
soxofaan, line 1870:
```suggestion
                raise Exception("On-demand preview only supported in Jupyter notebooks!")
```

openeo/rest/datacube.py
```diff
@@ -1829,6 +1829,75 @@ def validate(self) -> List[dict]:
...
+        except NameError:
+            raise Exception("On demand preview only supported in Jupyter notebooks!")
```
soxofaan, line 1872:
```suggestion
            raise Exception("On-demand preview only supported in Jupyter notebooks!")
```

soxofaan (commented):
some more quick notes

soxofaan:
> @soxofaan I guess this one cannot be merged yet right? Also something we'd like to use in trainings.

@jdries  If it's urgent, I think we could merge it as an experimental POC and do some polishing afterwards.

README.md
```diff
@@ -31,6 +31,10 @@ Allows you to interact with openEO backends from your own (local) Python environ
     
         pip install openeo
 
+* Installing with on-demand preview extra:
```
soxofaan, line 34: I think these "extra" install variants should be documented at https://open-eo.github.io/openeo-python-client/installation.html  (where there is more room for in-depth discussion) instead of the README.md  (which should stay as minimal as possible)
(the README.md should properly link to that of course)

openeo/rest/datacube.py
```diff
@@ -1829,6 +1829,75 @@ def validate(self) -> List[dict]:
...
         return self._connection.create_service(self.flat_graph(), type=type, **kwargs)
 
+    def _get_spatial_extent_from_load_collection(self):
+        pg = self.flat_graph()
+        for node in pg:
+            if pg[node]["process_id"] == "load_collection":
+                if "spatial_extent" in pg[node]["arguments"] and all(
+                    cd
+                    for cd in pg[node]["arguments"]["spatial_extent"]
+                    for cd in ["east", "west", "south", "north"]
```
soxofaan, line 1839: There is something off here: `cd` is used in two for loops at the same time

I think the `all()` will always resolve to True

soxofaan: maybe this was meant?
```
all(
    cd in pg[node]["arguments"]["spatial_extent"] 
    for cd in ["east", "west", "south", "north"]
)
```

soxofaan (commented):
some more notes

examples/notebooks/Untitled.ipynb
```diff
@@ -0,0 +1,143 @@
+{
```
soxofaan, line 1: Was it intended to add this notebook called "Untitled.ipynb"?

examples/notebooks/Datacube preview.ipynb
```diff
@@ -89,10 +100,22 @@
...
+   "outputs": [
+    {
+     "ename": "NameError",
+     "evalue": "name 'service' is not defined",
+     "output_type": "error",
+     "traceback": [
+      "\u001b[1;31m---------------------------------------------------------------------------\u001b[0m",
+      "\u001b[1;31mNameError\u001b[0m                                 Traceback (most recent call last)",
+      "Cell \u001b[1;32mIn [6], line 1\u001b[0m\n\u001b[1;32m----> 1\u001b[0m \u001b[43mservice\u001b[49m\u001b[38;5;241m.\u001b[39mdelete_service()\n",
+      "\u001b[1;31mNameError\u001b[0m: name 'service' is not defined"
```
soxofaan, line 115: Why is this error here? Is this a bug in the notebook?

openeo/rest/datacube.py
Ardweaden: I added just a simple wrapper if that's ok.

examples/notebooks/Untitled.ipynb
Ardweaden: Removed

README.md
Ardweaden: I only added the basic explanation.

setup.py
```diff
@@ -75,6 +80,7 @@
             "oschmod>=0.3.12"
         ],
         "localprocessing": localprocessing_require,
+        "ondemandpreview": ondemandpreview_require,
```
soxofaan, line 83: I'm wondering if we should put this feature under a very specific extra "ondemandpreview". 
Maybe it's more user friendly to just define a generic extra "jupyter" that allows a user to enable various jupyter notebook related extra/optional features (such as the "ondemand preview" feature). Another feature under this umbrella could be a bounding box selection widget based on ipywidget.

what do you think @m-mohr @clausmichele ?

openeo/rest/datacube.py
```diff
@@ -1900,6 +1900,86 @@ def validate(self) -> List[dict]:
...
+            raise OpenEoClientException("Backend does not support service type 'XYZ'.")
+
+        # Check we are inside Jupyer Notebook
+        try:
+            shell = get_ipython().__class__.__name__
+            if shell == "ZMQInteractiveShell":
+                try:
+                    from IPython.display import display
+                    import ipyleaflet
+                except ImportError:
```
soxofaan, line 1937: I think it's cleaner to first do the `import IPython` attempt, before doing that magic `get_ipython`  call. The latter will only work anyway if the former worked.

I'd also  avoid calling this magic `get_ipython` function that is nowhere imported, which trips up linting tools 
If I understand correctly it can be imported explicitly with 
```python
import IPython.core.getipython
IPython.core.getipython.get_ipython()
```

setup.py
m-mohr: I don't have a strong opinion as I'm not so much into all the details. Is jupyter the right name though? I also felt like "ondemandpreview" is quite lengthy.

I assume the Jupyter rendering via Vue components would stay in core, right?

Ardweaden: Was done.

openeo/rest/datacube.py
Ardweaden: Should be fixed.

openeo/rest/datacube.py
```diff
@@ -1900,6 +1900,87 @@ def validate(self) -> List[dict]:
...
+            # Check we are inside Jupyer Notebook
+            shell = get_ipython().__class__.__name__
+            if shell != "ZMQInteractiveShell":
+                raise Exception("On-demand preview only supported in Jupyter notebooks!")
+        except ImportError:
+            raise Exception(
+                "Additional modules must be installed for on demand viewer. Run `pip install openeo[jupyter]` or refer to the documentation."
+            )
+        except NameError:
+            raise Exception("On-demand preview only supported in Jupyter notebooks!")
```
soxofaan, line 1944: under #237/#420 I added `in_jupyter_context` helper:
https://github.com/Open-EO/openeo-python-client/blob/48edef8256450772a9dd5d30d59ab343b66edfec/openeo/internal/jupyter.py#L85-L92

this should make it possible to simplify this whole try-except construct

openeo/rest/datacube.py
```diff
@@ -1900,6 +1900,87 @@ def validate(self) -> List[dict]:
...
+        class Preview:
+            """
+            On-demand preview instance holding the associated XYZ service and ipyleaflet Map
+            """
+            def __init__(self, service: Service, ipyleaflet_map: ipyleaflet.Map):
+                self.service = service
+                self.map = ipyleaflet_map
+
+            def _repr_html_(self):
+                display(self.map)
```
soxofaan, line 1977: I would move the `from IPython.display import display` from above just inside this `_repr_html_`. 
I even wonder if it is possible to eliminate the need for an explicit import of that display function.
For example, does it work to just do `return self.map._repr_html_()`?

openeo/rest/datacube.py
```diff
@@ -1900,6 +1900,87 @@ def validate(self) -> List[dict]:
...
+            center=center or (0, 0),
+            zoom=zoom or 1,
+            scroll_wheel_zoom=True,
+            basemap=ipyleaflet.basemaps.OpenStreetMap.Mapnik,
+        )
+        service_layer = ipyleaflet.TileLayer(url=service_metadata["url"])
+        m.add_layer(service_layer)
+
+        if center is None and zoom is None:
+            spatial_extent = self._get_spatial_extent_from_load_collection()
```
soxofaan, line 1959: Isn't it risky to load the whole spatial extent from load_collection, which could be huge?
Shouldn't there be an upper limit for the spatial extent that is shown initially/by default?

Ardweaden: Well this would have to be something that the XYZ service specifies in metadata imo. 
In our case above a certain zoom level (depending on collection) empty images are returned.

dthiex:
@soxofaan We would really like to get this merged so we can promote a notebook using this functionality directly from the python client. I believe my college @Ardweaden fixed all of the mentioned issues and it's just awaiting a final review.

Do you think this could happen within the next days and how long do we (you) usually wait to release a new version of the client?

openeo/rest/datacube.py
```diff
@@ -1957,6 +1958,81 @@ def validate(self) -> List[dict]:
...
+        return None
+
+    def preview(
+        self,
+        center: Union[Iterable, None] = None,
+        zoom: Union[int, None] = None,
+    ):
+        """
+        Creates a service with the process graph and displays a map widget. Only supports XYZ.
+
```
soxofaan, line 1979: Add this new paragraph here:

```
.. warning:: experimental process: not generally supported, API subject to change.
.. versionadded:: 0.19.0
```

soxofaan (commented):
yeah sorry for the delay

I think this is practically ready to merge

could you also add a CHANGELOG entry

soxofaan:
> how long do we (you) usually wait to release a new version of the client

there is no fixed schedule for releases, I can do it this or next week

openeo/rest/datacube.py
Ardweaden: Added

dthiex:
> there is no fixed schedule for releases, I can do it this or next week

@soxofaan Great and also thanks for the fast reply! We added the 2 small additions requested and this is ready to be merged now. Could you please do this and include it in the next deploy (we both don't have the rights to do so).

soxofaan:
I just tried to play with the notebook, but it seems that the autofit/autozoom feature is not working: 
I just get a "whole world" view and I have no idea where to zoom in to to see the actual preview:
image

soxofaan:
I also noticed that my IDE (pycharm)  found usage of a deprecated API here:
```python
        m.add_layer(service_layer)
```


According to https://github.com/jupyter-widgets/ipyleaflet/releases/tag/0.17.0 that should be `m.add(...)`.

Ardweaden:
> I just tried to play with the notebook, but it seems that the autofit/autozoom feature is not working: I just get a "whole world" view and I have no idea where to zoom in to to see the actual preview: image
> 
> (working with ipyleaflet-0.17.3, which is latest version atm)

I cannot reproduce it. I tried with version 0.16.0 and it worked for me as well. Anyway, the version is now set to >= 0.17.0

soxofaan:
ok thanks, merged it now

soxofaan:
FYI: version 0.19.0 has been released now as well

PR #342, opened 2022-11-22
Update udp.rst

Pratichhya:
typo errors updated in udp file.


Pratichhya:
updated importing of BearerAuth

Pratichhya:
In the updated openEO version, providing crs showed an error for spatial_extent when it was set as string. Change in datatype was needed for executing the entire process.

examples/notebooks/Compositing.ipynb
```diff
@@ -441,7 +441,7 @@
    "source": [
     "timeseries = s2_radiometry\\\n",
     ".filter_temporal(\"2017-10-01\",\"2017-10-31\")\\\n",
-    ".filter_bbox(west=761104,east=763281,south=6543830,north=6544655,crs=\"EPSG:3857\")\n",
+    ".filter_bbox(west=761104,east=763281,south=6543830,north=6544655,crs=\3857\)\n",
```
soxofaan, line 444: This looks wrong, I don't think the backslashes should be preserved here `\`

examples/notebooks/Exploring_Backend_GEE.ipynb
```diff
@@ -34,7 +34,7 @@
     "IMAGE_EAST = 16.65\n",
     "IMAGE_NORTH = 48.31\n",
     "IMAGE_SOUTH = 48.1\n",
-    "IMAGE_SRS = \"EPSG:4326\"\n",
+    "IMAGE_SRS = \4326\n",
```
soxofaan, line 37: same as above: there seem to be unnecessary  `\` here now

examples/notebooks/PoC_EODC.ipynb
```diff
@@ -15,7 +15,7 @@
    "source": [
     "import logging\n",
     "import openeo\n",
-    "from openeo.auth.auth_bearer import BearerAuth\n",
+    "from openeo.rest.auth.auth import BearerAuth\n",
```
soxofaan, line 18: good catch, but this whole notebook looks completely outdated
e.g. a user should not have to import `BearerAuth` in the first place

examples/notebooks/Sentinel2_NDVI_GEE.ipynb
```diff
@@ -43,7 +43,7 @@
     "IMAGE_EAST = 16.524124\n",
     "IMAGE_NORTH = 48.320647\n",
     "IMAGE_SOUTH = 48.138600\n",
-    "IMAGE_SRS = \"EPSG:4326\"\n",
+    "IMAGE_SRS = \4326\n",
```
soxofaan, line 46: same `\` issue here

examples/notebooks/SimpleCompositor.ipynb
```diff
@@ -1371,7 +1371,7 @@
    "source": [
     "timeseries = s2_datacube\\\n",
     ".filter_temporal(\"2020-07-01\",\"2020-07-31\")\\\n",
-    ".filter_bbox(west=2.88254,east=3.24234,south=51.13243,north=51.45653,crs=\"EPSG:4326\")\n",
+    ".filter_bbox(west=2.88254,east=3.24234,south=51.13243,north=51.45653,crs=\4326\)\n",
```
soxofaan, line 1374: same `\` issue here

examples/notebooks/openEO_Python_Sentinel2_EURAC.ipynb
```diff
@@ -74,7 +74,7 @@
     "datacube = con.load_collection(collection_id=\"openEO_S2_32632_10m_L2A_D22\", bands=['B02', 'B03', 'B04', 'B08'])\n",
     "# perform spatial subsetting (e.g around the city of Bolzano)\n",
     "datacube = datacube.filter_bbox(west=11.279182434082033, south=46.464349400461145, \n",
-    "                                east=11.406898498535158, north=46.522729291844286, crs=\"EPSG:32632\")\n",
+    "                                east=11.406898498535158, north=46.522729291844286, crs=\32632\)\n",
```
soxofaan, line 77: same `\` issue here

examples/notebooks/openeo-terrascope-webinar.ipynb
```diff
@@ -247,7 +247,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "bbox = {\"west\": 5.05, \"south\": 51.20, \"east\": 5.15, \"north\": 51.24, \"crs\": \"EPSG:4326\"}\n",
+    "bbox = {\"west\": 5.05, \"south\": 51.20, \"east\": 5.15, \"north\": 51.24, \"crs\": \4326\}\n",
```
soxofaan, line 250: same `\` issue here

docs/udp.rst
```diff
@@ -311,7 +311,7 @@ from the dictionary above::
     connection.save_user_defined_process(
         user_defined_process_id="fahrenheit_to_celsius",
         process_graph=fahrenheit_to_celsius,
-        parameters=[Parameter.number(name="f", description="Degrees Fahrenheit")
+        parameters=[Parameter.number(name="f", description="Degrees Fahrenheit")]
```
soxofaan, line 314: good catch

I already cherry-picked these fixes in c4a117040a5208b68b8a09dc46af6195935d6508

soxofaan (commented):
great, thanks for the PR.

I think however something went wrong with fixing the  crs arguments from string to ints

docs/udf.rst
```diff
@@ -279,8 +279,46 @@ instead of the original digital number range (thousands):
 
 Illustration of data chunking in ``apply`` with a  UDF
 ========================================================
+Chunk_polygon can be used to apply a specific kind of process to spatial chunks of a data cube. In the following example, we repeat the method of rescaling a data cube using chunk_polygon. Furthermore, in this example, we show that with the Geojson file, users can try with shapefile if they want to and fed in as a feature collection. Though a function to read JSON is shown here, you can replace it with your process.
```
soxofaan, line 282: avoid putting a whole paragraph on a single line, this makes it hard to review/maintain

e.g see https://github.com/sembr/specification

docs/udf.rst
```diff
@@ -279,8 +279,46 @@ instead of the original digital number range (thousands):
...
 
-TODO
+.. code-block:: python
+    :linenos:
+    :caption: UDF alongwith chunk_polygon
```
soxofaan, line 286: I guess this "alongwith" is a typo

docs/udf.rst
```diff
@@ -279,8 +279,46 @@ instead of the original digital number range (thousands):
...
+    :emphasize-lines: 14-25
+
+    import openeo
+
+    # Create connection to openEO back-end
+    connection = openeo.connect("...").authenticate_oidc()
+
+    # Load your data cube based on your prefernce
+
+    S2_cube = eoconn.load_collection(
```
soxofaan, line 296: `eoconn` -> `connection` I guess

docs/udf.rst
```diff
@@ -279,8 +279,46 @@ instead of the original digital number range (thousands):
...
+    # Create a UDF object from inline source code.
+    my_code = """
+    from openeo.udf import XarrayDataCube
+
+    def apply_datacube(cube: XarrayDataCube, context: dict) -> XarrayDataCube:
+        array = cube.get_array()
+        array.values = 0.0001* array.values
+        return cube
+    """
+    my_udf = lambda data: data.run_udf(udf=my_code,runtime='python')
```
soxofaan, line 311: I would avoid using this complicated `lambda` construct, and prefer the new `openeo.UDF`  approach as illustrated in the "A first example: apply with an UDF to rescale pixel values" section

docs/udf.rst
```diff
@@ -279,8 +279,46 @@ instead of the original digital number range (thousands):
...
+
+    # Pass UDF object as child process to `chunk_polygon`.
```
soxofaan, line 313: this comment here does not make sense at the moment

docs/udf.rst
```diff
@@ -279,8 +279,46 @@ instead of the original digital number range (thousands):
...
+
+    aoi = read_json("./aoi/caro_multipoly.geojson")
```
soxofaan, line 315: bit of confusing here: what is this "caro"?  Is this a geometry a multipolygon or a feature collection. It seems the former, but the latter would actually be a better illustration of this feature

docs/udf.rst
```diff
@@ -279,8 +279,46 @@ instead of the original digital number range (thousands):
...
+    rescaled_chunks = S2_cube.chunk_polygon(chunks=aoi,process=my_udf)
+
+    # perform time dimension reduction
+    Rrescaled_chunks = rescaled_chunks.reduce_dimension(dimension="t", reducer="mean")
```
soxofaan, line 319: is this `Rrescaled_chunks`  a typo or intended?

docs/udf.rst
```diff
@@ -279,8 +279,46 @@ instead of the original digital number range (thousands):
...
+
+Once the process is completed you can download the result either by direct download (in case of the small spatial extent with few processing) or perform create a `batch job <https://open-eo.github.io/openeo-python-client/batch_jobs.html>`_ in case it is a heavy task over a large extent.
```
soxofaan, line 321: This paragraph seems a bit redundant here (this is discussed elsewhere in more detail already). I would focus on the UDF-specific details here. Also note that you should avoid absolute URLs for internal references (instead use a reference like ":ref:`batch-jobs-chapter`" here)

soxofaan:
Regarding the latest UDF doc updates: I think it's better to start a new feature branch and pull request for this, instead of appending to an existing PR, because this will get unmanageable in the long run

docs/udf.rst
Pratichhya: Updated

docs/udf.rst
Pratichhya: typo error updated

docs/udf.rst
Pratichhya: connection corrected

docs/udf.rst
Pratichhya: Actually, initially I used the openeo.UDF method, but received this error:
`AttributeError: 'NoneType' object has no attribute 'list_udf_runtimes'`

Thus I switched to this method, to avoid the error.

docs/udf.rst
Pratichhya: updated

docs/udf.rst
Pratichhya: It was intended, but I understand it to be confusing. Thus I renamed it.

docs/udf.rst
Pratichhya: updated

docs/udf.rst
soxofaan: I think I found the issue and fixed it in 8739413938e5fed44f76547e297511036b228b90, can you try again with that fix?

docs/udf.rst
```diff
@@ -279,8 +279,50 @@ instead of the original digital number range (thousands):
...
+
+    def apply_datacube(cube: XarrayDataCube, context: dict) -> XarrayDataCube:
+        array = cube.get_array()
+        array.values = 0.0001* array.values
+        return cube
+    """
+    my_udf = lambda data: data.run_udf(udf=my_code,runtime='python')
+    
+    #Though a function to read JSON is shown here, you can replace it with your process.
+    aoi = read_json("./aoi/your_aoi.geojson")
```
soxofaan, line 316: I think you should clarify what this geojson contains: just a simple polygon or a feature collection?

docs/udf.rst
```diff
@@ -279,8 +279,50 @@ instead of the original digital number range (thousands):
...
+
+    #use `chunk_polygon` to apply my udf on selected aoi.
+    rescaled_chunks = S2_cube.chunk_polygon(chunks=aoi,process=my_udf)
+
+    # perform time dimension reduction
+    Final_result = rescaled_chunks.reduce_dimension(dimension="t", reducer="mean")
```
soxofaan, line 322: variable names should not contain upper case, I would just call this `result`

docs/udf.rst
soxofaan: I see from  that stack trace that you are still using an old version of the library. Are you sure you switched to latest master from git and refreshed your notebook (assuming you are working in a notebook)?

Pratichhya: Yes, it worked, and I have updated my UDF.

Pratichhya:
I am sorry, I didn't realise this PR is still active. Should I still follow with this or create a new one if needed?

soxofaan:
I would indeed start a new feature branch and PR, this one collected too much cruft I'm afraid

Pratichhya:
Indeed I agree.

PR #345, opened 2022-11-28
request for information

Pratichhya:
not clear regarding the visualization

### DIFF
--- a/docs/basics.rst
+++ b/docs/basics.rst
@@ -236,6 +236,8 @@ In the next steps we will address cloud masking.
 
 .. image:: _static/images/basics/evi-composite.png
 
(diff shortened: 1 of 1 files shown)


soxofaan:
can you share the plots you reproduced?

Pratichhya:
This is my **evi-composite.tiff**  when I plot based on whatever output is received.
image
and this is masked_evi-composite.tiff
image

soxofaan:
for that first (unmasked) plot, I indeed probably added something like `vmin=0, vmax=1` to the plot to limit the visualization to the valid/interesting part of the evi range.
Can you try to add that

Pratichhya:
Yes, it is due to range selected for visualization but I don't think it is vmax=1 and vmin =0. It gives me different output than that in the guide even when I select in the range of -1 to 1. But may be if would be helpful is a note is provided. 
image
 
# image in the guide
image

soxofaan:
apparently the original evi-composite image is indeed out of sync with what you currently get from executing the related code

I've updated the inage in 03fdfcb

soxofaan:
@Pratichhya   FYI: for these kind of questions/requests, you should not create a pull request, but just create a basic issue from https://github.com/Open-EO/openeo-python-client/issues

Issue #198, opened 2021-03-29
Permissions issues with PrivateJsonFile on windows


soxofaan:
Windows does not support fine grained linux style permission `chmod`ing, causing perrmission issues like 

```
File "D:\tmp\env\lib\site-packages\openeo\rest\auth\config.py", line 85, in _write
    assert_private_file(self._path)
  File "D:\tmp\env\lib\site-packages\openeo\rest\auth\config.py", line 30, in assert_private_file
    p=path, a=mode, e=_PRIVATE_PERMS)
PermissionError: File C:\Users\john\AppData\Local\Temp\win-perms-test.json is readable by others: st_mode 100666 (expected permissions: 600).```

soxofaan:
did a first quick fix to make OIDC at least usable on Windows:
now a warning is raised when permissions are not correctly set

soxofaan:
next step: investigate oschmod (https://pypi.org/project/oschmod/)  to set permissions on windows fully (which does not work with vanilla `os.chmod`)

soxofaan:
Also interesting solution: some kind of logout feature to remove these local files: #250

soxofaan:
FYI failing tests related to this are blocker for 
- #176

soxofaan:
PR #351 was merged

PR #351, opened 2022-12-15
Issue198 permission issues with privatejsonfile on windows

JohanKJSchreurs:
Implements https://github.com/Open-EO/openeo-python-client/issues/198

This uses the oschmod library to be able to use similar file permission on Windows.
For now it only installs and uses it on WIndows systems.

Possibly this can be simplified a bit, making it more uniform across Windows, Linux and Mac by using oschmod for all platforms.
But I propose to test this first and see if there are any issues.


openeo/rest/auth/config.py
```diff
@@ -18,22 +18,29 @@
 from openeo.config import get_user_config_dir, get_user_data_dir
 from openeo.util import rfc3339, deep_get, deep_set
 
+if platform.system() == 'Windows':
```
soxofaan, line 21: Maybe instead of an "if windows: import"  just do "try: import ...; except: oschmod=None"
then the openeo client package is even usable in a windows context where  oschmod can not be installed for some reason

openeo/rest/auth/config.py
```diff
@@ -18,22 +18,29 @@
...
 log = logging.getLogger(__name__)
 
 
 def assert_private_file(path: Path):
     """Check that given file is only readable by user."""
-    mode = path.stat().st_mode
+
+    # use oschmod on Windows
+    # TODO: can we use oschmod for all operating systems, make the code simpler and consitent?
+    if platform.system() == "Windows":
```
soxofaan, line 35: building on comment above" just do "if oschmod"  here instead of "if windows"

setup.py
```diff
@@ -47,6 +47,8 @@
           'xarray>=0.12.3',
           'pandas>0.20.0',
           'deprecated>=1.2.12',
+          'oschmod; sys_platform == "win32"',
+          'pywin32; sys_platform == "win32"',
```
soxofaan, line 51: I don't think it's necessary to list pywin32 explicitly as dependency, we just use oschmod and pywin32 is one of it's implementation details, right?

tests/rest/auth/test_config.py
```diff
@@ -1,11 +1,19 @@
...
 import pytest
 
 import openeo.rest.auth.config
 from openeo.rest.auth.config import RefreshTokenStore, AuthConfig, PrivateJsonFile
 
+# TODO: We could simplify the code if we can verify oschmod works transparently on Linux/Mac as well.
+#   I think oschmod does work on Linux and Mac, but since it we use it for a
+#   security sensitive feature, let's make sure there are no problems first.
+#   For now, it is fine to have some Windows specific code.
+if platform.system() == 'Windows':
```
soxofaan, line 14: same

openeo/rest/auth/config.py
soxofaan: (e.g. it would allow a test run on Linux to test the "windows code path")

openeo/rest/auth/config.py
soxofaan: and maybe encapsulate the if in a new `get_file_mode` function, which you can also reuse in the tests

tests/test_config.py
```diff
@@ -159,6 +160,13 @@ def test_get_config_verbose(tmp_path, caplog, capsys, verbose, force_interactive
         config = get_config()
         assert config.get("general.verbose") == verbose
 
-    regex = re.compile(f"Loaded.*config from.*{config_path}")
+    # On Windows the path in config_path would make the regex invalid because
```
soxofaan, line 163: the `re` module has built-in escape function, just
```
regex = re.compile(f"Loaded.*config from.*{re.escape(config_path)}")
```
should do the trick cross-platform

setup.py
JohanKJSchreurs: Yes, we should be able to leave out pywin32.
When pip resolves dependencies it should know whether the platform needs it or not.

Will change it and test it.

openeo/rest/auth/config.py
JohanKJSchreurs: Good idea. I'm on it.

JohanKJSchreurs: Done.

openeo/rest/auth/config.py
JohanKJSchreurs: Done.

openeo/rest/auth/config.py
```diff
@@ -86,7 +93,12 @@ def _write(self, data: dict):
         # TODO: add file locking to avoid race conditions?
         with self._path.open("w", encoding="utf8") as f:
             json.dump(data, f, indent=2)
-        self._path.chmod(mode=_PRIVATE_PERMS)
+        # On Windows we use oschmod because Python chmod implementation doesn't work on Windows.
+        # TODO: can we use oschmod on all platforms? Seems like that is the intention of oschmod.
+        if platform.system() == "Windows":
```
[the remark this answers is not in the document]
JohanKJSchreurs: Done.

tests/rest/auth/test_config.py
JohanKJSchreurs: Done.

tests/test_config.py
JohanKJSchreurs: Done.

setup.py
JohanKJSchreurs: Left out pywin32

oschmod is a required dependency for Windows, but not for the other platforms.

I added oschmod as a new item in extra_require, which also called "oschmod"
This way it can be optionally installed on other platforms more easily, with a single pip-install command:
```
pip-install .[dev,oschmod]
```

soxofaan:
looks good

can you also add a CHANGELOG.md entry?

JohanKJSchreurs:
> can you also add a CHANGELOG.md entry?
CHANGELOG.md updated.

soxofaan:
merged in 1de20bafa29c4b2cedde64c4a1399503fd389c7c

Issue #350, opened 2022-12-15
Windows compatibility: unit test about configuration fail


JohanKJSchreurs:
Some unit tests fail on Windows. They seem to be related to configuration files and where they are located.

At first sight, the tests seem to make some assumptions, such as some default paths and that the path separator is / .
So it might just be that for these 3 tests, it it is only the test that needs to be fixed.

This is the test summary:

```
============================================================= short test summary info =============================================================
FAILED tests/test_config.py::test_get_user_config_dir - AssertionError: assert WindowsPath('C:/Users/testuser/AppData/Roaming/openeo-python-client') == (WindowsPath('C:/dev-projects/VITO/codebases/ope...
FAILED tests/test_config.py::test_get_user_data_dir - AssertionError: assert WindowsPath('C:/Users/testuser/AppData/Roaming/openeo-python-client') == (WindowsPath('C:/dev-projects/VITO/codebases/ope...
FAILED tests/rest/auth/test_cli.py::test_paths - AssertionError: assert '/auth-config.json' in 'openEO auth config: C:\\Users\\testuser\\AppData\\Local\\Temp\\pytest-of-testuser\\pytest-52\\test...
============================================ 3 failed, 1260 passed, 17 skipped, 179 warnings in 55.49s ============================================
```

soxofaan:
FYI:  this bit of conftest is probably important here:
https://github.com/Open-EO/openeo-python-client/blob/01f525615fead0db67810fc2f1b92f4e15b7a089/tests/conftest.py#L11-L13

it sets up some "home dir" related env vars that probably only work on linux

soxofaan:
PR  #354 was merged

PR #354, opened 2022-12-15
Issue #350: Fix unit tests that failed on Windows because of differen…

JohanKJSchreurs:
# Issue 350: Fix unit tests that failed on Windows because of difference in config paths.

This fixes the three test mentioned in issue #350 

On windows we needed to set another environment variable APPDATA in two of the tests, and use the path separator from os.sep in another test.

Together with issue the PR for issue #198,  which is PR https://github.com/Open-EO/openeo-python-client/pull/351, it should make all tests pass on Windows.


tests/test_config.py
```diff
@@ -15,11 +16,23 @@
 
 
 def test_get_user_config_dir():
-    assert get_user_config_dir() == Path(__file__).parent / "data/user_dirs/config/openeo-python-client"
+    # For Linux/Mac we configure the config and data directories in conftest.py
+    # via the environment variables XDG_CONFIG_HOME and XDG_DATA_HOME.
+    # But on Windows we need to set APPDATA instead, and we don't have separate
+    # variables for config and data.
+    if platform.system() == "Windows":
+        os.environ["APPDATA"] = str(Path(__file__).parent / Path("data/user_dirs/config"))
```
soxofaan, line 24: ca'tn we just put that `os.environ`  in conftest.py like https://github.com/Open-EO/openeo-python-client/blob/01f525615fead0db67810fc2f1b92f4e15b7a089/tests/conftest.py#L11-L13  ? does that work? even without an "if windows" check?

tests/test_config.py
```diff
@@ -15,11 +16,23 @@
...
 
 def test_get_user_data_dir():
-    assert get_user_data_dir() == Path(__file__).parent / "data/user_dirs/data/openeo-python-client"
+    # For Linux/Mac we configure the config and data directories in conftest.py
+    # via the environment variables XDG_CONFIG_HOME and XDG_DATA_HOME.
+    # But on Windows we need to set APPDATA instead, and we don't have separate
+    # variables for config and data.
+    if platform.system() == "Windows":
+        os.environ["APPDATA"] = str(Path(__file__).parent / Path("data/user_dirs/data"))
+    assert get_user_data_dir() == Path(__file__).parent / Path("data/user_dirs/data/openeo-python-client")
```
soxofaan, line 35: just to make sure I understand how things work on windows:
that change of `"data/user_d...` to `Path("data/user_d..`   is that to make sure that dir separators are normalized to windows style?

Also: doesn't that just happen automatically because of the `/` operator in  `Path(...) / "...."`?

tests/test_config.py
JohanKJSchreurs: The problem here is that the decision branch for Windows does not use those variables at all. 
Not unless we want to change that.

But more so, Windows does not have separate environment variables for config and data. It has just one: APPDATA.
So for the test you kind of have to set APPDATA to different values in each of the two tests, or alternatively you have to change the test even more to compare with a different expected path.

https://github.com/Open-EO/openeo-python-client/blob/master/openeo/config.py#L22-L42

soxofaan: ok I understand now.

I would then still set (for example):
```python
os.environ["APPDATA"] = str(Path(__file__).parent / "\\AppData\\Roaming"
```
in conftest,py (so that this config treats different os'es more equally)

and change the expected values in the tests accordingly, e.g.

```python
if platform.system() == "Windows":
    expected =...
else:
    expected = ...
assert get_user_config_dir() == expected
```

soxofaan: Then the expected values in the test also more reflect how it works in reality (that there is no distinction between data and config on windows)

tests/test_config.py
JohanKJSchreurs: That was indeed the why I initially wrapping that string in Path(...) 
However, I checked again and I was wrong. This is not necessary after all.  

Early on it seemed like it wasn't converting the string that was concatenated with the / operator. But I was mistaken and the error was something else.  So we don't actually need to wrap the second part in Path(...).

I have now removed the extra Path(...).
Simpler is better, less surprise is more readable.

tests/test_config.py
JohanKJSchreurs: Indeed that is cleaner.
Done.

PR #355, opened 2022-12-15
Docs update and extend documentation

JohanKJSchreurs:
- Extended section about development Installation on Windows
- Made it easier to use the docs/Makefile on Windows: python command is slightly different.


docs/Makefile
```diff
@@ -1,9 +1,22 @@
...
+# pyenv. And that shill will not work, because our sphinx is not installed there.
+PYTHON_EXE = python
+else
+PYTHON_EXE = python3
+endif
+
 # You can set these variables from the command line.
 SPHINXOPTS    =
-SPHINXBUILD   = python3 -msphinx
+SPHINXBUILD   = $(PYTHON_EXE) -m sphinx
```
soxofaan, line 19: just change this to  `python` for all platforms 
no need for platform specific stuff then

docs/installation.rst
```diff
@@ -103,3 +104,108 @@ you make a change.
 The ``[dev]`` (a so-called "extra") installs additional development related dependencies,
 for example to run the unit tests.
 
+
+
+
+Development Installation on Windows
```
soxofaan, line 110: This is adds a lot of dev-specific documentation on general installation page, which might intimidate a reader that is not closely looking

It's maybe better to put this in docs/development.rst

docs/Makefile
JohanKJSchreurs: Done.

docs/installation.rst
JohanKJSchreurs: That makes sense indeed.  I moved it to docs/development.rst

I also updated that section content-wise.
It looks like with the current code a dev installs works find in a virtualenv on Windows.
But maybe it is prudent to keep the documented solutions around, until we are sure this never happens anymore.

Issue #335, opened 2022-10-13
OIDC device code flow: support `verification_uri_complete`


soxofaan:
See https://discuss.eodc.eu/t/its-no-longer-required-to-add-the-authentication-code-when-running-login/482

EGI OIDC device auth request apparently includes `verification_uri_complete` in addition to the standard `verification_uri`, e.g.:
```python
{'device_code': '5EKr8rAg...', 
 'user_code': 'ABCD-UIXY', 
 'verification_uri': 
     'https://aai.egi.eu/auth/realms/egi/device', 
 'verification_uri_complete':
     'https://aai.egi.eu/auth/realms/egi/device?user_code=ABCD-UIXY',
 'expires_in': 900, 'interval': 5}
```

Using the `verification_uri_complete` link instead of the `verification_uri` one makes the flow more user friendly (less tying or copy-pasting, less forms to click through), but it might feel less secure to the savvy user.

At least we could make it an opt-in config option to use `verification_uri_complete` instead of `verification_uri`

soxofaan:
code to trigger this flow:

```
import openeo

connection = openeo.connect("openeo.cloud")

connection.authenticate_oidc_device()
```

PR #360, opened 2023-01-10
Issue #335, Use verification_uri_complete for OIDC authentication if available.

JohanKJSchreurs:
Implements issue #335

### DIFF
--- a/openeo/rest/auth/oidc.py
+++ b/openeo/rest/auth/oidc.py
@@ -627,7 +627,10 @@ class OidcRefreshTokenAuthenticator(OidcAuthenticator):
         return data
 
 
-VerificationInfo = namedtuple("VerificationInfo", ["verification_uri", "device_code", "user_code", "interval"])
+VerificationInfo = namedtuple(
+    "VerificationInfo",
+    ["verification_uri", "verification_uri_complete", "device_code", "user_code", "interval"]
+)
 
 
 class OidcDeviceAuthenticator(OidcAuthenticator):
@@ -671,6 +674,8 @@ class OidcDeviceAuthenticator(OidcAuthenticator):
             verification_info = VerificationInfo(
                 # Google OAuth/OIDC implementation uses non standard "verification_url" instead of "verification_uri"
                 verification_uri=data["verification_uri"] if "verification_uri" in data else data["verification_url"],
+                # verification_uri_complete is optional, will be None if this key is not present
+                verification_uri_complete=data.get("verification_uri_complete"),
                 device_code=data["device_code"],
                 user_code=data["user_code"],
                 interval=data.get("interval", 5),
@@ -683,9 +688,14 @@ class OidcDeviceAuthenticator(OidcAuthenticator):
     def get_tokens(self, request_refresh_token: bool = False) -> AccessTokenResult:
         # Get verification url and user code
         verification_info = self._get_verification_info(request_refresh_token=request_refresh_token)
-        self._display("To authenticate: visit {u} and enter the user code {c!r}.".format(
-            u=verification_info.verification_uri, c=verification_info.user_code)
-        )
+        if verification_info.verification_uri_complete:
+            self._display(
+                f"To authenticate: visit {verification_info.verification_uri_complete}."
+            )
+        else:
+            self._display("To authenticate: visit {u} and enter the user code {c!r}.".format(
+                u=verification_info.verification_uri, c=verification_info.user_code)
+            )
 
         # Poll token endpoint
         elapsed = create_timer()
--- a/openeo/rest/auth/testing.py
+++ b/openeo/rest/auth/testing.py
@@ -41,6 +41,7 @@ class OidcMock:
         scopes_supported: List[str] = None,
         device_code_flow_support: bool = True,
         oidc_discovery_url: Optional[str] = None,
+        use_verification_uri_complete: Optional[bool] = False,
     ):
         self.requests_mock = requests_mock
         self.oidc_issuer = oidc_issuer
@@ -56,6 +57,7 @@ class OidcMock:
             if device_code_flow_support
             else None
         )
+        self.use_verification_uri_complete = use_verification_uri_complete
         self.state = state or {}
         self.scopes_supported = scopes_supported or ["openid", "email", "profile"]
 
@@ -83,6 +85,12 @@ class OidcMock:
             self.requests_mock.post(
                 self.device_code_endpoint, text=self.device_code_callback
             )
+            if self.use_verification_uri_complete:
+                self._set_user_code()
+                end_point = (
+                    f"{self.device_code_endpoint}?user_code={self.state['user_code']}"
+                )
+                self.requests_mock.post(end_point, text=self.device_code_callback)
 
     def webbrowser_open(self, url: str):
         """Doing fake browser and Oauth Provider handling here"""
@@ -151,6 +159,9 @@ class OidcMock:
         assert params["scope"] == self.expected_fields["scope"]
         return self._build_token_response()
 
+    def _set_user_code(self):
+        self.state["user_code"] = random_string(length=6).upper()
+
     def device_code_callback(
         self, request: requests_mock.request._RequestObjectProxy, context
     ):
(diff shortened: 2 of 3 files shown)


soxofaan (approved):
Looks good already

Also to do: Changelog entry, something like "Support for simplified OIDC device code flow (issue ...."

openeo/rest/auth/testing.py
```diff
@@ -83,6 +85,12 @@ def __init__(
             self.requests_mock.post(
                 self.device_code_endpoint, text=self.device_code_callback
             )
+            if self.use_verification_uri_complete:
+                self._set_user_code()
+                end_point = (
+                    f"{self.device_code_endpoint}?user_code={self.state['user_code']}"
+                )
+                self.requests_mock.post(end_point, text=self.device_code_callback)
```
soxofaan, line 93: This should be a GET mock instead of a POST mock (the user just clicks this link).
This is actually also wrong on line 85 above

openeo/rest/auth/testing.py
```diff
@@ -83,6 +85,12 @@ def __init__(
             self.requests_mock.post(
                 self.device_code_endpoint, text=self.device_code_callback
             )
+            if self.use_verification_uri_complete:
+                self._set_user_code()
```
soxofaan, line 89: I think conceptually we can not set the user code here yet, it should be done in the `device_code_endpoint` callback

openeo/rest/auth/testing.py
```diff
@@ -151,15 +159,24 @@ def token_callback_resource_owner_password_credentials(self, params: dict, conte
...
         assert params["scope"] == self.expected_fields["scope"]
         self.state["device_code"] = random_string()
-        self.state["user_code"] = random_string(length=6).upper()
         self.state["scope"] = params["scope"]
+
+        # When the mock should also provide the verification_uri_complete then we
+        # we have already set the user_code in the constructor, because we set up
+        # and extra request_mock there for a URL that includes the user_code.
+        if not self.use_verification_uri_complete:
+            self._set_user_code()
```
soxofaan, line 178: as noted above: the user code should be generated in all cases (with and without use_verification_uri_complete)

tests/rest/auth/test_oidc.py
```diff
@@ -437,46 +437,106 @@ def test_oidc_device_flow_without_pkce_nor_secret(requests_mock, caplog):
         flags=re.DOTALL
     )
 
-
-@pytest.mark.parametrize(["mode", "client_id", "use_pkce", "client_secret", "expected_fields"], [
-    (
+# fmt: off
+@pytest.mark.parametrize(
```
soxofaan, line 441: FYI: instead of one `@pytest.mark.parametrize` huge listing, it's also possible to combine multiple (which are combined in the same way as `itertools.product`). This will make this parameterization more compact I think

```
@pytest.mark.parametrize(["mode", "client_id", "use_pkce", "client_secret", "expected_fields"], [ ....
@pytest.mark.parametrize("use_verification_uri_complete", [False, True]
def test....
```

openeo/rest/auth/oidc.py
```diff
@@ -683,9 +688,14 @@ def _get_verification_info(self, request_refresh_token: bool = False) -> Verific
     def get_tokens(self, request_refresh_token: bool = False) -> AccessTokenResult:
         # Get verification url and user code
         verification_info = self._get_verification_info(request_refresh_token=request_refresh_token)
-        self._display("To authenticate: visit {u} and enter the user code {c!r}.".format(
-            u=verification_info.verification_uri, c=verification_info.user_code)
-        )
+        if verification_info.verification_uri_complete:
+            self._display(
+                f"To authenticate: visit {verification_info.verification_uri_complete}."
```
soxofaan, line 693: I tried this with a real flow and the point just after the URL is a bit tricky: maybe some tools/context might consider this point to be part of the URL, which will break the authentication. I think we should at least put a space between the URL and "."

openeo/rest/auth/testing.py
soxofaan: sorry I'm confused myself:
- `self.device_code_endpoint` is actually a POST,
- there is no need for a mock of `f"{self.device_code_endpoint}?user_code={self.state['user_code']}"`

openeo/rest/auth/testing.py
```diff
@@ -169,15 +186,21 @@ def device_code_callback(
                 assert "code_challenge" not in params
             else:
                 raise ValueError(expect_code_challenge)
-        return json.dumps(
-            {
+
+        token = {
```
soxofaan, line 190: FYI: this is not a (user) token, but this is verification data for the OIDC client (not the user)

tests/rest/auth/test_oidc.py
```diff
@@ -437,46 +437,106 @@ def test_oidc_device_flow_without_pkce_nor_secret(requests_mock, caplog):
...
+            "default-with-pkce", None, "$3cr3t",
+            {"scope": "df openid", "client_secret": "$3cr3t", "code_challenge": ABSENT, "code_verifier": ABSENT},
+            True,
+        ),
+    ],
+)
+# fmt: on
 def test_oidc_device_flow_auto_detect(
-        requests_mock, caplog, mode, client_id, use_pkce, client_secret, expected_fields
+    requests_mock,
```
soxofaan, line 532: There are also a couple of other tests above that could be covered

soxofaan:
FYI: after some more tinkering with this, I decided to redo the testing part in this feature branch: https://github.com/Open-EO/openeo-python-client/tree/issue335_oidc-support-verification_uri_complete2 
I would continue with that one

openeo/rest/auth/oidc.py
JohanKJSchreurs: Good catch!

openeo/rest/auth/testing.py
JohanKJSchreurs: N.B. I will continue on your branch, just documenting the resolution of the comment here.

Yes, I was not sure whether this extra mock was needed to test the URL that included the query paramater, but we don't need it.

Removing this mock also makes the rest of the code a lot simpler, because those ugly changes that generate the user code at construction time were only necessary to have the user code at the time we set up the mock.

openeo/rest/auth/testing.py
JohanKJSchreurs: I agree. Was not ideal and was only needed for the mock that is now removed.

openeo/rest/auth/testing.py
JohanKJSchreurs: I agree. Was not ideal and was only needed for the mock that is now removed.

tests/rest/auth/test_oidc.py
JohanKJSchreurs: That was exactly the kind of thing I was looking for. Thanks!

openeo/rest/auth/testing.py
JohanKJSchreurs: Thanks for clarifying that.

tests/rest/auth/test_oidc.py
JohanKJSchreurs: Indeed. 
I will continue with your branch that has these covered.

JohanKJSchreurs:
Still a small problem to solve:
 
Though the correct code is now present on https://github.com/Open-EO/openeo-python-client/tree/issue335_oidc-support-verification_uri_complete2 , the pull request does not look at that branch,. So it won't to show the correct changes, not yet.

I will have to merge the new branch into the original branch, or even copy the files because I don't want to reintroduce the old unnecessary modifications.

soxofaan:
> the pull request does not look at that branch

I would just create a new PR, overruling (and referencing) this one

JohanKJSchreurs:
> > the pull request does not look at that branch
> 
> I would just create a new PR, overruling (and referencing) this one

That is indeed simpler. The new PR is this one:  https://github.com/Open-EO/openeo-python-client/pull/362

JohanKJSchreurs:
Closing this PR, because we would need to merge the new branch, so we continue on this new PR:  https://github.com/Open-EO/openeo-python-client/pull/362

PR #362, opened 2023-01-11
Issue335 (continued) OIDC support verification_uri_complete

JohanKJSchreurs:
Implements issue https://github.com/Open-EO/openeo-python-client/issues/335 and replaces PR https://github.com/Open-EO/openeo-python-client/pull/360


soxofaan:
almost forgot: we also have a bit in the docs about this that needs bit of finetuning now:
https://github.com/Open-EO/openeo-python-client/blob/master/docs/auth.rst#device-flow

JohanKJSchreurs:
> OK, I've updated that section of the docs.

To make the review a little bit easier and see the *rendered* result, here is the link to the relevant docs on this branch:

https://github.com/Open-EO/openeo-python-client/blob/issue335_oidc-support-verification_uri_complete2/docs/auth.rst#device-flow

PR #363, opened 2023-01-12
pre-commit config: Upgrade darker to version that avoids bug in black

JohanKJSchreurs:
Now that the [darker](https://github.com/akaihola/darker) 1.6 has been released it might be a good time to update the pre-commit config file.

Previously we needed to pin [black ](https://github.com/psf/black) to an older version because of a bug : https://github.com/akaihola/darker/issues/403

With darker 1.6.1 that problem should be solved.


soxofaan:
thanks!

Issue #361, opened 2023-01-10
move MultiBackendJobManager into python client


jdries:
This class:
https://github.com/openEOPlatform/openeo-classification/blob/main/src/openeo_classification/job_management.py#L129
should be moved into openeo python client

soxofaan:
post-processed and merged PR #364  in 10c8e56

PR #364, opened 2023-01-13
#361 Moved MultiBackendJobManager into python client

JohanKJSchreurs:
Implements one part of issue https://github.com/Open-EO/openeo-python-client/issues/361 , the side of the Python client.

We will eventually remove and replace the corresponding code in https://github.com/openEOPlatform/openeo-classification , but we will wait a bit to change that side.

To make those changes we would need a new release of openeo-python-client, and there may be other code relying on this that needs to be adapted and tested.


openeo/extra/job_management.py
```diff
@@ -0,0 +1,379 @@
...
+from openeo.rest import OpenEoApiError
+from openeo.util import deep_get
+
+from openeo.rest.connection import cached_connection
+
+
+_log = logging.getLogger(__name__)
+
+
+def run_jobs(
```
jdries, line 24: I believe we don't actually need this method? This was an older version, before we had MultibackendJobManager

openeo/extra/job_management.py
```diff
@@ -0,0 +1,379 @@
...
+    return status_df
+
+
+# TODO: Do we still need this function after moving the code to openeo-python-client?
+#   (https://github.com/openEOPlatform/openeo-classification/issues/361)
+#   It is never called in the python client, but it is called within openeo-classification,
+#   from main.py and spanish_sampling.py.
+#   If we do need to keep it in the client, then we need those resources in the python client as well.
+# TODO: invalid path reference (https://github.com/openEOPlatform/openeo-classification/issues/2)
+def create_or_load_job_statistics(path="resources/training_data/job_statistics.csv"):
```
jdries, line 134: no, not needed, please remove

openeo/extra/job_management.py
```diff
@@ -0,0 +1,379 @@
...
+
+                df.to_csv(outputFile, index=False)
+            else:
+                time.sleep(60)
+
+        except requests.exceptions.ConnectionError as e:
+            _log.warning(f"Skipping connection error: {e}")
+
+
+def running_jobs(status_df):
```
jdries, line 93: this method and the next one can also be removed, not used by job manager

openeo/extra/job_management.py
```diff
@@ -0,0 +1,379 @@
...
+        jobs_df = pd.DataFrame(....)
+        output_file = Path("jobs.csv")
+        def start_job(row, connection, **kwargs):
+            ...
+
+        manager.run_jobs(df=df, start_job=start_job, output_file=output_file)
+
+    """
+
+    def __init__(self, poll_sleep=60):
```
jdries, line 176: please add constructor docs

openeo/extra/job_management.py
```diff
@@ -0,0 +1,379 @@
...
+        self.backends: Dict[str, _Backend] = {}
+        self.poll_sleep = poll_sleep
+
+    def add_backend(
+        self,
+        name: str,
+        connection: Union[Connection, Callable[[], Connection]],
+        parallel_jobs=2,
+    ):
+        """Register a backend with a name and a Connection getter"""
```
jdries, line 186: add basic docs for parameters

openeo/extra/job_management.py
```diff
@@ -0,0 +1,379 @@
...
+        new_columns = {
+            col: val for (col, val) in required_with_default if col not in df.columns
+        }
+        df = df.assign(**new_columns)
+        # Workaround for loading of geopandas "geometry" column.
+        if "geometry" in df.columns and df["geometry"].dtype.name != "geometry":
+            df["geometry"] = gpd.GeoSeries.from_wkt(df["geometry"])
+        return df
+
+    def run_jobs(self, df: pd.DataFrame, start_job: Callable, output_file: Path):
```
jdries, line 215: add docstring

openeo/extra/job_management.py
```diff
@@ -0,0 +1,379 @@
...
+                            persists(df, output_file)
+
+            time.sleep(self.poll_sleep)
+
+    def on_job_done(self, job: BatchJob, row):
+        """
+        Handles jobs that have finished. Can be overridden to provide custom behaviour.
+
+        Default implementation downloads the results into a folder containing the title.
+        @param job:
```
jdries, line 313: change docstring convention (python client doesn't use @param)

openeo/extra/job_management.py
```diff
@@ -0,0 +1,379 @@
...
+        @return:
+        """
+        job_metadata = job.describe_job()
+        job.get_results().download_files(target=job_metadata["title"])
+        with open(Path(job_metadata["title"]) / f"job_{job.job_id}.json", "w") as f:
+            json.dump(job_metadata, f, ensure_ascii=False)
+
+    def on_job_error(self, job: BatchJob, row):
```
jdries, line 321: add basic docstring

openeo/rest/connection.py
```diff
@@ -1355,3 +1356,37 @@ def paginate(con: Connection, url: str, params: dict = None, callback: Callable
         page += 1
         params = {}
 
+
+
+#
+# For caching connections
+#
+
+_default_url = "openeo-dev.vito.be"
```
jdries, line 1365: Avoid adding these things here, they're not strictly necessary for the job manager, and too specific for certain backends.

openeo/extra/job_management.py
```diff
@@ -0,0 +1,379 @@
...
+from typing import Callable, Union, Dict, Optional
+
+import geopandas as gpd
+import pandas as pd
+import requests
+from openeo import Connection, BatchJob
+from openeo.rest import OpenEoApiError
+from openeo.util import deep_get
+
+from openeo.rest.connection import cached_connection
```
jdries, line 18: import can be removed if we remove a few methods that are not needed

jdries (changes requested):
review done, main thing is to ensure we have basic api docs, and remove a few things that are not needed by the jobmanager class.

openeo/rest/connection.py
soxofaan, line 1365: we can't put a VITO specific URL like this here

openeo/rest/connection.py
```diff
@@ -1355,3 +1356,37 @@ def paginate(con: Connection, url: str, params: dict = None, callback: Callable
...
+    @return:
+    """
+    return _cached_connection(url)
+
+
+# TODO: I think we can drop these partial functions. We can still define them in openeo-classification if we want.
+#   That is probably a more appropriate place for them.
+terrascope_dev = partial(cached_connection, "openeo-dev.vito.be")
+openeo_platform = partial(cached_connection, "openeo.cloud")
+creo = partial(cached_connection, "openeo-dev.creo.vito.be")
```
soxofaan, line 1392: same as above: these specific back-end urls don't belong in this general connection.py module

openeo/extra/job_management.py
```diff
@@ -0,0 +1,379 @@
...
+import contextlib
+import datetime
+import json
+import logging
+import os
+import time
+from pathlib import Path
+from typing import Callable, Union, Dict, Optional
+
+import geopandas as gpd
```
soxofaan, line 11: geopandas is not a listed dependency of the client

openeo/extra/job_management.py
```diff
@@ -0,0 +1,263 @@
...
+            ("duration", None),
+            ("backend_name", None),
+        ]
+        new_columns = {
+            col: val for (col, val) in required_with_default if col not in df.columns
+        }
+        df = df.assign(**new_columns)
+        # Workaround for loading of geopandas "geometry" column.
+        if "geometry" in df.columns and df["geometry"].dtype.name != "geometry":
+            df["geometry"] = gpd.GeoSeries.from_wkt(df["geometry"])
```
jdries, line 84: this is the only location where geopandas is really used. Can you try decoding wkt with:
https://shapely.readthedocs.io/en/stable/manual.html?highlight=wkt#shapely.wkt.loads
This way, we get a very similar geometry column without requiring geopandas.

Other proposal: try to import geopandas and only apply conversion if import succeeds?

jdries: reason is that we try to have as few (hard) dependencies as possible, to have a very light installation

jdries:
I think we can merge at this point afaic

JohanKJSchreurs:
> I think we can merge at this point afaic

Hold on. There are errors on python 3.6. I'm fixing it.

openeo/extra/job_management.py
JohanKJSchreurs: Has been removed

openeo/extra/job_management.py
JohanKJSchreurs: Has been removed.

openeo/extra/job_management.py
JohanKJSchreurs: Docstring added.

openeo/extra/job_management.py
JohanKJSchreurs: Has been removed.

openeo/extra/job_management.py
JohanKJSchreurs: Docstring added.

openeo/extra/job_management.py
JohanKJSchreurs: Docstring added.

openeo/extra/job_management.py
JohanKJSchreurs: Docstring updated

openeo/rest/connection.py
JohanKJSchreurs: Code for connection caching was no longer relevant, so I removed it.

JohanKJSchreurs: Code for connection caching was no longer relevant, so I removed it.

openeo/rest/connection.py
```diff
@@ -1355,3 +1356,37 @@ def paginate(con: Connection, url: str, params: dict = None, callback: Callable
...
+def set_backend(url: str):
+    _default_url = url
+
+
+@lru_cache(maxsize=None)
+def _cached_connection(url: str) -> Connection:
+    return openeo.connect(url).authenticate_oidc()
+
+
+def cached_connection(url: Optional[str] = _default_url) -> Connection:
```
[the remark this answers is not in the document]
JohanKJSchreurs: Code for connection caching was no longer relevant, so I removed it.

openeo/extra/job_management.py
JohanKJSchreurs: Followed the advice of @jdries to use shapely for converting the geometry column. So we no longer need geopandas as a installation dependency (It was part of the testing dependencies though)

openeo/extra/job_management.py
JohanKJSchreurs: Done: using shapely for converting the geometry column.

JohanKJSchreurs:
> > I think we can merge at this point afaic
> 
> Hold on. There are errors on python 3.6. I'm fixing it.

Ok, build checks are passing again. Problem should be fixed.

openeo/rest/connection.py
JohanKJSchreurs: This code has been removed.

openeo/extra/job_management.py
JohanKJSchreurs: Import removed.

setup.py
```diff
@@ -18,61 +18,63 @@
     long_description = fh.read()
 
 tests_require = [
-    'pytest>=4.5.0',
```
soxofaan, line 21: Is there anything that actually changed here in setup.py?
if not, can we avoid that it shows up in this PR?
It now pollutes the history of this file quite a bit

openeo/extra/job_management.py
```diff
@@ -0,0 +1,290 @@
...
+            df["geometry"] = df["geometry"].apply(shapely.wkt.loads)
+        return df
+
+    # TODO: long method with deep nesting. Refactor it to make it more readable.
+    def run_jobs(self, df: pd.DataFrame, start_job: Callable, output_file: Path):
+        """Runs jobs, specified in a dataframe, and tracks parameters.
+
+        :param df:
+            DataFrame that specifies the jobs, and tracks the jobs' statuses.
+        :param start_job:
```
soxofaan, line 109: From reading the implementation it seems that `start_job` should actually be called `create_job`, right @jdries ?

also the documentation should state that this callable is expected to return a `openeo.rest.job.BatchJob` object

openeo/extra/job_management.py
```diff
@@ -0,0 +1,290 @@
...
+
+        Default implementation downloads the results into a folder containing the title.
+
+        :param job: The job that has finished.
+        :param row: DataFrame row containing the job's metadata.
+        """
+        # TODO: param `row` is never accessed in this method. Remove it? Is this intended for future use?
+
+        job_metadata = job.describe_job()
+        job.get_results().download_files(target=job_metadata["title"])
```
soxofaan, line 218: couple of things:
- This downloads to a folder directly under working dir. I think it's better to have an option to set a different root dir, e.g. through argument in `MultiBackendJobManager.__init__` (defaulting to `"."`)
- It downloads to subfolder based on title, but this creates high risk on overwriting results (if user is not careful about job titles). I'd add a prefix, e.g. using the first 8 or 10 chars of the job id

openeo/extra/job_management.py
```diff
@@ -0,0 +1,290 @@
...
+        """
+        # TODO: param `row` is never accessed in this method. Remove it? Is this intended for future use?
+
+        logs = job.logs()
+        error_logs = [l for l in logs if l.level.lower() == "error"]
+        job_metadata = job.describe_job()
+
+        title = job_metadata["title"]
+        if len(error_logs) > 0:
+            f"job_{title}_errors.json".write_text(json.dumps(error_logs, indent=2))
```
soxofaan, line 239: I don't think `str.write_text` works, I guess a `Path()` around the str is missing.

also, instead of writing to working dir, I would use a configurable root dir from `__ini__` as described above

openeo/extra/job_management.py
```diff
@@ -0,0 +1,290 @@
...
+                    self.on_job_done(the_job, df.loc[i])
+                if df.loc[i, "status"] != "error" and job_metadata["status"] == "error":
+                    self.on_job_error(the_job, df.loc[i])
+
+                df.loc[i, "status"] = job_metadata["status"]
+                for key in job_metadata.get("usage", {}).keys():
+                    df.loc[i, key] = _format_usage_stat(job_metadata, key)
+
+            except OpenEoApiError as e:
+                print(f"error for {backend_name}")
```
soxofaan, line 272: This error printing should also contain `i`, or `job_id` I think to be more useful

tests/extra/test_job_management.py
```diff
@@ -0,0 +1,161 @@
...
+    Preventing it would be better, but sometimes that is difficult.
+    TODO: What if we want to keep those files for inspecting after the test run, but preferably in a separate folder?
+    """
+
+    orig_workdir: Path = Path.cwd()
+    # Make the CWD a bit more specific because test can use tmp_dir for other things too.
+    temp_word_dir: Path = tmp_path / "test-workdir"
+    if not temp_word_dir.exists():
+        temp_word_dir.mkdir()
+    os.chdir(temp_word_dir)
```
soxofaan, line 27: Having to use `os.chdir` juggling from all tests feels like code smell to me. 
As mentioned above, I would make sure `MultiBackendJobManager` takes a root/working dir option to work in

tests/extra/test_job_management.py
```diff
@@ -0,0 +1,161 @@
...
+                "year": [2018, 2019, 2020, 2021, 2022],
+                # Use simple points in WKT format to test conversion to the geometry dtype
+                "geometry": ["POINT (1 2)"] * 5,
+            }
+        )
+        output_file = tmp_path / "jobs.csv"
+
+        def start_job(row, connection, **kwargs):
+            year = row["year"]
+            return RESTJob(job_id=f"job-{year}", connection=connection)
```
soxofaan, line 105: FTY: `RESTJob` is deprecated, use `BatchJob` instead

soxofaan (commented):
Some more notes from me

setup.py
soxofaan: (I can also strip it out when merging the PR, I just want to be sure nothing actually changed)

JohanKJSchreurs: No, there are no real changes anymore in setup.py. The only change left is formatting by black/darker.
So let's keep this file out of the PR.

Initially I made the geopandas test dependency an installation dependency, but I reverted that change.
However that triggered black/darker.

But if we want to switch to black's formatting rules, I think it is better to do that as a single commit (or a PR even) that *only* changes that, so we can easily see that the only change was formatting.


I propose to do a `git checkout master setup.py` to restore the file, but it would still be in the commit history though.
So you may want to purge it from the PR anyway.

Nevertheless, it would be good to restore it to make the CI doublecheck that everything still works.

tests/extra/test_job_management.py
JohanKJSchreurs: Indeed this workaround is a code smell. We need to be able to set the directory where it stores files.

openeo/extra/job_management.py
soxofaan: nevermind, the start_job callable is apparently also allowed to start the job already

setup.py
soxofaan: indeed, file-wide formatting fixing commits should not contain anything else.
Which can then be ignored in blame/history views through this "ignore-revs-file" feature (see https://www.git-scm.com/docs/git-blame#Documentation/git-blame.txt---ignore-revs-fileltfilegt and https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view )

I'll purge it when merging the PR using interactive rebase

openeo/extra/job_management.py
```diff
@@ -101,18 +118,22 @@ def _normalize_df(self, df: pd.DataFrame) -> pd.DataFrame:
...
+    def run_jobs(
+        self, df: pd.DataFrame, start_job: Callable[[], BatchJob], output_file: Path
+    ):
         """Runs jobs, specified in a dataframe, and tracks parameters.
 
         :param df:
             DataFrame that specifies the jobs, and tracks the jobs' statuses.
         :param start_job:
             A callback which will be invoked with the row of the dataframe for which a job should be started.
+            This callable should return a openeo.rest.job.BatchJob object.
```
soxofaan, line 130: FYI: you can user RestructuredText references like `:py:class:`openeo.rest.job.BatchJob`  which will turn in hyperlinks in the generated API docs

soxofaan:
maybe also add a simple API documentation page like done in https://github.com/Open-EO/openeo-python-client/blob/master/docs/cookbook/spectral_indices.rst  (added to doc tree from https://github.com/Open-EO/openeo-python-client/blob/master/docs/cookbook/index.rst)

openeo/extra/job_management.py
JohanKJSchreurs: Added to docstring that startt_job should return openeo.rest.job.BatchJob and updated the type hint accordingly.

openeo/extra/job_management.py
JohanKJSchreurs: - Added the option for the root directory.
- Avoid jobs overwriting each other's data: In the end it was easier to use the only the job_id in the filenames, instead of a combination of job_id and title.
- To keep things more manageable it makes a subfolder per job, because there can also be result files which are downloaded when the job finishes.

openeo/extra/job_management.py
JohanKJSchreurs: Indeed this produced errors. Updated along with the other stuff related to configurable root directory.
I added a basic test case to check the error logging works.

openeo/extra/job_management.py
JohanKJSchreurs: Added the job_id to the f-string.

tests/extra/test_job_management.py
JohanKJSchreurs: Now the job manager support a configurable root directory.
Removed the fixture that wrangles the current working.

tests/extra/test_job_management.py
JohanKJSchreurs: Done.

soxofaan:
did some git massaging of PR (to remove edits on setup.py and connection.py) and merged in 10c8e56

thanks!

Issue #340, opened 2022-11-10
Easily access result through xarray (to_xarray)


m-mohr:
A result from a discussion in the Pangeo discourse: https://discourse.pangeo.io/t/pangeo-openeo-backend/815/12

It seems that a function such as `to_xarray` would be a nice convenience for Pangeo folks where you automatically get an `xarray.DataArray` for any result (batch job and synchronous).
Synchronous would likely need to go directly through xarray while for batch job results it may need intake-stac or so.

soxofaan:
is basically a duplicate of  #7 (one of the earliest feature request :smile: )  , but I'd close that one and keep #340 which is more concrete

PR #369, opened 2023-01-20
First steps to have more flexible Connection.execute response handling

soxofaan:
related to #7 #340 #338


soxofaan:
I'm intending to have this merged after 0.14.0 release

soxofaan:
@clausmichele   the idea here is to introduce a "handling" option to  `Connection.execute()` (which is called by `cube.execute()`)  which is "auto" by default, meaning that the actual implementation of `Connection.execute()` is allowed to choose whatever makes most sense: e.g. a loaded xarray data array/set in case of LocalConnection

soxofaan:
this relatively old PR doesn't cleanly apply anymore. I think it's best to close it for now. Can be kept as inspiration for a next attempt

PR #371, opened 2023-01-25
Feature branch2

Pratichhya:
Logging message in UDF


docs/udf.rst
```diff
@@ -402,6 +402,34 @@ An example code can be found `here <https://github.com/Open-EO/openeo-python-cli
 
 
 
+Logging from the UDFs
+==============================
+
+In some cases, users might want to use print statements within a UDF so that they can monitor their
```
jdries, line 408: 'users might want to' => I would address the user more directly:
In some cases, you may want to log information from your user defined function, for instance to provide debug information or to log warnings.

docs/udf.rst
```diff
@@ -402,6 +402,34 @@ An example code can be found `here <https://github.com/Open-EO/openeo-python-cli
 
 
 
+Logging from the UDFs
+==============================
+
+In some cases, users might want to use print statements within a UDF so that they can monitor their 
+results that makes debugging easier. One possible way is to use ``inspect`` logging function in the UDF.
```
jdries, line 409: 'inspect' is the only way: You can use the inspect logging function to achieve this.
Maybe also turn inspect into a link to the relevant api docs?

docs/udf.rst
```diff
@@ -402,6 +402,34 @@ An example code can be found `here <https://github.com/Open-EO/openeo-python-cli
...
+
+For example, in the previous example of rescaling an RGB image(from SENTINEL2_L2A), suppose user
```
jdries, line 411: suppose user want => suppose you want to

docs/udf.rst
```diff
@@ -402,6 +402,34 @@ An example code can be found `here <https://github.com/Open-EO/openeo-python-cli
...
+    from openeo.udf.debug import inspect
+
+    def apply_datacube(cube: XarrayDataCube, context: dict) -> XarrayDataCube:
+        array = cube.get_array()
+        inspect(data=[array.shape], message="UDF logging shape of my array")
+        array.values = 0.0001 * array.values
+        return cube
+    """)
+
+Then, for users using a Jupyter notebook you can use the ```job.logs()``` to request the job logs where you will
```
jdries, line 426: job.logs() will also work outside of jupyter notebooks, so can be generalized?

jdries: Or maybe specify that the nice rendering only happens in jupyter, while in plain python you will get a dict.

jdries (changes requested):

docs/udf.rst
Pratichhya: updated

Pratichhya:
updated

docs/data_access.rst
```diff
@@ -115,3 +115,24 @@ on the collection and the catalog that is used by it. This is not a problem caus
...
+However, if you attempt to execute a process on significantly large data, you would probably encounter a 'Request Entity Too Large' 
+error because your 'POST' request was too large for the system to handle. In order to avoid this error, you can upload
+your vector data to a public location(e.g., via Google Drive) and use it as an URL. The data stored in the URL can be loaded using 
+:py:meth:`~openeo.rest.connection.Connection.vectorcube_from_paths`
+
+The code-snippets shown below provides an example of how this can be achieved:
+
+.. code-block:: python
+
+    parcels = connection.vectorcube_from_paths(["https://artifactory.vgt.vito.be/testdata-public/parcels/32TPT.pq"],format="parquet")
```
soxofaan, line 131: I would avoid putting a real VITO (artifactory) URL here, just use a dummy URL, like `https://example.com/parcels.pq`

docs/data_access.rst
```diff
@@ -115,3 +115,24 @@ on the collection and the catalog that is used by it. This is not a problem caus
...
+    datacube = connection.load_collection("SENTINEL1_GAMMA0_SENTINELHUB",
+                temporal_extent = ["2021-05-12",'2021-06-01'])
+    s1_cube = datacube.aggregate_spatial(geometries=parcels,
+                           reducer="mean")
```
soxofaan, line 135: please try to follow the code style guidelines for example snippets in the docs: https://open-eo.github.io/openeo-python-client/best_practices.html#line-length-management

docs/data_access.rst
```diff
@@ -115,3 +115,24 @@ on the collection and the catalog that is used by it. This is not a problem caus
 standardization between catalogs of EO data.
 
 
+Handling large vector data
+#######################
```
soxofaan, line 119: I think you're missing a couple of `#` to make the heading syntax work

docs/data_access.rst
```diff
@@ -115,3 +115,24 @@ on the collection and the catalog that is used by it. This is not a problem caus
...
+
+Handling large volumes of data is crucial for making decisions, improving processes and having an efficient solution. 
+However, if you attempt to execute a process on significantly large data, you would probably encounter a 'Request Entity Too Large'
```
soxofaan, line 122: > attempt to execute a process on significantly large data, you would probably encounter a 'Request Entity Too Large'

I think this is a bit misleading: openEO, by design, allows to work on large (raster) data sets. The problem is that your REST requests should not be too large (e.g. avoid embedding large vector data sets directly in the process graph)

docs/data_access.rst
jdries: or maybe a real url on github? Maybe also working with a smaller file, so that it can be tested easily without running a very large job.

Pratichhya:
updated

soxofaan:
that added example parquet file is pretty large (4795 rows, covering relatively large area), could you reduce that (less rows, small spatial extent)?

Pratichhya:
added a smaller parquet file. But haven't executed yet, so I am not sure if it worked well similar to a Geoparquet.

soxofaan:
you replaced `parquet.pq` with two new files `Selected_aoi.parquet` and `aoi.pq` but I can read neither of them with `geopandas.read_parquet()` (the original `parquet.pq` was readable that way)

Pratichhya:
> you replaced `parquet.pq` with two new files `Selected_aoi.parquet` and `aoi.pq` but I can read neither of them with `geopandas.read_parquet()` (the original `parquet.pq` was readable that way)

Maybe the way I created the parquet file went wrong. I updated it to geoparquet file. Can you please check again? Actually, I also run into error when using `geopandas.read_parquet()` but when trying to use it with pyarrow it shows the result. Sorry I am a bit confused myself.

docs/data_access.rst
soxofaan: I see you  updated it, but I think it should be more like this (following the [black code style](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html)):
```python
    parcels = connection.vectorcube_from_paths(
        ["https://github.com/Open-EO/openeo-python-client/blob/master/tests/data/Selected_aoi.parquet"],
        format="parquet",
    )
    datacube = connection.load_collection(
        "SENTINEL2_L2A",
        ...
```

Pratichhya: Sorry I missed that again. Thank you for reminding

soxofaan:
I still fail to read latest `tests/data/example_aoi.geoparquet` with `geopandas.read_parquet`:
```
ValueError: Missing geo metadata in Parquet/Feather file.
            Use pandas.read_parquet/read_feather() instead.
```

Note that `geopandas.read_parquet()` is also used backend-side in the VITO back-end, so it seems quite important to get it working with that

Pratichhya:
> I still fail to read latest `tests/data/example_aoi.geoparquet` with `geopandas.read_parquet`:
> 
> ```
> ValueError: Missing geo metadata in Parquet/Feather file.
>             Use pandas.read_parquet/read_feather() instead.
> ```
> 
> Note that `geopandas.read_parquet()` is also used backend-side in the VITO back-end, so it seems quite important to get it working with that

Oh, ok. I will try to fix it, for sure I guess it is how I am generating the file.

soxofaan:
Note that I pushed some finetuning commits to the feature branch in the mean time, so be sure to first pull before continuing working on the branch

Pratichhya:
> > I still fail to read latest `tests/data/example_aoi.geoparquet` with `geopandas.read_parquet`:
> > ```
> > ValueError: Missing geo metadata in Parquet/Feather file.
> >             Use pandas.read_parquet/read_feather() instead.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Note that `geopandas.read_parquet()` is also used backend-side in the VITO back-end, so it seems quite important to get it working with that
> 
> Oh, ok. I will try to fix it, for sure I guess it is how I am generating the file.

@soxofaan I update the file under the same name, and it seems to work with `geopandas.read_parquet()` as well. Can you please confirm and let me know if it went fine whenever you are free(no hurry :) )

soxofaan:
ok, squashed feature branch, finetuned some more and merged in 155a174faa2531b71805a49705cbd50ff4f8a829

Issue #365, opened 2023-01-15
JobManager: add resilience for backend failures


jdries:
Backends have all kind of intermittent failures that go away over time. This usually means they send 50x errors.
Especially for the job manager, it can be better to keep trying until the backend works again. The python requests library can be configured to do this. 
This way, long running tasks are more resilient.

jdries:
FYI, my current approach to add retries in connection.py:
```

        from requests.adapters import HTTPAdapter, Retry
        retries = Retry(total=5,read=50,other=50,status=50,
                        backoff_factor=0.1,
                        status_forcelist=[ 502, 503, 504, 404],
                        method_whitelist=["HEAD", "GET", "OPTIONS","POST"])
        self.session.mount('https://', HTTPAdapter(max_retries=retries))
        self.session.mount('http://', HTTPAdapter(max_retries=retries))
```

PR #373, opened 2023-01-26
Issue #365 Add resilience to job manager

JohanKJSchreurs:
Implements issue https://github.com/Open-EO/openeo-python-client/issues/365

### DIFF
--- a/openeo/extra/job_management.py
+++ b/openeo/extra/job_management.py
@@ -90,9 +90,9 @@ class MultiBackendJobManager:
             Maximum number of jobs to allow in parallel on a backend.
         """
 
-        # TODO: Code could be simpler if _Backend is a class and this logic is move there.
-        # We would need to keep add_backend here as part of the public API though.
-        # But the amount of unrelated "stuff to manage" would be less (better cohesion)
+        # TODO: Code might become simpler if we turn _Backend into class move this logic there.
+        #   We would need to keep add_backend here as part of the public API though.
+        #   But the amount of unrelated "stuff to manage" would be less (better cohesion)
         if isinstance(connection, Connection):
             c = connection
             connection = lambda: c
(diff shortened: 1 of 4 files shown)


openeo/extra/job_management.py
```diff
@@ -95,6 +101,57 @@ def add_backend(
...
+        retries = Retry(
+            total=5,
+            read=50,
+            other=50,
+            status=50,
+            backoff_factor=0.1,
+            status_forcelist=status_forcelist,
+            allowed_methods=["HEAD", "GET", "OPTIONS", "POST"],
+        )
+        connection.session.mount("https://", HTTPAdapter(max_retries=retries))
```
jdries, line 152: Here we rely on the fact that our Connection implementation is based on requests, but not sure how to improve this. I would perhaps do it like this for now.

openeo/extra/job_management.py
```diff
@@ -121,17 +178,40 @@ def _normalize_df(self, df: pd.DataFrame) -> pd.DataFrame:
...
+                    Like connection in add_backend:
+                    - either a Connection to the backend,
+                    - or a callable to create a backend connection.
+
+                connection:
+                    The Connection itself, that has already been created.
+
+                provider:
+                    The name of the backend that will run the job.
+
```
JohanKJSchreurs, line 214: 1) The way we define what start_job should look like is a bit awkward. I think start_job would be a good candidate for using a Protocol, that defines and documents what the expected / supported function's parameters are, but Protocol was only introduced in Python 3.8.
Maybe a TODO for when we drop support for python 3.7 and 3.6?

2) That said we need to document what parameters the job manager passes to start_jobs so we know what parameters you can (maximally) have in your implementation.
This way to document it is a little clunky but I think enough for now.
Suggestions for a better solution are welcome.

3) We allow two similar parameters which is a little confusion: both a Connection as well as a callback to create a connection.
But I think the reason was that in practice the implementation would use either one or the other, if it is using them at all. If this is correct, that can help improve to improve the docstring.

Also, it looks like the intention is that start_job always uses keyword arguments and `*args, **kwargs`, so you can simply leave out parameters that you aren't going to use, even though the job manager will pass them to start_jobs.

JohanKJSchreurs: I updated the docstring to make it more clear about the points above.
That is probably good enough for now.

jdries:
looks good, merging!

Issue #332, opened 2022-10-05
Less verbose log printing on failed batch job


soxofaan:
https://github.com/Open-EO/openeo-python-client/blob/a184da6354323c21338e8cab7533bd927e8fc1e5/openeo/rest/job.py#L226-L235

This automatic log printing when a batch job failed  is currently very verbose (up to debug level).
I have notebooks with failed jobs that have thousands of lines of this log dumping. That's not very practical to work with (e.g. copy-paste it to the support forum)

I guess we  should only show warn and error level by default, and show instructions to get more detailed logs

soxofaan:
@jdries what do you think?

jdries:
Indeed, filtering on even only 'ERROR' logs would make more sense.

soxofaan:
merged #374

PR #374, opened 2023-02-01
Issue332 less verbose log printing on failed batch job

JohanKJSchreurs:
Implements https://github.com/Open-EO/openeo-python-client/issues/332


jdries:
looks good for me

tests/rest/test_job.py
```diff
@@ -264,6 +265,124 @@ def test_get_job_logs(session040, requests_mock):
...
+    )
+
+    log_entries = session040.job("f00ba5").logs()
+
+    assert len(log_entries) == 1
+    assert log_entries[0].message == "error processing batch job"
+
+
+@pytest.mark.parametrize(
+    "log_level,exp_num_messages",
```
soxofaan, line 305: FYI: you don't have to comma-serialize multiple parameters  as single string, you can just provide a list `["log_level", "exp_num_messages"]`

openeo/rest/job.py
```diff
@@ -136,14 +136,84 @@ def get_results(self) -> "JobResults":
...
+        log_level = BatchJob._normalize_log_level(log_level)
+        entries = [
+            LogEntry(log)
+            for log in logs
+            if BatchJob._string_to_log_level(log["level"]) >= log_level
+        ]
         return VisualList('logs', data=entries)
 
+    @classmethod
+    def _normalize_log_level(cls, log_level: Optional[Union[int, str]]) -> int:
```
soxofaan, line 176: I'd move this normalization logic to `openeo/api/logs.py`, because logs are not only for batch jobs (secondary services too)

soxofaan (commented):
some minor remarks

tests/rest/test_job.py
JohanKJSchreurs: Done.

openeo/rest/job.py
JohanKJSchreurs: Done.

openeo/rest/job.py
```diff
@@ -136,12 +136,40 @@ def get_results(self) -> "JobResults":
...
+
+            You can use either constants from Python's standard module ``logging``
+            or their names (case insensitive).
+
+            For example:
+                ``logging.INFO``, ``"info"`` or ``"INFO"`` can all be used to show the messages
+                for level ``logging.INFO`` and above, i.e. also ``logging.WARNING`` and
+                ``logging.ERROR`` will be included.
+
+            Defaults to ``logging.ERROR`` (also when explicitly passing log_level=None or log_level="").
```
soxofaan, line 161: Hmm, thinking about this some more, this means that we are changing the original  behavior here and diverging from the behavior implied by the API. 
I think the default behavior of `job.logs()` should be to return all logs (like it originally was) instead of only returning error logs.
Also note that the related jupyter widget also supports on-the-fly filtering, so we should not filter too much up front.


I think we should only limit  the logs printed from a failing batch job, and leave the `.logs()` method untouched.

JohanKJSchreurs: The original behavior, returning all logs, is the default again. 

I also adapted the helper functions `normalize_log_level` and ` string_to_log_level` to return `logging.DEBUG` as a default. I think that also for these helper functions it is better to have a default that will not cause any surprises (when new code eventually starts to use those functions), and `logging.DEBUG` would be the value that works like a no-operation. Seems safer.

openeo/rest/job.py
```diff
@@ -230,9 +262,13 @@ def soft_error(message: str):
...
             # TODO: render logs jupyter-aware in a notebook context?
-            # TODO: only print the error level logs? Or the tail of the logs?
             print("Printing logs:")
-            print(self.logs())
+            print("Only printing messages of log level ERROR.")
+            print(
+                "To get logs for all log levels, use the method `BatchJob.logs` with the parameter `log_level` ."
+            )
+            print('For example: my_batch_job.logs(log_level="info")')
+            print(self.logs(log_level=logging.ERROR))
```
soxofaan, line 271: 5 `print` statements looks a bit much I think. The first two can be merged for example. 
Third and fourth can be combined more compactly too

openeo/api/logs.py
```diff
@@ -46,3 +54,62 @@ def level(self):
         return self["level"]
 
     # TODO: add properties for "code", "time", "path", "links" and "data" with sensible defaults?
+
+
+def normalize_log_level(log_level: Optional[Union[int, str]]) -> int:
```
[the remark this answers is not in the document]
JohanKJSchreurs: Indeed. Added the default value.

openeo/rest/job.py
JohanKJSchreurs: I tried to come up with a more concise message that is still as informative as possible, and bundled it in one cleaner print statement.
Let me know what you think.

openeo/rest/job.py
```diff
@@ -9,7 +9,7 @@
 
 import requests
 
-from openeo.api.logs import LogEntry
+from openeo.api.logs import LogEntry, normalize_log_level, string_to_log_level
```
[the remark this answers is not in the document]
JohanKJSchreurs: That was bothering me too. I initially kept ``string_to_log_level`` because ``BatchJob.logs`` was calling it directly.
But that wasn't really necessary anymore.

I decided to eliminate ``string_to_log_level`` entirely because it helped to make the code simpler and cleaner overall.

openeo/rest/job.py
soxofaan: I merged the PR already  but then  noticed there is even more duplication with a couple of lines above.
did a follow up in 0ceced3ba685c0d789702b3c92e9de94e2692e9d

Issue #377, opened 2023-02-17
Incomplete /files support?


m-mohr:
It seems the Python client supports `GET /files` to list all user-uploaded files, but I can't find any information about the other file-related endpoints. How can users upload/download and delete files?

soxofaan:
indeed, file support in python client is very limited atm

soxofaan:
(rebase+) merged #378  in d16417c98290c98ffea0efc4b463374ed9ea6b14

PR #378, opened 2023-02-17
Support for the /files endpoints #377

m-mohr:
Implements #377

It seems the Python client only allowed to list files so far, which was not very useful. Added full support for the /files endpoints.

What do you think, @soxofaan ?
It's available in all clients except for Python so I think it should be added to push towards full openEO API support.

Interface looks like this (filenames/paths don't make any sense...):
image

Has been tested successfully against the GEE back-end.

### DIFF
--- a/openeo/internal/jupyter.py
+++ b/openeo/internal/jupyter.py
@@ -102,6 +102,8 @@ def render_component(component: str, data = None, parameters: dict = None):
     # Set the data as the corresponding parameter in the Vue components
     key = COMPONENT_MAP.get(component, component)
     if data is not None:
+        if isinstance(data, list):
+            data = [(x.to_dict() if "to_dict" in dir(x) else x) for x in data]
         parameters[key] = data
 
     # Construct HTML, load Vue Components source files only if the openEO HTML tag is not yet defined
--- a/openeo/internal/jupyter.py
+++ b/openeo/internal/jupyter.py
@@ -103,7 +103,7 @@ def render_component(component: str, data = None, parameters: dict = None):
     key = COMPONENT_MAP.get(component, component)
     if data is not None:
         if isinstance(data, list):
-            data = [(x.to_dict() if "to_dict" in dir(x) else x) for x in data]
+            data = [(x.to_dict() if hasattr(x, "to_dict") else x) for x in data]
         parameters[key] = data
 
     # Construct HTML, load Vue Components source files only if the openEO HTML tag is not yet defined
--- a/openeo/rest/connection.py
+++ b/openeo/rest/connection.py
@@ -1089,9 +1089,9 @@ class Connection(RestApiConnection):
         """Get batch job logs."""
         return BatchJob(job_id=job_id, connection=self).logs(offset=offset)
 
-    def list_files(self):
+    def list_files(self) -> List[UserFile]:
         """
-        Lists all files that the logged in user uploaded.
+        Lists all files that the logged-in user uploaded.
 
         :return: file_list: List of the user-uploaded files.
         """
@@ -1100,16 +1100,20 @@ class Connection(RestApiConnection):
         files = [UserFile(file['path'], connection=self, metadata=file) for file in files]
         return VisualList("data-table", data=files, parameters={'columns': 'files'})
 
-    def get_file(self, path: str) -> UserFile:
+    def get_file(self, path: Union[str, PurePosixPath]) -> UserFile:
         """
         Gets a handle to a file in the user workspace on the back-end.
 
-        :param path: The  path on the user workspace.
+        :param path: The path on the user workspace.
         :return: UserFile object.
         """
         return UserFile(path, connection=self)
 
-    def upload_file(self, source: Union[Path, str], target: str = None) -> UserFile:
+    def upload_file(
+        self,
+        source: Union[Path, str],
+        target: Optional[Union[str, PurePosixPath]] = None,
+    ) -> UserFile:
         """
         Uploads a file to the given target location in the user workspace.
 
@@ -1120,7 +1124,7 @@ class Connection(RestApiConnection):
         :return: UserFile object.
         """
         if target is None:
-            target = PurePosixPath(source).name
+            target = Path(source).name
         return self.get_file(target).upload(source)
 
     def _build_request_with_process_graph(self, process_graph: Union[dict, Any], **kwargs) -> dict:
--- a/openeo/rest/connection.py
+++ b/openeo/rest/connection.py
@@ -33,6 +33,7 @@ from openeo.rest.auth.oidc import OidcClientCredentialsAuthenticator, OidcAuthCo
 from openeo.rest.datacube import DataCube
 from openeo.rest.imagecollectionclient import ImageCollectionClient
 from openeo.rest.mlmodel import MlModel
+from openeo.rest.userfile import UserFile
 from openeo.rest.job import BatchJob, RESTJob
 from openeo.rest.rest_capabilities import RESTCapabilities
 from openeo.rest.service import Service
@@ -1092,20 +1093,20 @@ class Connection(RestApiConnection):
         """
         Lists all files that the logged in user uploaded.
 
-        :return: file_list: List of the user uploaded files.
+        :return: file_list: List of the user-uploaded files.
         """
 
         files = self.get('/files', expected_status=200).json()['files']
+        files = [UserFile(file['path'], self, file) for file in files]
         return VisualList("data-table", data=files, parameters={'columns': 'files'})
 
-    def create_file(self, path):
+    def get_file(self, path) -> UserFile:
         """
-        Creates virtual file
+        Gets a (virtual) file for the user workspace
 
-        :return: file object.
+        :return: UserFile object.
         """
-        # No endpoint just returns a file object.
-        raise NotImplementedError()
+        return UserFile(path, self)
 
     def _build_request_with_process_graph(self, process_graph: Union[dict, Any], **kwargs) -> dict:
         """
--- a/openeo/rest/connection.py
+++ b/openeo/rest/connection.py
@@ -5,7 +5,6 @@ import datetime
 import json
 import logging
 import shlex
-import os
 import sys
 import warnings
 from collections import OrderedDict
@@ -1103,7 +1102,7 @@ class Connection(RestApiConnection):
 
     def get_file(self, path: str) -> UserFile:
         """
-        Gets a (virtual) file for the user workspace
+        Gets a handle to a file in the user workspace on the back-end.
 
         :return: UserFile object.
         """
@@ -1116,7 +1115,7 @@ class Connection(RestApiConnection):
 
         :return: UserFile object.
         """
-        return self.get_file(os.path.basename(source)).upload(source)
+        return self.get_file(Path(source).name).upload(source)
 
     def _build_request_with_process_graph(self, process_graph: Union[dict, Any], **kwargs) -> dict:
         """
--- a/openeo/rest/connection.py
+++ b/openeo/rest/connection.py
@@ -1091,11 +1091,10 @@ class Connection(RestApiConnection):
 
     def list_files(self) -> List[UserFile]:
         """
-        Lists all files that the logged-in user uploaded.
+        Lists all user-uploaded files in the user workspace on the back-end.
 
-        :return: file_list: List of the user-uploaded files.
+        :return: List of the user-uploaded files.
         """
-
         files = self.get('/files', expected_status=200).json()['files']
         files = [UserFile.from_metadata(metadata=f, connection=self) for f in files]
         return VisualList("data-table", data=files, parameters={'columns': 'files'})
@@ -1104,10 +1103,9 @@ class Connection(RestApiConnection):
         self, path: Union[str, PurePosixPath], metadata: Optional[dict] = None
     ) -> UserFile:
         """
-        Gets a handle to a file in the user workspace on the back-end.
+        Gets a handle to a user-uploaded file in the user workspace on the back-end.
 
         :param path: The path on the user workspace.
-        :return: UserFile object.
         """
         return UserFile(path=path, connection=self, metadata=metadata)
 
@@ -1117,13 +1115,13 @@ class Connection(RestApiConnection):
         target: Optional[Union[str, PurePosixPath]] = None,
     ) -> UserFile:
         """
-        Uploads a file to the given target location in the user workspace.
+        Uploads a file to the given target location in the user workspace on the back-end.
 
         If a file at the target path exists in the user workspace it will be replaced.
 
         :param source: A path to a file on the local file system to upload.
-        :param target: The desired path on the user workspace. If not set, defaults to the filename (without path) of the local file.
-        :return: UserFile object.
+        :param target: The desired path on the user workspace.
+            If not set, defaults to the filename (without path) of the local file.
         """
         source = Path(source)
         target = target or source.name
--- a/openeo/rest/connection.py
+++ b/openeo/rest/connection.py
@@ -1097,7 +1097,7 @@ class Connection(RestApiConnection):
         """
 
         files = self.get('/files', expected_status=200).json()['files']
-        files = [UserFile(file['path'], self, file) for file in files]
+        files = [UserFile(file['path'], connection=self, metadata=file) for file in files]
         return VisualList("data-table", data=files, parameters={'columns': 'files'})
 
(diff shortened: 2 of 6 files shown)


openeo/rest/userfile.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def __repr__(self):
+        return '<{c} file={i!r}>'.format(c=self.__class__.__name__, i=self.path)
+
+    def _get_endpoint(self) -> str:
+        return "/files/{}".format(self.path)
+
+    def get_metadata(self, key) -> Any:
+        """ Get metadata about the file, e.g. file size (key: 'size') or modification date (key: 'modified')."""
+        if key in self.metadata:
```
soxofaan, line 26: FYI this whole if construct can be a idiomatic oneliner `return self.metadata.get(key)`  (`dict.get()` returns `None` as fallback on missing keys)

Sidenote: attribute `metadata` is public, so added value of `get_metadata()` is unclear in this regard: direct usage is  more intuitive and flexible than using the method:
```
file.metadata["size"]
# vs
file.get_metadata("size")
```

openeo/rest/userfile.py
```diff
@@ -0,0 +1,58 @@
...
+            return self.metadata[key]
+        else:
+            return None
+
+    def download_file(self, target: Union[Path, str]) -> Path:
```
soxofaan, line 31: the class is a "file thing", so method name can just be `download()` I think

openeo/rest/userfile.py
```diff
@@ -0,0 +1,58 @@
...
+        """ Downloads a user-uploaded file."""
+        # GET /files/{path}
+        response = self.connection.get(self._get_endpoint(), expected_status=200, stream=True)
+
+        path = Path(target)
```
soxofaan, line 36: I think it would be handy if the user can also just specify a folder as target ("." by default even), and the client just takes the remote file's name/path

for example see https://github.com/Open-EO/openeo-python-client/blob/d58650c4c39e486fda01252bece5ddc734a23f9e/openeo/rest/job.py#L307-L320

openeo/rest/userfile.py
```diff
@@ -0,0 +1,58 @@
...
+
+        return path
+
+
+    def upload_file(self, source: Union[Path, str]):
+        # PUT /files/{path}
+        """ Uploaded (or replaces) a user-uploaded file."""
+        path = Path(source)
+        with path.open(mode="rb") as f:
+            self.connection.put(self._get_endpoint(), expected_status=200, data=f)
```
soxofaan, line 49:
```suggestion
            self.connection.put(self._get_endpoint(), expected_status=200, data=f.read())
```

openeo/rest/userfile.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def delete_file(self):
+        """ Delete a user-uploaded file."""
+        # DELETE /files/{path}
+        self.connection.delete(self._get_endpoint(), expected_status=204)
+
+    def to_dict(self) -> Dict[str, Any]:
```
soxofaan, line 56: I don't understand what this method is for if there is already  the (public) `metadata` attribute

openeo/rest/connection.py
```diff
@@ -1092,20 +1093,20 @@ def list_files(self):
...
         """
-        Creates virtual file
+        Gets a (virtual) file for the user workspace
 
-        :return: file object.
+        :return: UserFile object.
         """
-        # No endpoint just returns a file object.
-        raise NotImplementedError()
+        return UserFile(path, self)
```
soxofaan, line 1109:
```suggestion
        return UserFile(path, connection=self)
```
to make it look less weird

openeo/rest/connection.py
```diff
@@ -1092,20 +1093,20 @@ def list_files(self):
         """
         Lists all files that the logged in user uploaded.
 
-        :return: file_list: List of the user uploaded files.
+        :return: file_list: List of the user-uploaded files.
         """
 
         files = self.get('/files', expected_status=200).json()['files']
+        files = [UserFile(file['path'], self, file) for file in files]
```
soxofaan, line 1100:
```suggestion
        files = [UserFile(file['path'], connection=self, metadata=file) for file in files]
```

openeo/rest/userfile.py
```diff
@@ -0,0 +1,58 @@
+import typing
+from typing import Any, Dict, List, Union
+from pathlib import Path
+
+if typing.TYPE_CHECKING:
+    # Imports for type checking only (circular import issue at runtime).
+    from openeo.rest.connection import Connection
```
soxofaan, line 7: This if-import is not necessary: you're not using `Connection` as a class 
(only as a type hint, but as a string, which works without import)

openeo/rest/userfile.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+class UserFile:
+    """Represents a file in the user-workspace of openeo."""
+
+    def __init__(self, path: str, connection: 'Connection', metadata: Dict[str, Any] = {}):
+        self.path = path
+        self.metadata = metadata
```
soxofaan, line 15: I'd make the argument `metadata` `None` by default (it's bad style to have mutable objects like dicts as default values) and fall back on only the path data:
```python
def __init__(self, path: str, connection: 'Connection', metadata: Optional[Dict[str, Any]] = None):
        self.path = path
        self.metadata = metadata or {"path": path}
```

that will simplify `to_dict` too

openeo/rest/userfile.py
```diff
@@ -0,0 +1,58 @@
...
+
+        path = Path(target)
+        with path.open(mode="wb") as f:
+            for chunk in response.iter_content(chunk_size=None):
+                f.write(chunk)
+
+        return path
+
+
+    def upload_file(self, source: Union[Path, str]):
```
soxofaan, line 44: I'd just call this `upload()`

soxofaan: Also it does not make sense I think to define this as a `UserFile` method (as you don't call this on an existing `UserFile` object, but you create a new one).
I think it should be a method on `Connection`

soxofaan (commented):
looks already good
a couple of quick notes

openeo/rest/userfile.py
m-mohr: It works without read. Wouldn't that also disable streaming?

openeo/rest/userfile.py
m-mohr: Copied from Service, so we should check whether it can also be removed there.

soxofaan: aha interesting
removed in 20e4c5e8a260a7ac45808ed0a145680245e226d4

openeo/rest/userfile.py
soxofaan: aha, indeed, didn't know about the feature that an open file can be passed there

openeo/rest/userfile.py
soxofaan: I stand corrected, linting with flake8 apparently requires this import even for string type hints.
https://github.com/Open-EO/openeo-python-client/actions/runs/4207089915/jobs/7301432268

Reverted in 32c70d03664eb74610e6f56067acc817556244c8

openeo/rest/userfile.py
m-mohr: As this can also be used to override an existing file I think it makes also sense to have it here.
I remembered you were mentioning "decluttering" the Connection object, which also seemed like a good argument to have it here, but I can add a convenience upload_file in the Connection if you want.

openeo/rest/userfile.py
m-mohr: Fine for me. Some languages prefer explicit getters/setters and I wasn't sure whether that's the case here.

openeo/rest/userfile.py
m-mohr: Fine for me. I just copied the style from the Service class where it's always ..._service, too. But without suffix is nicer indeed.

openeo/rest/userfile.py
m-mohr: It's used in the JupyerLab integration to correctly generate the output in a generalized way as the public members are named differently in other classes.

openeo/rest/userfile.py
m-mohr: Done

openeo/rest/userfile.py
m-mohr: Done

openeo/rest/userfile.py
```diff
@@ -0,0 +1,63 @@
...
+if typing.TYPE_CHECKING:
+    # Imports for type checking only (circular import issue at runtime).
+    from openeo.rest.connection import Connection
+
+
+class UserFile:
+    """Represents a file in the user-workspace of openeo."""
+
+    def __init__(self, path: str, connection: 'Connection', metadata: Dict[str, Any] = None):
+        self.path = path
```
soxofaan, line 16:
```suggestion
        self.path = Path(path)
```

openeo/rest/userfile.py
```diff
@@ -0,0 +1,63 @@
...
+        self.metadata = metadata or {"path": path}
+        self.connection = connection
+
+    def __repr__(self):
+        return '<{c} file={i!r}>'.format(c=self.__class__.__name__, i=self.path)
+
+    def _get_endpoint(self) -> str:
+        return "/files/{}".format(self.path)
```
soxofaan, line 24:
```suggestion
        return f"/files/{self.path!s}"
```

openeo/rest/userfile.py
```diff
@@ -0,0 +1,63 @@
...
+         :param target: download target path. Can be an existing folder 
+             (in which case the file name advertised by backend will be used) 
+             or full file name. By default, the working directory will be used.
+        """
+        # GET /files/{path}
+        response = self.connection.get(self._get_endpoint(), expected_status=200, stream=True)
+
+        target = Path(target or Path.cwd()) 
+        if target.is_dir():
+            target = target / os.path.basename(self.path)
```
soxofaan, line 39:
```suggestion
            target = target / self.path.name
```

And related to note higher: user might want to keep original folder structure, in which case it should be:
```suggestion
            target = target / self.path
```

openeo/internal/jupyter.py
```diff
@@ -102,6 +102,8 @@ def render_component(component: str, data = None, parameters: dict = None):
     # Set the data as the corresponding parameter in the Vue components
     key = COMPONENT_MAP.get(component, component)
     if data is not None:
+        if isinstance(data, list):
+            data = [(x.to_dict() if "to_dict" in dir(x) else x) for x in data]
```
soxofaan, line 106: it's more idiomatic (and more efficient I guess too) to do `if hasattr(x, "to_dict")`  than `if "to_dict" in dir(x)`

openeo/rest/userfile.py
soxofaan: hmm this is a well hidden connection then. Can you at least put a comment about this here?

openeo/rest/connection.py
```diff
@@ -5,6 +5,7 @@
 import json
 import logging
 import shlex
+import os
```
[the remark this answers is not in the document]
m-mohr:
```suggestion
```

openeo/rest/userfile.py
m-mohr: Does this create missing folders automatically?

openeo/internal/jupyter.py
m-mohr:
```suggestion
            data = [(x.to_dict() if hasattr(x, "to_dict") else x) for x in data]
```

openeo/rest/userfile.py
```diff
@@ -0,0 +1,63 @@
+import typing
+from typing import Any, Dict, Union
+import os
```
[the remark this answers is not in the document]
m-mohr:
```suggestion
```

openeo/rest/connection.py
```diff
@@ -1092,20 +1094,29 @@ def list_files(self):
...
+        :return: UserFile object.
         """
-        # No endpoint just returns a file object.
-        raise NotImplementedError()
+        return UserFile(path, connection=self)
+
+    def upload_file(self, source: Union[Path, str]) -> UserFile:
+        """
+        Uploads a file to the user workspace and stores it with the filename of the source given.
+        If a file with the name exists on the user workspace it will be replaced.
```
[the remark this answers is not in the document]
m-mohr: Yeah, this would be better suited in an API issue.

openeo/rest/connection.py
```diff
@@ -1092,20 +1094,29 @@ def list_files(self):
...
-        :return: file object.
+        :return: UserFile object.
         """
-        # No endpoint just returns a file object.
-        raise NotImplementedError()
+        return UserFile(path, connection=self)
+
+    def upload_file(self, source: Union[Path, str]) -> UserFile:
+        """
+        Uploads a file to the user workspace and stores it with the filename of the source given.
```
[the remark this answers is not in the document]
m-mohr: Solved, can now also be a path.

openeo/rest/userfile.py
m-mohr: Done

m-mohr:
Updated according to the review, seems to work fine:

image

soxofaan:
side-note: we started using pre-commit for automated code cleanups in commits and PRs, you might want to enable that too. More info: https://open-eo.github.io/openeo-python-client/development.html#precommit

openeo/rest/userfile.py
```diff
@@ -0,0 +1,71 @@
+import typing
+from typing import Any, Dict, Union
+from pathlib import Path
```
soxofaan, line 3:
```suggestion
from pathlib import Path, PurePosixPath
```

see lower

openeo/rest/userfile.py
```diff
@@ -0,0 +1,71 @@
...
+if typing.TYPE_CHECKING:
+    # Imports for type checking only (circular import issue at runtime).
+    from openeo.rest.connection import Connection
+
+
+class UserFile:
+    """Represents a file in the user-workspace of openeo."""
+
+    def __init__(self, path: str, connection: 'Connection', metadata: Dict[str, Any] = None):
+        self.path = Path(path)
```
soxofaan, line 15: based on you usage example screenshot (which shows this as `WindowsPath`), I now think this should be
```suggestion
        self.path = PurePosixPath(path)
```

because it is a non-concrete, posix-style path on the back-end

soxofaan: I'm not sure however how well this plays with path construction operations elsewehere in this PR.

openeo/rest/userfile.py
```diff
@@ -0,0 +1,71 @@
...
+            target = target / self.path.name
+        ensure_dir(target.parent)
+        
+        with target.open(mode="wb") as f:
+            for chunk in response.iter_content(chunk_size=None):
+                f.write(chunk)
+
+        return target
+    
+    def upload(self, source: Union[Path, str], target = None):
```
soxofaan, line 47: argument target is unused apparently

m-mohr: Oops, a left over.... fixed.

openeo/rest/userfile.py
m-mohr: Seems to work. I've made it so that PurePosixPath is only used for the back-end paths, but the download() methods still returns a path depending on the OS as it points to a locale file. But maybe also do a sanity check...

m-mohr:
@soxofaan Fixed the remaining comments and applied the pre-commit stuff...

soxofaan:
FYI Pushed a commit with some basic test coverage and additional finetuning

soxofaan:
Also changed location of upload logic: I think it makes more sense on Connection than on UserFile because upload creates a new UserFile. That's weird with UserFile method because it's not clear that it invalidates the original UserFile  instance.

I would even consider removing the UserFile.upload method

openeo/rest/connection.py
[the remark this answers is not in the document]
soxofaan: - https://github.com/Open-EO/openeo-api/issues/481

soxofaan:
I rebased this branch (to address wrongly merging of CHANGELOG.md) and merged in d16417c98290c98ffea0efc4b463374ed9ea6b14

thanks for pushing this!

m-mohr:
Thanks for completing this.

Honestly, I'm not a big fan of the rebase because it removes the affiliation to the original author completely and now it looks like you are the only author of this.

Issue #376, opened 2023-02-16
Simplify documentation about dev installation on Windows


JohanKJSchreurs:
Now that we have a conda package for the Python client we can probably simplify the developer installation instructions for the Python client as well.

The dev install on Windows would be a bit complicated before, partly because we didn't have the option to install a conda package.

PR #381, opened 2023-02-27
Issue #376 Shorten and simplify documentation about dev install on Wi…

JohanKJSchreurs:
# Issue #376 Shorten and simplify documentation about dev install on Windows

Link to issue: https://github.com/Open-EO/openeo-python-client/issues/376

### DIFF
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -270,128 +270,26 @@ Normally you can install the client the same way on Windows as on Linux, like so
     pip install -e .[dev]
 
 This should be working with the most recent code, however, in the past we sometimes had issues with a development installation.
-Should you experience problems, then this section describes some ways to solve them.
 
-Known issue
------------
-
-The specific problem we experienced on Windows is that the geopandas Python package depends on a few libraries that are a bit trickier to install.
-They need some compiled C/C++ code and unfortunately these libraries do not provide officially supported python wheels for Windows.
-
-Root Cause
-~~~~~~~~~~
-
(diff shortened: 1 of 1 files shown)


soxofaan:
looks good that the windows dependency hell stuff can be dropped, but what remains has little added value in the context of a "development" install (`pip install -e ...`)

soxofaan:
I had a quick look at `conda install`, and there seems to be a `--only-deps` option, so that might be a handy solution to do `conda install` for deps only and `pip install -e` for openeo itself

JohanKJSchreurs:
> I had a quick look at `conda install`, and there seems to be a `--only-deps` option, so that might be a handy solution to do `conda install` for deps only and `pip install -e` for openeo itself

Indeed, what was described would not have worked for development. 

I updated the instructions and tested it out.  The `--only-deps` option was indeed what we needed. After that step, installing the code in editable mode works correctly.

docs/development.rst
```diff
@@ -270,128 +270,39 @@ Normally you can install the client the same way on Windows as on Linux, like so
...
+There is a also shorter version of the above, if you prefer.
+You can create the environment and install openeo's dependencies in one go:
 
-.. code-block::
+.. code-block:: console
 
-    # In your activate virtualenv
-    # install the wheels:
-    pip install <path to GDAL whl file> <path to fiona whl file>
+    conda create -n openeopyclient --only-deps -c conda-forge openeo
```
soxofaan, line 304: I would not provide this short version in the docs, doing `conda create` with `--only-deps` might get ugly if users also add other packages to the `conda create`  line

JohanKJSchreurs: Good point. I removed that paragraph.

PR #382, opened 2023-02-27
added pattern for list_collections

clausmichele:
@soxofaan I was considering that it could be useful to have the ability to filter the collections directly by providing a keyword/pattern when calling list_collections(). What do you think?

### DIFF
--- a/openeo/rest/connection.py
+++ b/openeo/rest/connection.py
@@ -640,9 +640,11 @@ class Connection(RestApiConnection):
     def user_jobs(self) -> dict:
         return self.list_jobs()
 
-    def list_collections(self) -> List[dict]:
+    def list_collections(self, pattern: Optional[str] = None) -> List[dict]:
         """
         List basic metadata of all collections provided by the back-end.
+
+        :param pattern: a pattern that matches the collections of interest, e.g. "sentinel"
 
         .. caution::
 
@@ -653,6 +655,10 @@ class Connection(RestApiConnection):
         :return: list of dictionaries with basic collection metadata.
         """
         data = self.get('/collections', expected_status=200).json()["collections"]
+
+        if pattern is not None:
+            data = [collection for collection in data if 'id' in collection and pattern in collection['id']]
+
         return VisualList("collections", data=data)
 
     def list_collection_ids(self) -> List[str]:


soxofaan:
I'm a bit hesitant to this, as a user might want more flexibility than what the current PR allows: e.g. filter strictly on prefix/suffix, exclude instead of include, use a regex instead of single string, check multiple patterns at same time, ... 
so this might grow in an overkill solution at this place.

What might make more sense is to make `list_collections` return some kind of new `CollectionList` object that provides a `.filter` method that allows `.filter(include="sentinel2")`, `.filter(include=["sentinel2", "s2"])`, `.filter(exclude="experimental", case_sensitive=False)`, ...

Also, a user might want to filter on other fields than just `id`, so that is also something to consider here

clausmichele:
Indeed, this PR was just an idea that we can discuss about and that could evolve in different ways. I do consider this metadata filtering as something to tackle as possible new developments.

soxofaan:
when generalized in some way, this feature could also be useful in the context of:
- `list_services`
- `list_processes`
- `list_jobs`
- `list_user_defined_processes`
- batch job logs and secondary service logs

m-mohr:
fyi: In the Jupyter interface there's already a search control for the Vue components :-)

clausmichele:
> fyi: In the Jupyter interface there's already a search control for the Vue components :-)

I know, but sometimes it would be good to avoid user interactions and show directly what we would like to! So, if an example is focused on Sentinel-2, instead of writing "please type Sentinel-2 to filter the collections", we can just present them the available ones given a keyword.

m-mohr:
Sentinel 2 is a good example for an implementation, because it should likely match "Sentinel2", "Sentinel 2", "Sentinel-2" and "Sentinel_2" and "Sentinel 2A". And you may want to match it in ID, title and maybe even more (keywords, description?). Thus the code must do some normalization, which for example leads to quite a bit of code in the Vue components: https://github.com/Open-EO/openeo-vue-components/blob/master/utils.js#L434
The code can be an inspiration how to do it here as well. It should likely not be pure 1:1 "string matching" because otherwise you likely don't get the results you are looking for.

Issue #385, opened 2023-03-10
Improve UDF usage for  vector data cube processing


soxofaan:
While working on https://github.com/Open-EO/openeo-geopyspark-driver/issues/367, I noticed UDF based processing of a vector data cube is not on the same level as the efforts done in #312

Issue #386, opened 2023-03-10
`DataCube.aggregate_spatial` should return a vector cube


soxofaan:
`DataCube.aggregate_spatial` currently returns a (raster) `DataCube`:

https://github.com/Open-EO/openeo-python-client/blob/c590db2f796fa92466e4bda08e9938d8453069ff/openeo/rest/datacube.py#L962-L997

but it should actually return a vector cube.

Changing this should be easy, but I'm unsure if this would cause some trouble down the line in usecases that depend on the current implementation (of returning a DataCube object)

jdries:
I would change it. The only use case I can think of is running a udf on the result of aggregate_spatial, so if our new vector cube has the same methods for that, we should be fine?

soxofaan:
I indeed started with a PR (#389). I'm already touching more code than I anticipated, and I still have failing tests

soxofaan:
done

> The only use case I can think of is running a udf on the result of aggregate_spatial, so if our new vector cube has the same methods for that, we should be fine?

Indeed. That's actually even being improved in #385 / #388

PR #388, opened 2023-03-10
Improve support for running UDF directly on vector cube

soxofaan:
addresses #385


soxofaan:
depends on #386 (in sense that that one is fixed first preferably)

Issue #393, opened 2023-03-15
Some unit tests of python client are failing on Windows


JohanKJSchreurs:
### Some unit tests are failing on Windows

See pytest report below:

```
=========================================================================================== short test summary info ============================================================================================
SKIPPED [2] tests\udf\test_xarraydatacube.py:255: _save_DataArray_to_NetCDF introduces band names if they don't exist, which fails equality test
SKIPPED [2] tests\udf\test_xarraydatacube.py:265: _save_DataArray_to_NetCDF introduces band names if they don't exist, which fails equality test
SKIPPED [2] tests\udf\test_xarraydatacube.py:275: _save_DataArray_to_NetCDF introduces band dim if it doesn't exist, which fails equality test
FAILED tests/test_config.py::test_get_config_verbose[print-False-True] - assert None
FAILED tests/test_config.py::test_get_config_verbose[print-True-True] - assert None
FAILED tests/test_config.py::test_get_config_verbose[auto-False-False] - assert None
FAILED tests/test_config.py::test_get_config_verbose[auto-True-True] - assert None
FAILED tests/test_config.py::test_get_config_verbose[off-False-False] - assert None
FAILED tests/test_config.py::test_get_config_verbose[off-True-False] - assert None
FAILED tests/rest/test_userfile.py::TestUserFile::test_upload - AssertionError: assert b'hello world\r\n' == b'hello world\n'
FAILED tests/rest/test_userfile.py::TestUserFile::test_connection_upload[None-to-upload.txt] - AssertionError: assert b'hello world\r\n' == b'hello world\n'
FAILED tests/rest/test_userfile.py::TestUserFile::test_connection_upload[foo.txt-foo.txt] - AssertionError: assert b'hello world\r\n' == b'hello world\n'
FAILED tests/rest/test_userfile.py::TestUserFile::test_connection_upload[data/foo.txt-data/foo.txt] - AssertionError: assert b'hello world\r\n' == b'hello world\n'
========================================================================== 10 failed, 1203 passed, 6 skipped, 146 warnings in 39.74s ===========================================================================

```` 

### Possible causes

At least for a few of the tests, it looks like the cause is that the assert compares strings with newlines but the EOL characters are different on Windows,  "\r\n" instead of "\n"

The others seem to be related to paths to the config files, which will be a bit different on Windows. Could be an incorrect escape sequence in the test perhaps.

soxofaan:
[...]

Issue #387, opened 2023-03-10
openeo.connect error instead of warning in newer version


theroggy:
In newer versions of the client the warning has been changed to an error, so I'm not able to upgrade anymore.

I'm on a windows 10 system, and using this backend and authentication:
```
openeo.connect("https://openeo.vito.be").authenticate_oidc(provider_id="egi")
```
In older versions this gives this warning, in newer version an error:
```
File C:\Users\PIEROG\AppData\Roaming\openeo-python-client\refresh-tokens.json could be readable by others: mode 666 (expected: 600)
```

I tried all sorts of things by changing the access rights but I'm not able to get the mode to 600... the file always stays writable for "groups".
I also tried using setting it to 600 using the oschmod package, but this doesn't help either.

I'm working from a corporate network... maybe this is the problem?

Is there a possibility to bypass this check or another solution?

JohanKJSchreurs:
> In newer versions of the client the warning has been changed to an error, so I'm not able to upgrade anymore.
> 
> I'm on a windows 10 system, and using this backend and authentication:
> 
> ```
> openeo.connect("https://openeo.vito.be").authenticate_oidc(provider_id="egi")
> ```
> 
> In older versions this gives this warning, in newer version an error:
> 
> ```
> File C:\Users\PIEROG\AppData\Roaming\openeo-python-client\refresh-tokens.json could be readable by others: mode 666 (expected: 600)
> ```
> 
> I tried all sorts of things by changing the access rights but I'm not able to get the mode to 600... the file always stays writable for "groups". I also tried using setting it to 600 using the oschmod package, but this doesn't help either.
> 
> I'm working from a corporate network... maybe this is the problem?
> 
> Is there a possibility to bypass this check or another solution?

Hi,
Thanks for letting us know about this error.

We tried to find a workaround that would help you get unstuck while we fix the issue in the software.
Unfortunately, we didn't find a quick workaround, so instead we are working on getting a fix out sooner.

Basically, we are going back to the previous behavior: logging a message instead of raising an error when there is a potential risk that the file is readable to others.

#### Temporary solution:

In the meantime, I think your best option is to stay on the older version for a little longer, until the version with the fix is released. Version 0.13 should work, that is the last version before this change was introduced.


#### Regarding the workarounds you tried and what you observed:

Indeed, it seems likely that some security policy on the corporate network would be what is preventing the Python client from setting stricter file access permissions on refresh-tokens.json.

[...]

PR #394, opened 2023-03-15
Issue #393 Fixed unit tests that failed on Windows

JohanKJSchreurs:
Fixes issue https://github.com/Open-EO/openeo-python-client/issues/393

Also helps solving issue https://github.com/Open-EO/openeo-python-client/issues/387 though #387 does not strictly depend on it.
I just prefer to a have 100% passing test on the Windows side so I know failing tests are due to new or modified code.


soxofaan:
I already cherry-picked the userfile fix on a related PR #396

openeo/config.py
```diff
@@ -180,7 +180,16 @@ def get_config() -> ClientConfig:
     global _global_config
     if _global_config is None:
         _global_config = ConfigLoader.load()
-        message = f"Loaded openEO client config from sources {_global_config.sources}"
+
+        # In this case we avoid f-strings and use a more complicated expression,
+        # because on Windows we already have '\\' in the paths and then the f-string
+        # will escape those a *second* time, so we end up with '\\\\' .
+        # That is unreadable, and also a lot harder to deal with in unit tests.
```
soxofaan, line 187: I think this explanation is a bit misleading as escaping and double escaping issue is because of implicit `repr` in the pytest reporting, it's not related to f-strings or especially relevant in this non-test code path.

e.g. in #396  I'm trying a variant that still uses an f-string to keep things a bit more compact

CHANGELOG.md
```diff
@@ -42,6 +42,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fixed a few unit tests that were failing on Windows due to different EOL character and path separators ([#393](https://github.com/Open-EO/openeo-python-client/issues/393))
```
soxofaan, line 45: a changelog entry is not that necessary here,  unit test fixes  are generally not that valuable for the end user

soxofaan:
I cherry picked the fixes on #396, which confirms that the tests now pass on windows, so I merged #396 in master already, which practically closes this PR #394

PR #399, opened 2023-03-16
Issue #387 openeo.connect, reinstate old behavior: no PermissionError on Windows

JohanKJSchreurs:
## Issue #387 Quick fix, reinstate old behavior: no PermissionError on Windows for PrivateJsonFile, log message instead

Fixes https://github.com/Open-EO/openeo-python-client/issues/387

Reinstated the old behavior of authenticate_oidc on Windows: 
When PrivateJsonFile may be readable by others, log message instead of raising a PermissionError.

There is a follow-up issue here #400, if we still want to find a more secure solution for Windows.
To be decided.

Also, for now I have left in the code that uses oschmod because:

1) when it does work, then it still adds a bit more security;
2) we would like to get a solution out quickly, but removing oschmod cleanly takes a bit more work and care.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,7 +42,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- Reinstated old behavior of authenticate_oidc on Windows: when PrivateJsonFile may be readable by others, log message instead of raising PermissionError ([387](https://github.com/Open-EO/openeo-python-client/issues/387))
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -42,6 +42,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Reinstated old behavior of authenticate_oidc on Windows: when PrivateJsonFile may be readable by others, log message instead of raising PermissionError ([387](https://github.com/Open-EO/openeo-python-client/issues/387))
```
soxofaan, line 45:
```suggestion
- Reinstated old behavior of authentication related user files (e.g. refresh token store) on Windows: when `PrivateJsonFile` may be readable by others, just log a message instead of raising `PermissionError` ([387](https://github.com/Open-EO/openeo-python-client/issues/387))
```

soxofaan (approved):

Issue #401, opened 2023-03-24
execute_batch silently adds save_result even if there is already one


soxofaan:
```python
cube = cube.save_result(format="NetCDF")
cube.execute_batch()
```

the last `execute_batch()` call will silently add a `save_result("GTiff")` causing to post a process graph with two `save_result` nodes (in series), which is not intended by the user and not valid either (https://github.com/Open-EO/openeo-processes/issues/334)

soxofaan:
addressing #402  will probably simplify and help with the unification of this feature

soxofaan:
I think this issue has already been resolved some time ago, e.g. with fa3a4d249432be595e4f04c0e4bd0c0ae113dcee I think

PR #403, opened 2023-03-28
Improve automatic adding of `save_result`

soxofaan:
Issue #401


soxofaan:
This PR started to drag in too much changes/tweaks

starting over at #415

PR #408, opened 2023-04-03
Pass log_level to backend when getting BatchJob logs.

JohanKJSchreurs:
Part of implementation for Open-EO/openeo-python-driver#170

This fix has to be done across three code bases with their respective PRs

The main PR is: https://github.com/Open-EO/openeo-python-driver/pull/174

The code bases are:

- https://github.com/Open-EO/openeo-geopyspark-driver : Where logs are actually retrieved from Elasticsearch
- https://github.com/Open-EO/openeo-python-driver : needs to link through log_level parameter to backend implementation
- https://github.com/Open-EO/openeo-python-client : needs to be use log_level when retrieving logs via the openeo-python-driver

### DIFF
--- a/openeo/rest/job.py
+++ b/openeo/rest/job.py
@@ -177,21 +177,16 @@ class BatchJob:
             url, params={"offset": offset, "log_level": log_level}, expected_status=200
         ).json()["logs"]
 
-        #
-        # TODO: we should still support client-side log_level filtering when the backend
-        #   itself doesn't support filtering. What is the best way to check that, the API version?
-        #   https://github.com/Open-EO/openeo-python-driver/issues/170
-        #
-
         # Only filter logs when specified.
-        if self.connection.api_version.at_most("1.1.0"):
-            if log_level is not None:
-                log_level = normalize_log_level(log_level)
-                logs = (
-                    log
-                    for log in logs
-                    if normalize_log_level(log.get("level")) >= log_level
-                )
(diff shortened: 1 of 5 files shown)


openeo/rest/job.py
```diff
@@ -173,9 +173,13 @@ def logs(
         :return: A list containing the log entries for the batch job.
         """
         url = f"/jobs/{self.job_id}/logs"
-        logs = self.connection.get(url, params={'offset': offset}, expected_status=200).json()["logs"]
+        logs = self.connection.get(
+            url, params={"offset": offset, "log_level": log_level}, expected_status=200
```
soxofaan, line 177: openEO API only defines log levels as strings, so `log_level` (which at this point can still be an int) in the `params` should be normalized to one of: "error" "warning" "info" "debug"

JohanKJSchreurs: Added a utility function `log_level_name` that converts log_level to a lowercase string, one of the four supported log levels names:
https://github.com/Open-EO/openeo-python-client/pull/408/commits/73d4ad8ac27c59254d165aa234a01a4e7b45197f

soxofaan:
merged in b681ada5bf321af9feceb5fa05ab84e3ed1ed017  
(did some final fixup because request parameter will be `level` instead of `log_level`, see https://github.com/Open-EO/openeo-api/pull/488)

PR #409, opened 2023-04-03
[localprocessing] Fixed required packages versions

clausmichele:
Fixed required versions of openeo_pg_parser_networkx and openeo_processes_dask to the latest tested versions, to avoid incompatibilities and functionalities breaks due to updates.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -39,8 +39,8 @@ docs_require = [
 localprocessing_require = [
     "rioxarray",
     "pyproj",
-    "openeo_pg_parser_networkx>=2023.1.2",
-    "openeo_processes_dask>=2023.3.0",
+    "openeo_pg_parser_networkx==2023.3.1",
+    "openeo_processes_dask[implementations]==2023.3.2",
 ]
 
 


setup.py
```diff
@@ -37,10 +37,10 @@
 ]
 
 localprocessing_require = [
-    "rioxarray",
+    "rioxarray>=0.13.0",
     "pyproj",
-    "openeo_pg_parser_networkx>=2023.1.2",
-    "openeo_processes_dask>=2023.3.0",
+    "openeo_pg_parser_networkx==2023.3.1",
+    "openeo_processes_dask[implementations]==2023.3.2",
```
soxofaan, line 43: Generally it's bad style to pin versions down to an exact version in setup.py.
It's common practice to use `==`  in requirements.txt, but setup.py should be more liberal and typically just contain  `>=` constraints.

Is there a particular reason to pin down the versions like that?

soxofaan:
FYI: I already cherrypicked the rioxarray constraint on master

setup.py
clausmichele: I followed the suggestion of @LukeWeidenwalker, since he said that the next versions could introduce braking changes.

clausmichele:
> 

Ok, please note that rioxarray 0.13.0 is available only for Python 3.8+

setup.py
clausmichele: @soxofaan I removed the == in setup.py and added a requirements.txt file just with those two packages. Makes sense or do we need to include more in that file?

PR #411, opened 2023-04-03
[localprocessing] processing.py updates

clausmichele:
Adapted the syntax to comply with the new versions of the dependencies.


soxofaan:
is this related to the version bumps of #409?

soxofaan:
I probably didn't make myself clear when I proposed to split up #407 :  
these dependency bumps should be part of PR #411 and not be done as a separate PR (#409 in this case)

openeo/local/processing.py
```diff
@@ -6,22 +6,31 @@
...
+processes_from_module = [
     func
     for _, func in inspect.getmembers(
         importlib.import_module("openeo_processes_dask.process_implementations"),
         inspect.isfunction,
     )
 ]
 
-PROCESS_REGISTRY = ProcessRegistry(wrap_funcs=[openeo_process])
+specs_module = importlib.import_module("openeo_processes_dask.specs")
```
soxofaan, line 23: Any particular reason you use `importlib.import_module("openeo_processes_dask.specs")`
instead of just `import openeo_processes_dask.specs` and just work with `openeo_processes_dask.specs` as an object?

soxofaan: same actually with `importlib.import_module("openeo_processes_dask.process_implementations"),` above

openeo/local/processing.py
```diff
@@ -34,7 +43,15 @@ def load_local_collection(*args, **kwargs):
...
         data = xr.open_dataset(kwargs['id'],chunks={},decode_coords='all').to_array(dim='bands') # Add decode_coords='all' if the crs as a band gives some issues
     elif '.tiff' in collection.suffixes or '.tif' in collection.suffixes:
-        data = rioxarray.open_rasterio(kwargs['id']).rename({'band':'bands'})
+        data = rioxarray.open_rasterio(kwargs['id'],chunks={},band_as_variable=True)
+        for d in data.data_vars:
+            data_attrs_lowercase = [x.lower() for x in data[d].attrs]
+            data_attrs_original  = [x for x in data[d].attrs]
+            data_attrs = dict(zip(data_attrs_lowercase,data_attrs_original))
+            if 'description' in data_attrs_lowercase:
+                data = data.rename({d:data[d].attrs[data_attrs['description']]})
```
soxofaan, line 52: That looks quite convoluted. If I understand correctly, you just want to find a "description" attr (if any) in a case insensitive manner. 
I think this should do it:
```python
descriptions = [v for k, v in data[d].attrs if k.lower() == "description"]
if descriptions:
    data = data.rename({d: data[d].attrs[descriptions[0]])
```

openeo/local/processing.py
clausmichele: @LukeWeidenwalker I need your help here, this code comes from: https://github.com/Open-EO/openeo-processes-dask/blob/9afc89aca7f1090810572e4d8c42c0a450db02d5/tests/conftest.py#L57

openeo/local/processing.py
clausmichele: thanks, I'll try it soon!

clausmichele:
> 

indeed, I will merge the two PRs

PondiB (approved):
👍🏾

requirements.txt
```diff
@@ -0,0 +1,3 @@
+# Tested versions of specific packages used in openeo[localprocessing]
+openeo-pg-parser-networkx==2023.3.1
+openeo-processes-dask==2023.3.2
```
soxofaan, line 3: This looks, with a filename like `requirements.txt`, the official list of dependencies for all users.
Also, some tools could automatically pick this up wrongly.
I would at least call it `requirements-localprocessing.txt` and put a more clear comment at the top (e.g. also mentioning this is an exprimental, optional feature)

clausmichele: Done! Thanks for the clarification, I was in fact unsure about the naming of the file.

openeo/local/processing.py
LukeWeidenwalker: Ah, this was copied over from the process registry initialisation code in the executor where I did something like `importlib.import_module(f"{module_source}.specs")` to grab processes from different sources. Afaics this can be safely removed here as @soxofaan suggests!

openeo/local/processing.py
```diff
@@ -6,22 +6,31 @@
...
+    func.__name__: getattr(specs_module, func.__name__)
+    for func in processes_from_module
+}
 
-for func in standard_processes:
-    PROCESS_REGISTRY[func.__name__] = func
+for func in processes_from_module:
+    PROCESS_REGISTRY[func.__name__] = Process(
+        spec=specs[func.__name__], implementation=func
+    )
```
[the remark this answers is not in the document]
clausmichele: Fixed with the latest commit

openeo/local/processing.py
clausmichele: Fixed with the latest commit

openeo/local/processing.py
```diff
@@ -6,22 +6,36 @@
...
+    specs = {
+        func.__name__: getattr(openeo_processes_dask.specs, func.__name__)
+        for func in processes_from_module
+    }
+
+    for func in processes_from_module:
+        process_registry[func.__name__] = Process(
+            spec=specs[func.__name__], implementation=func
+        )
+    return process_registry
```
soxofaan, line 35: Unless I'm missing something, I think this can be done more compactly with:

```python
for name, func in inspect.getmembers(
    openeo_processes_dask.process_implementations,
    inspect.isfunction,
):
    process_registry[name] = Process(
        spec=getattr(openeo_processes.specs, name),
        implementation=func
    )
```

Also, wouldn't it be better that openeo_processes_dask provide helpers to list implementations/spec, instead of having to use ugly `inspect.getmembers` and `getattr` constructs?

openeo/local/processing.py
```diff
@@ -34,7 +48,13 @@ def load_local_collection(*args, **kwargs):
...
+            descriptions = [v for k, v in data[d].attrs.items() if k.lower() == "description"]
+            if descriptions:
+                data = data.rename({d: descriptions[0]})
+        data = data.to_array(dim='bands')
     return data
 
-PROCESS_REGISTRY["load_collection"] = load_local_collection
\ No newline at end of file
+from openeo_processes_dask.specs import load_collection as load_collection_spec
+PROCESS_REGISTRY["load_collection"] = Process(spec=load_collection_spec, implementation=load_local_collection)
```
soxofaan, line 60: I don't think it is necessary to do the `from openeo_processes_dask.specs import load_collection as load_collection_spec` import, you can just define load_collection with

```python
PROCESS_REGISTRY["load_collection"] = Process(
    spec=openeo_processes_dask.specs.load_collection, 
    implementation=load_local_collection,
)
```

soxofaan:
merged in master

openeo/local/processing.py
LukeWeidenwalker: Ah apologies, I missed this tag - yes, agree with Stefaan, this is something openeo-processes-dask could do better - have created an issue! 
https://github.com/Open-EO/openeo-processes-dask/issues/92

PR #430, opened 2023-05-25
Job manager: trigger on_job_done for finished job from all active states

StijnCaerts:
Right now, the `on_job_done()` method will only be triggered when the cached state is `running` and the job status is `finished`. But it might be possible that the job manager was stopped and resumed later and jobs in another _active_ state (`created`, `queued`) finished in the mean time.
For example, this means that jobs going from `queued` to `finished` will not trigger the `on_job_done()` method.

This PR changes the behaviour of the job manager to trigger the `on_job_done()` method on the transition from **any** active state to the `finished` state.

### DIFF
--- a/openeo/extra/job_management.py
+++ b/openeo/extra/job_management.py
@@ -419,10 +419,7 @@ class MultiBackendJobManager:
                 _log.info(
                     f"Status of job {job_id!r} (on backend {backend_name}) is {job_metadata['status']!r}"
                 )
-                if (
-                    df.loc[i, "status"] == "running"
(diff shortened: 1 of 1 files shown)


StijnCaerts:
It shouldn't make a difference, as the active jobs are filtered before iterating. Jobs already in the finished state will not be considered.
https://github.com/Open-EO/openeo-python-client/blob/a1a2ec3f36eb755f8a10cde850c830b8cd7df86c/openeo/extra/job_management.py#L404-L412

PR #434, opened 2023-06-01
localprocessing updates

clausmichele:
This PR contains:

- added the implementation of resample_cube_spatial based on xArray, since the one in openeo-processes-dask based on opendatacube relies on metadata provided only when the data comes from ODC
- added two example notebooks
- updated the minimum required library versions


openeo/local/processing.py
```diff
@@ -59,4 +68,74 @@ def load_local_collection(*args, **kwargs):
 PROCESS_REGISTRY["load_collection"] = Process(
     spec=openeo_processes_dask.specs.load_collection, 
     implementation=load_local_collection,
+)
+
+def resample_cube_spatial_rioxarray(*args, **kwargs):
+    pretty_args = {k: type(v) for k, v in kwargs.items()}
+    _log.debug(f"Running process resample_cube_spatial")
+    _log.debug(f"kwargs: {pretty_args}")
+    _log.debug("-" * 80)
```
soxofaan, line 77: don't log like you would `print` (e.g. multiple lines, `"-" * 80`) , keep things that belong together in same log call

openeo/local/processing.py
```diff
@@ -59,4 +68,74 @@ def load_local_collection(*args, **kwargs):
...
+        "min": 9,
+        "med": 10,
+        "q1": 11,
+        "q3": 12,
+        "sum": 13,
+        "rms": 14
+    }
+    data = kwargs['data']
+    target = kwargs['target']
+    method = kwargs['method']
```
soxofaan, line 98: if `data`, `target` and `method` are required: why not defining these arguments directly instead of using `kwargs`? It makes things more clear, self-documenting, you can add type annotations, the error message are nicer, ...

openeo/local/processing.py
```diff
@@ -59,4 +68,74 @@ def load_local_collection(*args, **kwargs):
...
+    # The target doesn't need to have the same dimensions as the data
+    input_data_dims = (
+        data.openeo.band_dims + data.openeo.temporal_dims + data.openeo.spatial_dims + data.openeo.other_dims
+    )
+
+    if method not in methods_list.keys():
```
soxofaan, line 104: FYI: with dicts you can just do 
```suggestion
    if method not in methods_list:
```

openeo/local/processing.py
```diff
@@ -59,4 +68,74 @@ def load_local_collection(*args, **kwargs):
...
     implementation=load_local_collection,
+)
+
+def resample_cube_spatial_rioxarray(*args, **kwargs):
+    pretty_args = {k: type(v) for k, v in kwargs.items()}
+    _log.debug(f"Running process resample_cube_spatial")
+    _log.debug(f"kwargs: {pretty_args}")
+    _log.debug("-" * 80)
+
+    methods_list = {
```
soxofaan, line 79: don't call a dictionary a list, that's quite confusing

also: where do these int values come from, is that ad-hoc, or is that defined somewhere?

openeo/local/processing.py
```diff
@@ -59,4 +68,74 @@ def load_local_collection(*args, **kwargs):
...
+    data = kwargs['data']
+    target = kwargs['target']
+    method = kwargs['method']
+    # The target doesn't need to have the same dimensions as the data
+    input_data_dims = (
+        data.openeo.band_dims + data.openeo.temporal_dims + data.openeo.spatial_dims + data.openeo.other_dims
+    )
+
+    if method not in methods_list.keys():
+        raise Exception(
```
soxofaan, line 105: don't use generic `Exception`, In this case it makes more sense to use `ValueError`

examples/notebooks/Client_Side_Processing/client_side_processing_color_composites.ipynb
```diff
@@ -0,0 +1,443 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "7413f5bd-542e-4efa-a3e0-f338c254f725",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import os\n",
+    "if not os.path.exists('./openeo-localprocessing-data'):\n",
+    "    !git clone https://github.com/Open-EO/openeo-localprocessing-data.git"
```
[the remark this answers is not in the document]
clausmichele: git clone works fine if git is installed I guess! Anyway, I'm implementing `load_stac`, once ready we won't even need to clone the repo to get sample data.

openeo/local/processing.py
clausmichele: I was copying the style of the implemented processes in `openeo-processes-dask`. @LukeWeidenwalker any thought on this?

LukeWeidenwalker: Agree with @soxofaan, this is something that could be improved in openeo-processes-dask too!

examples/notebooks/Client_Side_Processing/client_side_processing_color_composites.ipynb
```diff
@@ -0,0 +1,443 @@
...
+   "source": [
+    "## Requirements"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "51a9e53a-481c-4ea5-a175-652d6398828d",
+   "metadata": {},
+   "source": [
+    "To use this functionality, you need `3.9<=Python<3.11`.\n",
```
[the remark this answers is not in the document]
clausmichele: Comes from here: https://github.com/Open-EO/openeo-processes-dask/blob/6563c1809fc4108581b78763a970b3bea1d5d935/pyproject.toml#L25

@LukeWeidenwalker how could I help to test more Python versions? It would be good to add 3.11 at least.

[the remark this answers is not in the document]
LukeWeidenwalker: When I initially set up the repo, I put 3.9, 3.10 and 3.11 in the CI matrix for github actions. 3.11 failed for a reason that I can't recall now, and since our backend runs on 3.9, I've just ignored it since. Would be in favor of becoming 3.11 compatible though! I'd start by including 3.11 in the matrix here: https://github.com/Open-EO/openeo-processes-dask/blob/main/.github/workflows/main.yml in a new PR and just seeing what breaks. Once the build passes, we can remove this constraint!

openeo/local/processing.py
clausmichele: They come from the rioxarray - rasterio docs:

https://corteva.github.io/rioxarray/stable/rioxarray.html#rioxarray.raster_array.RasterArray.reproject_match
https://rasterio.readthedocs.io/en/stable/api/rasterio.enums.html#id0

Should I include the link in the code as a comment?

openeo/local/processing.py
clausmichele: Done

openeo/local/processing.py
clausmichele: Done

openeo/local/processing.py
clausmichele: Done

openeo/local/processing.py
clausmichele: I will improve this once we agree in `openeo-processes-dask` how to improve the logs format

openeo/local/processing.py
clausmichele: I created this mapping since in openEO we have "near" instead of "nearest" and "cubicspline" instead of "cubic_spline"

soxofaan: then I would just translate these minor differences, but let `rasterio.enums.Resampling` do the heavy lifting, e.g.
```python
method = {
    "near": "nearest",
    "cubicspline": "cubic_spline"
}.get(method, method)
try:
    resampling = rasterio.enums.Resampling[method]
except KeyError:
    ValueError(
            f'Selected resampling method not available.... select one of {[r.name for r in rasterio.enums.Resampling]}" ...
```

clausmichele: Fine. However, if we raise that ValueError, we would list nearest and cubic_spline as allowed resampling methods, that if used in a process graph would make it invalid if there's some validation mechanism in place.

openeo/local/processing.py
LukeWeidenwalker: Proposal here: https://github.com/Open-EO/openeo-processes-dask/pull/128

clausmichele: Modified according to the new format implemented by @LukeWeidenwalker

openeo/local/processing.py
clausmichele: @soxofaan I think that this is the last point open before being able to merge. Can we agree on my latest version?

soxofaan:
merged 

(FYI did some post-merge cleanups in  fbf69e6 , including removal of unused `input_data_dims` which you might want to double check)

Issue #442, opened 2023-06-15
avoid "No cube:dimensions metadata" warning or errors from `datacube_from_process`


soxofaan:
`connection.datacube_from_process()` causes 
```
openeo-python-client/openeo/metadata.py:255: UserWarning: No cube:dimensions metadata
  complain("No cube:dimensions metadata")
```

which is not very helpful and can not be avoided because there is not really a way to provided expected metadata

`datacube_from_process` should not trigger that warning by default (or it should make it possible/easy to specify metadata)

soxofaan:
another use case:

```
cube = connection.datacube_from_process(...
cube.apply_dimension(dimension="t", ...
```
fails with 
```
ValueError: Invalid dimension 't'. Should be one of []
```

Now, the user is forced to explicitly add define the dimensions clientside in order to get the `apply_dimension`  working.

I think it would be better to skip the client-side dimension checks when there is no metadata, as the backend will complain anyway if something is wrong

soxofaan:
another related issue: 
- https://github.com/Open-EO/openeo-python-client/issues/471

soxofaan:
merged

PR #445, opened 2023-06-28
Add load_stac to localprocessing

clausmichele:
Added a copy paste version of load_stac into the LocalConnection object.
This will allow me to use it in this way:

```python
import openeo
local_conn = LocalConnection('')
url = "https://planetarycomputer.microsoft.com/api/stac/v1/collections/landsat-c2-l2"
spatial_extent = {"west": 11.259613, "east": 11.406212, "south": 46.461019, "north": 46.522237}
temporal_extent = ["2022-06-01","2022-06-15"]
bands = ["red"]
properties = {"eo:cloud_cover": dict(lt=50)}
s2_cube = local_conn.load_stac(url=url,
                    spatial_extent=spatial_extent,
                    temporal_extent=temporal_extent,
                    bands=bands,
                    properties=properties)
s2_cube.execute()
```

@soxofaan we would need to understand if we could reuse the code I'm developing here https://github.com/Open-EO/openeo-processes-dask/pull/127 to get the cube metadata, since from the url only is not so easy, especially for big STAC Collections.
If the metadata is not set correctly, the client unfortunately doesn't allow to proceed in many cases. By the way, is it possible to deactivate manually the checks on the metadata?

### DIFF
--- a/openeo/local/connection.py
+++ b/openeo/local/connection.py
@@ -4,7 +4,7 @@ from pathlib import Path
 import xarray as xr
 from typing import Dict, List, Tuple, Union, Callable, Optional, Any, Iterator
 
-from openeo.metadata import CollectionMetadata
+from openeo.metadata import CollectionMetadata, SpatialDimension, TemporalDimension, BandDimension, Band
 from openeo.internal.graph_building import PGNode, as_flat_graph
 from openeo.rest.datacube import DataCube
 from openeo.internal.jupyter import VisualDict, VisualList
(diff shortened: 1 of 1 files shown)


soxofaan:
> we would need to understand if we could reuse the code I'm developing here https://github.com/Open-EO/openeo-processes-dask/pull/127 to get the cube metadata

The local processing feature currently depends on `openeo_processes_dask[implementations]`, so it should be possible to reuse functionality from there.

> By the way, is it possible to deactivate manually the checks on the metadata?

yes, I'm considering something like that under #442

clausmichele:
@soxofaan I added the code that, once the `load_stac` PR in `openeo-processes-dask` will be merged, will load the metadata using that implementation. It adds some overhead in terms of time, but in the meanwhile validates also the `load_stac` call.

openeo/local/connection.py
```diff
@@ -84,7 +85,156 @@ def load_collection(
...
+        if temporal_extent:
+            arguments["temporal_extent"] = DataCube._get_temporal_extent(temporal_extent)
+        if bands:
+            arguments["bands"] = bands
+        if properties:
+            arguments["properties"] = properties
+        cube = self.datacube_from_process(process_id="load_stac", **arguments)
+        # detect actual metadata from URL
+        # run load_stac to get the datacube metadata
+        xarray_cube = cube.execute()
```
soxofaan, line 215: You "execute" the `load_stac` only for the metadata here.
But then later when doing the real processing, you will execute the `load_stac` again?

clausmichele: Yes, far from optimal. Do you have any ideas to avoid this?
A possibility would be to somehow remove the load_stac process from the graph and pass directly the xarray.DataArray previously loaded to openeo-processes-dask, but it would be a breaking change (there's the process graph parser in between) and not so straightforward to implement.

openeo/local/connection.py
```diff
@@ -84,7 +85,156 @@ def load_collection(
...
+        attrs = xarray_cube.attrs
+        attrs = {}
+        for at in attrs:
```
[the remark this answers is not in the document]
clausmichele: Solved

openeo/local/connection.py
```diff
@@ -84,7 +85,156 @@ def load_collection(
...
+                BandDimension(name=xarray_cube.openeo.band_dims[0],
+                              bands=[Band(x) for x in xarray_cube[xarray_cube.openeo.band_dims[0]].values])
+            ],
+        )
+        cube.metadata = metadata
+        return cube
+
+
+    def execute(self, process_graph: Union[dict, str, 
+                                          ]) -> xr.DataArray:
```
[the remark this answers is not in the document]
clausmichele: Done

openeo/local/connection.py
clausmichele: Anyway, consider that the execute() call will just interact with the STAC Catalog/Items without actually downloading any data, just the metadata

soxofaan: > Yes, far from optimal. Do you have any ideas to avoid this?

some options
- leave as is for now in this proof-of-concept phase and only tackle this later when it becomes a problem (performance, users complaining about duplicated data requests, ...)
- is there a way to do a variation of "load_stac" that only considers/extracts metadata?

Current implementation conflicts conceptually a bit with how the python client operates elsewhere: first build some kind of virtual process graph representation, and only "execute" it on explicit user request. Now there is an hidden "execute" by just doing `con.load_stac`. I just worried if this could cause troubles in long term

> consider that the execute() call will just interact with the STAC Catalog/Items without actually downloading any data, just the metadata

how can this `execute` know to only fetch metadata? At some point there will be another `execute` that needs the pixel data?

clausmichele: > 



> > Yes, far from optimal. Do you have any ideas to avoid this?
> 
> some options
> 
>     * leave as is for now in this proof-of-concept phase and only tackle this later when it becomes a problem (performance, users complaining about duplicated data requests, ...)
> 
>     * is there a way to do a variation of "load_stac" that only considers/extracts metadata?
> 
> 
> Current implementation conflicts conceptually a bit with how the python client operates elsewhere: first build some kind of virtual process graph representation, and only "execute" it on explicit user request. Now there is an hidden "execute" by just doing `con.load_stac`. I just worried if this could cause troubles in long term

I also don't like it much, but if we can't deactivate the metadata checks on the client side it's necessary to allow many operations that would be stopped due to missing metadata.

> 
> > consider that the execute() call will just interact with the STAC Catalog/Items without actually downloading any data, just the metadata
> 
> how can this `execute` know to only fetch metadata? At some point there will be another `execute` that needs the pixel data?

It returns an xarray.DataArray with dask chunks, only when you actually use the values of the data or call explicitly `.compute()` it will start loading data into memory. If you want we can have a meeting tomorrow and I'll explain you the details!

clausmichele: Not using .execute anymore, but calling directly `load_stac` for metadata.

clausmichele:
@soxofaan the changes that we discussed about have been implemented, you could now review and merge!

requirements-localprocessing.txt
```diff
@@ -1,5 +1,5 @@
 # Those requirements are for the experimental and optional client side processing feature
 # that can be installed via pip install openeo[localprocessing].
 # Please be aware that this functionality is not fully stable and may change quickly.
-openeo-pg-parser-networkx==2023.3.1
-openeo-processes-dask==2023.3.2
```
soxofaan, line 5: I think we should eliminate `requirements-localprocessing.txt` and handle everything through the `localprocessing` extra of `setup.py`,

docs/cookbook/localprocessing.rst
```diff
@@ -26,18 +26,68 @@ Installation
 Usage
 -----
 
-Local Collections
+Every openEO process graph relies on data, which was always provided by a cloud infrastructure (the openEO back-end) until now.
```
soxofaan, line 29:
```suggestion
Every openEO process graph relies on data which is typically provided by a cloud infrastructure (the openEO back-end).
```

docs/cookbook/localprocessing.rst
```diff
@@ -26,18 +26,68 @@ Installation
...
+The client-side processing adds the possibility to read and use local netCDFs, geoTIFFs and ZARR files and remote STAC Collections or Items for your experiments.
```
soxofaan, line 30:
```suggestion
The client-side processing adds the possibility to read and use local netCDFs, geoTIFFs, ZARR files, and remote STAC Collections or Items for your experiments.
```

docs/cookbook/localprocessing.rst
```diff
@@ -26,18 +26,68 @@ Installation
...
+
+    >>> from openeo.local import LocalConnection
+    >>> local_conn = LocalConnection("./")
+
+    >>> url = "https://earth-search.aws.element84.com/v1/collections/sentinel-2-l2a"
+    >>> spatial_extent = {"west": 11, "east": 12, "south": 46, "north": 47}
+    >>> temporal_extent = ["2019-01-01","2019-06-15"]
+    >>> bands = ["red"]
+    >>> properties = {"eo:cloud_cover": dict(lt=50)}
+    >>> s2_cube = local_conn.load_stac(url=url,
```
soxofaan, line 51: I'd use a 4 space indent here

soxofaan (commented):
some minor stuff

requirements-localprocessing.txt
soxofaan: the docs already instruct to do
```
pip install openeo[localprocessing]
```

anyway

openeo/local/connection.py
```diff
@@ -88,7 +91,161 @@ def load_collection(
...
+        cube.metadata = metadata
+        return cube
+
+    def execute(
+        self,
+        process_graph: Union[
+            dict,
+            str,
+        ],
+    ) -> xr.DataArray:
```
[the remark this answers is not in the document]
clausmichele: Ok, I think it was automatically edited by the pre-commit hooks.

clausmichele:
@soxofaan points from review addressed!

soxofaan:
merged in 8d1d947e8fb795480fb606dd8da1674c758c7b13

Issue #432, opened 2023-05-26
MultibackendJobManager: stop when finished


jdries:
when all jobs are processed, script should finish instead of remaining in infinite loop

soxofaan:
is that because status "error" is missing from 
https://github.com/Open-EO/openeo-python-client/blob/2b8914c4ff2394bffbb1ec25cb3f906076fa44a6/openeo/extra/job_management.py#L254-L261 ?

jdries:
seems likely

PR #452, opened 2023-07-28
Issue432 multibackendjobmanager stop when finished

JohanKJSchreurs:
Fixes: https://github.com/Open-EO/openeo-python-client/issues/432


soxofaan:
@JohanKJSchreurs this is a small fix, but I would add a changelog entry for it

Issue #259, opened 2021-11-25
Allow epsg-code to be provided as string?


m-mohr:
The specification asks to provide EPSG codes as numbers (e.g. 4326 instead of EPSG:4326). We see many code snippets that actually use the string variant, which strictly speaking should be WKT2.

So an idea coming up in recent discussions is that we use the clients to correct this common mistake.
So for subtype `epsg-code` a client could simply match against the pattern/regexp `/^EPSG:(\d{4,})$/i` and replace it with the first match converted to an int.

In pseudo-code something like:
```
if $subtype == 'epsg-code':
  $matches = $value.match(/^EPSG:(\d{4,})$/i)
  if $matches:
    $value = int($matches[1])
```

clausmichele:
The Equi7 requires the following string to be parsed for using it in load_collection
https://discuss.eodc.eu/t/get-specific-data-extent-using-equi7-coordinates/223
`"crs": "PROJCRS[\"Azimuthal_Equidistant\",BASEGEOGCRS[\"WGS 84\",DATUM[\"World Geodetic System 1984\",ELLIPSOID[\"WGS 84\",6378137,298.257223563,LENGTHUNIT[\"metre\",1]]],PRIMEM[\"Greenwich\",0,ANGLEUNIT[\"degree\",0.0174532925199433]],ID[\"EPSG\",4326]],CONVERSION[\"Modified Azimuthal Equidistant\",METHOD[\"Modified Azimuthal Equidistant\",ID[\"EPSG\",9832]],PARAMETER[\"Latitude of natural origin\",53,ANGLEUNIT[\"degree\",0.0174532925199433],ID[\"EPSG\",8801]],PARAMETER[\"Longitude of natural origin\",24,ANGLEUNIT[\"degree\",0.0174532925199433],ID[\"EPSG\",8802]],PARAMETER[\"False easting\",5837287.81977,LENGTHUNIT[\"metre\",1],ID[\"EPSG\",8806]],PARAMETER[\"False northing\",2121415.69617,LENGTHUNIT[\"metre\",1],ID[\"EPSG\",8807]]],CS[Cartesian,2],AXIS[\"easting\",east,ORDER[1],LENGTHUNIT[\"metre\",1,ID[\"EPSG\",9001]]],AXIS[\"northing\",north,ORDER[2],LENGTHUNIT[\"metre\",1,ID[\"EPSG\",9001]]]]"`

m-mohr:
[...]

PR #453, opened 2023-07-28
Issue259 allow epsg code as string

JohanKJSchreurs:
Implements https://github.com/Open-EO/openeo-python-client/issues/259


openeo/util.py
```diff
@@ -627,3 +630,74 @@ def get(self, fraction: float) -> str:
         width = self.width - len(self.left) - len(self.right)
         bar = self.bar * int(round(width * clip(fraction, min=0, max=1)))
         return f"{self.left}{bar:{self.fill}<{width}s}{self.right}"
+
+
+class EPSGCodeNotFound(Exception):
```
soxofaan, line 635: "not found" seems like a strange expressions for a CRS code, maybe make it "invalid EPSG code". Or just keep it to a vanilla `ValueError` with a good message?

JohanKJSchreurs: This was indeed not ideal, I was in fact already planning to remove that one.
Now these are ValueErrors. (And typerrors for anything that is not str, int or None)

jdries:
fine for me!

JohanKJSchreurs:
Have found out that we should also support PROJJSON.  Adding some test coverage for that as well.

See also:
https://github.com/Open-EO/openeo-processes/issues/58
https://github.com/Open-EO/openeo-processes/commit/2684dab980b02bad8fe7aa2d59049462ca382760

PR #458, opened 2023-08-09
Issue #259  Make pyproj an optional dependency when parsing CRS strin…

JohanKJSchreurs:
Issue #259 continued: Make pyproj an optional dependency when parsing CRS string / EPSG code

- Now the import of pyproj happens locally in the function that needs it.
- CRS values are only processed further with pyproj when they are neither type int nor a string that represents an integer.

### DIFF
--- a/openeo/util.py
+++ b/openeo/util.py
@@ -15,7 +15,6 @@ from typing import Any, Callable, Optional, Tuple, Union
 from urllib.parse import urljoin
 
 import requests
-import pyproj.crs
 import shapely.geometry.base
 from deprecated import deprecated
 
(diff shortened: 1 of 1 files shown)


soxofaan:
did some refactoring in the end in b2a0e00ef9f61c75a91167328ded9a09f6a43522  (tagged this ticket wrong apparently).
Biggest change is that the normalization function also can return a WKT2 CRS string (not only EPSG int), because that is also allowed in openEO API. This made it actually possible to simplify the function and rework the optional `pyproj` import stuff

soxofaan:
pff that pyproj stuff is highly version sensitive. It took more commits that I anticipated to get things right. I should have worked in a feature branch in hindsight :stuck_out_tongue:

Issue #421, opened 2023-05-08
feature request: allow specifying temporal extent with single string


soxofaan:
Temporal extents have to be specified as a list/tuple with two dates.
Often, when experimenting/exploring, you just need a single year or month, it would be handy if you can just pass `temporal_extent="2022"`  or `temporal_extent="2023-04"`, which would be converted automatically by the client to appropriate date ranges

soxofaan:
PR #461 was merged but I'd like to iterate a bit further on it:

I don't think that `string_to_temporal_extent` or even `get_temporal_extent` should be user-facing helpers (related: #465).  They should just be reusable helpers for actual user-facing methods like `Connection.load_collection`, `DataCube.filter_temporal`, ... As such I would not document this feature through the doc strings of these functions, but in the rst docs somewhere (e.g. https://open-eo.github.io/openeo-python-client/data_access.html seems like a good option), which could then be linked to from the docs of `Connection.load_collection`, `DataCube.filter_temporal`, ...

PR #461, opened 2023-08-11
Issue421 specifying temporal extent with single string

JohanKJSchreurs:
Implements: https://github.com/Open-EO/openeo-python-client/issues/421


openeo/util.py
```diff
@@ -281,9 +281,78 @@ def get_temporal_extent(*args,
...
+def string_to_temporal_extent(start_date: str) -> Tuple[dt.date, dt.date]:
+    """Convert a string that represents a year or a month, into a date range.
+
+    For example:
+    "2021" : means all data for 2021
+    "2022-08": means all data for the month of august in 2022.
+
+    If the day is included in the string then we leave it alone.
+    Note that `get_temporal_extent` already handles having one date string,
+    where the day is present in the string.
```
jdries, line 298: Can we also add proper parameter and return type documentation? So that this is rendered properly in api docs:
https://open-eo.github.io/openeo-python-client/api.html#module-openeo.util

jdries (approved):

PR #463, opened 2023-08-23
define a new udf signature to separate metadata from data transformation

jdries:
Separating metadata transformation allows process graph analysis to work better when UDF's perform more complex transformations.
One example was requested in this issue:
https://github.com/Open-EO/openeo-geopyspark-driver/issues/412

Applies for instance to UDF's that:
- change band labels
- change spatial resolution
- change crs of spatial dimension
- change timestamps on temporal dimension

I requested some reviews as this is an extension of the UDF API. It's a non-breaking change, UDF's that work will continue to work.
@m-mohr  I'll want to make this an agenda topic on one of the next dev meetings. Do we require PSC approval for changes to UDF API?

### DIFF
--- a/docs/udf.rst
+++ b/docs/udf.rst
@@ -277,6 +277,19 @@ instead of the original digital number range (thousands):
 .. image:: _static/images/udf/apply-rescaled-histogram.png
 
 
+UDF's that transform cube metadata
+==================================
+In some cases, a UDF can have impact on the metadata of a cube, but this can not always
+be easily inferred by process graph evaluation logic without running the actual
+(expensive) UDF code. This limits the possibilities to validate process graphs,
+or for instance make an estimate of the size of a datacube after applying a UDF.
+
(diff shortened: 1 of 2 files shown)


LukeWeidenwalker:
At EODC we do not support UDFs right now and it's currently not on our roadmap, so nothing to comment on from my side! (maybe @christophreimer has something to add)

m-mohr:
@jdries Added to the agenda for September, @clausmichele is leading it this month.

The PSC guideslines are not overly clear on this: "Anything that could cause backward compatibility issues in normative specifications (e.g., API or process specifications)." Is the UDF API a normative specification? :-) But I'd say it doesn't hurt to at least present the anticipated changes there. Next PSC meeting is in October.

christophreimer (approved):

soxofaan (commented):
Looks good to me.

PR #466, opened 2023-09-04
Add stac notebook

clausmichele:
Adding a new example for Client Side Processing, using data from different STAC Catalogues


soxofaan:
interesting

Is it possible to eliminate some of these bits of visual noise?
-  user warnings `UserWarning: The argument 'infer_datetime_format' is deprecated and will be removed in a future version. `
- `Deserialised process graph into nested structure Walking node root-b29...`

clausmichele:
@soxofaan I could do it easily erasing those parts in the notebook "text" file, otherwise I would have to set some filtering in the code itself. It depends if you would like to have a nicer notebook preview or get the same also when executing the code.

soxofaan:
I don't mean to filter out these warnings, but to avoid that they are triggered in the first place. 

e.g. isn't this triggered from  the localconnection code (or another openeo library) and can't we just fix that deprecation?
> stackstac\stackstac\prepare.py:369: UserWarning: The argument 'infer_datetime_format' is deprecated and will be removed in a future version. A strict version of it is now the default, see https://pandas.pydata.org/pdeps/0004-consistent-to-datetime-parsing.html. You can safely remove this argument.
  times = pd.to_datetime(

clausmichele:
Ok, they are triggered by [stackstac](https://github.com/gjoseph92/stackstac) and [openeo-pg-parser-networkx](https://github.com/Open-EO/openeo-pg-parser-networkx/).

I will open a PR for fixing the warning in stackstac and for setting the log level to debug in openeo-pg-parser-networkx.
However, I would like to merge this PR and have this example notebook ready in the meanwhile.

clausmichele:
@soxofaan can we merge this please?

soxofaan:
ok merged

PR #475, opened 2023-09-14
Updated UDP documentation

Pratichhya:
Updates the examples for UDP. However, I would need help/guidance on:

1. How to differentiate between:  Through “process functions” and  From a parameterized data cube (https://github.com/Open-EO/openeo-python-client/blob/udp_description/docs/udp.rst#build-and-save-using-process-functions)?
2. How to include 'subtype' in the parameter scheme using a helper function such as "Parameter.array()"?

### DIFF
--- a/docs/udp.rst
+++ b/docs/udp.rst
@@ -447,11 +447,9 @@ Here, let us consider that we want to recalculate the EVI-timeseries
 as done in :ref:`example-use-case-evi-map-and-timeseries` but for a different area.
 Therefore for future, instead of redefining the whole process, we simply 
 want to define it as a UDP ``EVI_timeseries``, that can be easily shared 
-and reused later without dealing with the entire workflow each time.
-
+and reused later without dealing with the entire workflow each tim:
 
 .. code-block:: python
-    :caption: EVI Timeseries UDP example
 
     import openeo
     from openeo.api.process import Parameter
@@ -527,7 +525,6 @@ Now, we can use use :func:`~openeo.rest.connection.Connection.datacube_from_proc
 to fetch the saved process for further use.
 
 .. code-block:: python
-    :linenos:
 
     date = ["2020-01-01", "2021-12-31"]
     bbox = {"west": 5.14, "south": 51.17, "east": 5.17, "north": 51.19}
--- a/docs/udp.rst
+++ b/docs/udp.rst
@@ -255,6 +255,7 @@ In addition, the expected type of the array items can also be specified,
 e.g. an array of integers:
 
 .. code-block:: json
+
     {
         "type": "array",
         "items": {"type": "integer"}
@@ -264,8 +265,6 @@ Another, more complex type is ``{"type": "object"}`` for parameters
 that are like Python dictionaries (or mappings).
 For example, to define a bounding box parameter
 that should contain certain fields with certain type::
-
-.. code-block:: json
 
     {
         "type": "object",
@@ -287,7 +286,6 @@ in the `openeo-processes project <https://github.com/Open-EO/openeo-processes>`_
 
 For example, as defined for the ``EVI_timeseries``
 
-.. code-block:: python
     schema = {
         "type": "object",
         "subtype": "geojson"
@@ -326,7 +324,9 @@ The openEO Python Client Library defines the
 official processes in the :py:mod:`openeo.processes` module,
 which can be used to build a process graph as follows:
 
-from openeo.processes import subtract, divide
+.. code-block:: python
+
+    from openeo.processes import subtract, divide
     from openeo.api.process import Parameter
 
     # Define the input parameter.
@@ -447,9 +447,10 @@ while building a :class:`~openeo.rest.datacube.DataCube`.
     :class:`~openeo.api.process.Parameter` arguments.
     Please submit a bug report when you encounter missing or wrong parameterization support.
 
-We can now store this as a user-defined process called "Hello_openEO" on the back-end::
+We can now store this as a user-defined process called "Hello_openEO" on the back-end:
 
 .. code-block:: python         
+
     # Store user-defined process in openEO back-end.
     connection.save_user_defined_process(
         user_defined_process_id="EVI_timeseries",
@@ -460,7 +461,7 @@ We can now store this as a user-defined process called "Hello_openEO" on the bac
 
 If you want to inspect its openEO-style process graph representation,
 use the :meth:`~openeo.rest.datacube.DataCube.to_json()`
-or :meth:`~openeo.rest.datacube.DataCube.print_json()` method::
+or :meth:`~openeo.rest.datacube.DataCube.print_json()` method.
 
 .. code-block:: python
 
--- a/docs/udp.rst
+++ b/docs/udp.rst
@@ -92,7 +92,6 @@ the openEO Python client will hide this normally).
 The important point here is the parameter reference ``{"from_parameter": "collection_id"}`` or 
 ``{"from_parameter": "aoi"}`` in the above process graph.
 When we call this user-defined process we will have to provide a value.
-For example with  degrees Fahrenheit (again in openEO JSON format here)
 
 .. code-block:: json
     {
@@ -151,7 +150,7 @@ Building and storing user-defined process
 There are a couple of ways to build and store user-defined processes:
 
 - using predefined :ref:`process functions <create_udp_through_process_functions>`
-- :ref:`parameterized building of a data cube <create_udp_parameterized_cube>`
+- :ref:`parameterized building of a datacube <create_udp_parameterized_cube>`
 - :ref:`directly from a well-formatted dictionary <create_udp_from_dict>` process graph representation
 
 
@@ -179,7 +178,7 @@ Some useful parameter helpers (class methods of the :py:class:`~openeo.api.proce
     e.g. to parameterize the a band selection  in a ``load_collection`` call in your UDP.
 -   :py:meth:`Parameter.datacube() <openeo.api.process.Parameter.datacube>`
     (or its legacy, deprecated cousin :py:meth:`Parameter.raster_cube() <openeo.api.process.Parameter.raster_cube>`)
-    to create a data cube parameter.
+    to create a datacube parameter.
 
 Consult the documentation of these helper class methods for additional features.
 
@@ -241,7 +240,7 @@ For example, as defined for the ``download_subset``
         "subtype": "collection-id",
     }
     
-Additionally, the schema of an openEO data cube is:
+Additionally, the schema of an openEO datacube is:
 
 .. code-block:: json
     {
@@ -259,7 +258,7 @@ Building, saving and storing user-defined process
 There are a couple of ways to build and store user-defined processes:
 
 - using predefined :ref:`process functions <create_udp_through_process_functions>`
-- :ref:`parameterized building of a data cube <create_udp_parameterized_cube>`
+- :ref:`parameterized building of a datacube <create_udp_parameterized_cube>`
 - :ref:`directly from a well-formatted dictionary <create_udp_from_dict>` process graph representation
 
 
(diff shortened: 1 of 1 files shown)


docs/udp.rst
```diff
@@ -34,79 +34,188 @@ This turns your openEO process into a web application
...
-    c = (f - 32) / 1.8
+.. code-block:: python
 
-you could define a user-defined process ``fahrenheit_to_celsius``,
-consisting of two simple mathematical operations
-(pre-defined processes ``subtract`` and ``divide``).
+    datacube = connection.load_collection(
+            collection_id = collection,
+            temporal_extent = temporal_interval
+            spatial_extent = polygon,
```
soxofaan, line 87: again it is not clear what these variables `collection`, `temporal_interval` ... are

docs/udp.rst
```diff
@@ -34,79 +34,188 @@ This turns your openEO process into a web application
...
+            bands = bands
+            )
 
-We can represent this in openEO's JSON based format as follows
+
+Moreover, you have the flexibility to predefine any of these 
+parameters as constants, just as we did earlier.
```
soxofaan, line 93: I would not use the term `constant` here, because you are using variables

docs/udp.rst
```diff
@@ -34,79 +34,188 @@ This turns your openEO process into a web application
...
+
+As shown above, one of the predefined processes used here is 
+``load_collection``.
```
soxofaan, line 96: I'm not sure what this statement is for or what it refers ("above") to

docs/udp.rst
```diff
@@ -34,79 +34,188 @@ This turns your openEO process into a web application
...
+        "arguments": {
+          "bands": [
+            "B02",
+            "B04",
+            "B08",
+            "SCL"
+          ],
+          "id": "SENTINEL2_L2A",
+          "spatial_extent": {
+            "from_parameter": "bbox"
```
soxofaan, line 116: The link between "bbox" here and "polygon" higher is not clear

docs/udp.rst
```diff
@@ -34,79 +34,188 @@ This turns your openEO process into a web application
...
-        description="Degrees Fahrenheit",
-        schema={"type": "number"}
+    temporal_interval = Parameter(
+        name = "temporal_interval",
+        description = "The date range to load.",
+        schema = {"type": "array", "subtype": "temporal-interval"},
+        default = ["2018-06-15", "2018-06-27"]
+    )
+    polygon = Parameter(
+        name = "bbox",
```
soxofaan, line 173: as mentioned above, it is confusing to link "polygon" (which sounds very geojson-like) with "bbox" (which is just a west-east-south-north thing)

docs/udp.rst
```diff
@@ -34,79 +34,188 @@ This turns your openEO process into a web application
...
-        name="fahrenheit", description="Degrees Fahrenheit"
+In the example above, the "temporal_interval" parameter (an array) can also be created more compactly
+with the :py:meth:`Parameter.array() <openeo.api.process.Parameter.array>` helper.
+
+.. code-block:: python
+
+    polygon = Parameter.array(
+        name = "bbox",
+        description = "The bounding box to load.",
+        subtype: "geojson"
```
soxofaan, line 194: invalid python here

docs/udp.rst
```diff
@@ -177,7 +283,19 @@ for even more features.
 On top of these generic types, the openEO API also defines a couple of custom (sub)types
 in the `openeo-processes project <https://github.com/Open-EO/openeo-processes>`_
 (see the ``meta/subtype-schemas.json`` listing).
-For example, the schema of an openEO data cube is::
+
+For example, as defined for the ``EVI_timeseries``
+
+.. code-block:: json
+
+    schema = {
```
soxofaan, line 291: `schema =`  is not valid JSON syntax

docs/udp.rst
```diff
@@ -291,37 +449,62 @@ while building a :class:`~openeo.rest.datacube.DataCube`.
...
+We can now store this as a user-defined process called "Hello_openEO" on the back-end:
+
+.. code-block:: python         
 
+    # Store user-defined process in openEO back-end.
     connection.save_user_defined_process(
-        "fancy_load_collection",
-        cube,
-        parameters=[spatial_extent]
+        user_defined_process_id="EVI_timeseries",
```
soxofaan, line 458: prefer lowercase (+ underscore) for process and UDP names

docs/udp.rst
```diff
@@ -335,70 +518,48 @@ Another developer already prepared it for you,
...
+.. code-block:: python
 
-    connection.save_user_defined_process(
-        user_defined_process_id="fahrenheit_to_celsius",
-        process_graph=fahrenheit_to_celsius,
-        parameters=[Parameter.number(name="f", description="Degrees Fahrenheit")]
+    shared_graph = connection.datacube_from_json(
+        'shared_process_graph.json',
+        parameters={
+            "temporal_interval": date,
```
soxofaan, line 531: `date` is confusing here as it looks like a single date, but you have to pass a two-item list/tuple here right?

docs/udp.rst
```diff
@@ -407,29 +568,66 @@ Evaluate user-defined processes
...
+    created_process = conn.datacube_from_process(
+        process_id = "EVI_timeseries",
+        temporal_interval = date,
+        box = bbox,
+        feature = aoi
+    )
+See :ref:`datacube_from_process` for more information on :func:`~openeo.rest.connection.Connection.datacube_from_process`.
 
+Additionally, if you wish to share your UDP with a wider audience,
+you can register it in the `openEO Algorithm Plaza <https://marketplace-portal.dataspace.copernicus.eu/catalogue>`_.
```
soxofaan, line 613: I'm not sure we should put VITO/CDSE specific things in the generic openEO python client docs

soxofaan:
> How to include 'subtype' in the parameter scheme using a helper function such as "Parameter.array()"?

this is a feature request we have to address separately I guess

docs/udp.rst
```diff
@@ -433,3 +433,131 @@ which we can process further and download::
...
+Working Example: Create EVI Timeseries UDP
+==========================================
+
+In this first example, we'll create an openEO workflow script,
+including declaring parameters, building, storing and loading a UDP
+
+Here, let us consider that we want to recalculate the EVI-timeseries
+as done in :ref:`example-use-case-evi-map-and-timeseries` but for a different area.
+Therefore for future, instead of redefining the whole process, we simply 
+want to define it as a UDP ``EVI_timeseries``, that can be easily shared
```
soxofaan, line 449: note: UDP sharing is not an official openEO feature, so I would not mention that here.

docs/udp.rst
```diff
@@ -433,3 +433,131 @@ which we can process further and download::
...
+==========================================
+
+In this first example, we'll create an openEO workflow script,
+including declaring parameters, building, storing and loading a UDP
+
+Here, let us consider that we want to recalculate the EVI-timeseries
+as done in :ref:`example-use-case-evi-map-and-timeseries` but for a different area.
+Therefore for future, instead of redefining the whole process, we simply 
+want to define it as a UDP ``EVI_timeseries``, that can be easily shared 
+and reused later without dealing with the entire workflow each tim:
```
soxofaan, line 450:
```suggestion
and reused later without dealing with the entire workflow each time:
```

docs/udp.rst
```diff
@@ -433,3 +433,131 @@ which we can process further and download::
...
+    temporal_interval = Parameter(
+        name="temporal_interval",
+        description="The date range to load.",
+        schema={"type": "array", "subtype": "temporal-interval"},
+        default =["2018-06-15", "2018-06-27"]
+    )
+    bbox = Parameter(
+        name="bbox",
+        description="The bounding box to load.",
+        schema={"type": "object", "subtype": "geojson"},
```
soxofaan, line 470: `"subtype": "geojson"` is for GeoJSON objects, while this is a west-south-east-north bbox, which is not GeoJSON

docs/udp.rst
```diff
@@ -433,3 +433,131 @@ which we can process further and download::
...
+        default={
+            "west": 5.14,
+            "south": 51.17,
+            "east": 5.17,
+            "north": 51.19,
+            "crs": 4326,
+        }
+    )
+    feature = Parameter(
+        name="feature",
```
soxofaan, line 480: In most user and API documentation we use "geometry" instead of "feature" for this kind of object

docs/udp.rst
```diff
@@ -433,3 +433,131 @@ which we can process further and download::
...
+        description="Provide polygon as FeatureCollection for extracting the EVI (Enhanced Vegetation Index) timeseries for these regions.",
```
soxofaan, line 481: > polygon as FeatureCollection

Not clear with is meant here: should the user provide a single GeoJSON Polygon and wrap it in a GeoJSON feature collection themself? Should user provide a single GeoJSON Polygon which will automatically be converted to FeatureCollection? 

In reality user can provide anything: Single (Multi)Polygon, single Feature or a FeatureCollection

docs/udp.rst
```diff
@@ -433,3 +433,131 @@ which we can process further and download::
...
+        schema={"type": "object", "subtype": "geojson"},
+        default={
```
soxofaan, line 483: this default is not valid GeoJSON. I don't think there should be a default here

docs/udp.rst
```diff
@@ -433,3 +433,131 @@ which we can process further and download::
...
+        reducer = "mean",
+    )
+
+    # Store user-defined process in openEO back-end.
+    process_name = "EVI_timeseries"
+    connection.save_user_defined_process(
+        user_defined_process_id=process_name,
+        process_graph=evi_aggregation,
+        parameters=[temporal_interval,bbox,feature],
+        public="true",
```
soxofaan, line 517: This `public` is non-standard, and not necessary for the discussion, so I would not mention it

docs/udp.rst
```diff
@@ -433,3 +433,131 @@ which we can process further and download::
...
+    nir = sentinel2_cube.band("B08") * 0.0001
+    evi = 2.5 * (nir - red) / (nir + 6.0 * red - 7.5 * blue + 1.0)
+
+    evi_aggregation = evi_masked.aggregate_spatial(
+        geometries = feature,
+        reducer = "mean",
+    )
+
+    # Store user-defined process in openEO back-end.
+    process_name = "EVI_timeseries"
```
soxofaan, line 512: it's more standard to use "process_id" instead of "process_name"

docs/udp.rst
```diff
@@ -433,3 +433,131 @@ which we can process further and download::
...
+Now, we can use use :func:`~openeo.rest.connection.Connection.datacube_from_process`
+to fetch the saved process for further use.
+
+.. code-block:: python
+
+    # pass the input parameters
+    dates = ["2020-01-01", "2021-12-31"]
+    box = {"west": 5.14, "south": 51.17, "east": 5.17, "north": 51.19}
+    features = {
+        "type": "FeatureCollection",
```
soxofaan, line 529: It's not necessary to wrap the polygon in a feature collection. Providing a single polygon will work and makes the snippet shorter

docs/udp.rst
```diff
@@ -433,3 +433,131 @@ which we can process further and download::
...
+            "crs": 4326,
+        }
+    )
+
+    # Load raw collection data
+    sentinel2_cube = connection.load_collection(
+        "SENTINEL2_L2A",
+        spatial_extent = bbox,
+        temporal_extent = temporal_interval,
+        bands = ["B02", "B04", "B08", "SCL"],
```
soxofaan, line 497: SCL is not used, so I would not include it here

docs/udp.rst
soxofaan: Also: If you use `aggregate_spatial`  already, it is not necessary to use `spatial_extent` in `load_collection` because that will be inferred automatically.
Dropping `bbox`  will make this example more compact and to the point I think

docs/udp.rst
```diff
@@ -433,3 +433,131 @@ which we can process further and download::
...
+
+    import openeo
+    from openeo.api.process import Parameter
+
+    # Create connection to openEO back-end
+    connection = openeo.connect("...").authenticate_oidc()
+
+    # declaring parameters
+    temporal_interval = Parameter(
+        name="temporal_interval",
```
soxofaan, line 462: I would name this parameter `temporal_extent` iso `temporal_interval` to stay closer to the standard

docs/udp.rst
soxofaan: Also note that the standard processes don't user upper case in their id, so I would use process id `evi_timeseries` or even just `evi`

docs/udp.rst
```diff
@@ -433,3 +433,131 @@ which we can process further and download::
     cube.download("fancy.tiff", format="GTiff")
 
 See :ref:`datacube_from_process` for more information on :func:`~openeo.rest.connection.Connection.datacube_from_process`.
+
+
+.. _workflow_example:
```
soxofaan, line 438: use a less generic, more descriptive reference, e.g.

```suggestion
.. _udp_example_evi:
```

docs/udp.rst
```diff
@@ -433,3 +433,131 @@ which we can process further and download::
...
+
+Working Example: Create EVI Timeseries UDP
```
soxofaan, line 440: Just asking: any reason to call this a "working" example? The Fahrenheit example above  works too

I would just call this something like

```suggestion
UDP Example: EVI timeseries
```

docs/udp.rst
```diff
@@ -433,3 +433,131 @@ which we can process further and download::
...
+==========================================
+
+In this first example, we'll create an openEO workflow script,
+including declaring parameters, building, storing and loading a UDP
+
+Here, let us consider that we want to recalculate the EVI-timeseries
+as done in :ref:`example-use-case-evi-map-and-timeseries` but for a different area.
```
soxofaan, line 447: Where does this `example-use-case-evi-map-and-timeseries` point to? This does not seem to exist already, unless I'm misunderstanding

docs/udp.rst
Pratichhya: Noted. Removed the parameter  bbox within the load collection and used feature alone

docs/udp.rst
Pratichhya: renamed it as geometry

docs/udp.rst
Pratichhya: Rephrase the sentence as:

Provide a polygon that represents the specific areas where you want to calculate the 'EVI' (Enhanced Vegetation Index) over a period of time.

docs/udp.rst
Pratichhya: sure, removed the variable.

docs/udp.rst
Pratichhya: mistakenly left out. Corrected

docs/udp.rst
Pratichhya: noted and corrected. Didn't think about it; thank you for correcting me.

docs/udp.rst
Pratichhya: removed

docs/udp.rst
Pratichhya: updated

soxofaan:
merged in d42a87498b4a717fd64907ec4f4e28c3fe0ed484

thanks

PR #480, opened 2023-09-29
add reduce_spatial

clausmichele:
`reduce_spatial` added to the DataCube class.
Created the necessary function to handle the metadata.


openeo/metadata.py
```diff
@@ -448,6 +448,18 @@ def reduce_dimension(self, dimension_name: str) -> CollectionMetadata:
...
+    def reduce_spatial(self) -> CollectionMetadata:
+        """Create new metadata object by reducing the spatial dimensions."""
+        dimensions = None
+        for d in self.spatial_dimensions:
+            dimension_name = d.name
+            loc = self.dimension_names().index(dimension_name)
+            if dimensions is None:
+                dimensions = self._dimensions[:loc] + self._dimensions[loc + 1 :]
+            else:
+                dimensions = dimensions + self._dimensions[:loc] + self._dimensions[loc + 1 :]
```
soxofaan, line 460: I think you can simplify this whole for loop to something like 

```python
dimensions = [d for d in self._dimensions if not isinstance(d, SpatialDimension)]
```

openeo/rest/datacube.py
```diff
@@ -1141,6 +1141,40 @@ def reduce_dimension(
             metadata=self.metadata.reduce_dimension(dimension_name=dimension) if self.metadata else None,
         )
 
+    @openeo_process
+    def reduce_spatial(
+        self,
+        reducer: Union[str, typing.Callable, UDF, PGNode],
+        context: Optional[dict] = None,
+        process_id="reduce_spatial",
+        band_math_mode: bool = False,
```
soxofaan, line 1150: Is there a reason to expose `process_id` and `band_math_mode` here? Or is that from copying `reduce_dimension`. If the latter: don't provide these arguments

openeo/rest/datacube.py
```diff
@@ -1141,6 +1141,40 @@ def reduce_dimension(
...
+            and returns a single numerical value.
+            For example:
+
+            -   ``"mean"`` (string)
+            -   :py:func:`absolute <openeo.processes.max>` (:ref:`predefined openEO process function <openeo_processes_functions>`)
+            -   ``lambda data: data.min()`` (function or lambda)
+
+        :param context: Additional data to be passed to the process.
+        """
+        reducer = self._get_callback(process=reducer, parent_parameters=["data", "context"], connection=self.connection)
```
soxofaan, line 1171: `_get_callback` doesn't exist anymore. It's `build_child_callback` now

openeo/rest/datacube.py
```diff
@@ -1141,6 +1141,40 @@ def reduce_dimension(
...
+
+        return self.process_with_node(
```
soxofaan, line 1173: I don't think it's necessary to use `process_with_node` here. Just using `self.process()` will do.

clausmichele: I just tried. Using self.process leads to an error. `process_with_node` work fine instead.

```
File ~\Documents\GitHub\openeo-python-client\openeo\internal\graph_building.py:298, in FlatGraphNodeIdGenerator.generate(self, process_id)
    296 def generate(self, process_id: str):
    297     """Generate new key for given process id."""
--> 298     self._counters[process_id] += 1
    299     return "{p}{c}".format(p=process_id.replace('_', ''), c=self._counters[process_id])

TypeError: unhashable type: 'PGNode'
```

soxofaan: I guess you tried something like `self.process(PGNode(...`
This should work however:
```python
self.process(
    process_id="reduce_spatial",
    data=THIS,
    reducer=reducer,
    context=context,
    metadata=...
```

soxofaan:
can you also add a CHANGELOG entry and a test like this one https://github.com/Open-EO/openeo-python-client/blob/9e62255faf467ed3460fcdcb0c72e53277c907be/tests/rest/datacube/test_datacube100.py#L1211-L1224

clausmichele:
@soxofaan done! The pre-commit fixed also the style of some parts of test_datacube100.py

soxofaan:
merged. Thanks!

PR #502, opened 2023-11-14
added list_udf_runtimes for localconnection

VincentVerelst:


openeo/local/connection.py
```diff
@@ -253,6 +253,15 @@ def load_stac(
         cube.metadata = metadata
         return cube
 
+    def list_udf_runtimes(self) -> dict:
+        """
+        Loads all available UDF runtimes.
+
+        :return: data_dict: Dict All available UDF runtimes
+        """
+        runtimes = {"Python": "Python"}
```
soxofaan, line 262: see https://openeo.org/documentation/1.0/developers/api/reference.html#tag/Capabilities/operation/list-udf-runtimes for expected schema here. e.g. should be something like

```json
{
  "Python": {
    "title": "Python 3",
    "type": "language",
    "versions": {"3": {"libraries": {}}},
    "default": "3"
  }
}
```

openeo/local/connection.py
```diff
@@ -253,6 +253,15 @@ def load_stac(
         cube.metadata = metadata
         return cube
 
+    def list_udf_runtimes(self) -> dict:
+        """
+        Loads all available UDF runtimes.
+
+        :return: data_dict: Dict All available UDF runtimes
```
soxofaan, line 260:
```suggestion
        :return: All available UDF runtimes
```

openeo/local/connection.py
soxofaan: FYI: also see https://openeo.vito.be/openeo/1.1/udf_runtimes for what VITO backend currently returns in practice

Issue #499, opened 2023-11-08
`connection.execute()`: only JSON-decode when response is JSON


soxofaan:
In #498 a user used `con.execute()` trying to get a GeoTIFF, but that failed because `con.execute()` is doing JSON decoding by default:

https://github.com/Open-EO/openeo-python-client/blob/d1e77e2fc9ab262cf2f55c602a59fd50caada083/openeo/rest/connection.py#L1578

this should only be done if response mime type is actually JSON.
And return raw bytes otherwise?

VictorVerhaert:
Added a simple try/catch approach. However not quite sure how to test this.

PR #529, opened 2024-01-18
catch json decode errors in connection.execute

VictorVerhaert:
github issue #499
added try/except to connection.execute to catch json decode errors
If a json decode error occurs, the response is return raw. 
(Follows [eafp](https://docs.python.org/2/glossary.html#term-eafp))

### DIFF
--- a/openeo/rest/connection.py
+++ b/openeo/rest/connection.py
@@ -1600,11 +1600,10 @@ class Connection(RestApiConnection):
         if auto_decode:
             try:
                 return response.json()
-            except requests.exceptions.JSONDecodeError:
-                _log.warning(
+            except requests.exceptions.JSONDecodeError as e:
+                raise OpenEoClientException(
                     "Failed to decode response as JSON. For other data types use `download` method instead of `execute`."
-                )
-                raise
+                ) from e
         else:
             return response
 
--- a/openeo/rest/connection.py
+++ b/openeo/rest/connection.py
@@ -1578,23 +1578,27 @@ class Connection(RestApiConnection):
         validate: Optional[bool] = None,
     ):
         """
-        Execute a process graph synchronously and return the result (assumed to be JSON).
+        Execute a process graph synchronously and return the result. If the result is a JSON object, it will be parsed.
 
         :param process_graph: (flat) dict representing a process graph, or process graph as raw JSON string,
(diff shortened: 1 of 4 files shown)


soxofaan:
FYI: you have here a couple of tests that a close to what you want to test (but use `.download()` iso `.execute()`): 
https://github.com/Open-EO/openeo-python-client/blob/7ce92925dd9544ff116b14c0339de203f854ec27/tests/rest/datacube/test_datacube.py#L533-L546

openeo/rest/connection.py
```diff
@@ -1578,23 +1578,27 @@ def execute(
...
             path="/result",
             json=pg_with_metadata,
             expected_status=200,
             timeout=timeout or DEFAULT_TIMEOUT_SYNCHRONOUS_EXECUTE,
-        ).json()  # TODO: only do JSON decoding when mimetype is actually JSON?
+        )
+        try:
+            return response.json()
+        except requests.exceptions.JSONDecodeError:
+            return response
```
soxofaan, line 1601: I think the raw bytes should be returned here instead of the response object

soxofaan:
just to avoid that we lose track of what needs to be done here:
- [ ] add tests

openeo/rest/connection.py
```diff
@@ -1576,25 +1576,36 @@ def execute(
...
+        Execute a process graph synchronously and return the result. If the result is a JSON object, it will be parsed.
 
         :param process_graph: (flat) dict representing a process graph, or process graph as raw JSON string,
             or as local file path or URL
         :param validate: Optional toggle to enable/prevent validation of the process graphs before execution
             (overruling the connection's ``auto_validate`` setting).
+        :param auto_decode: Boolean flag to enable/disable automatic JSON decoding of the response. Defaults to True.
 
-        :return: parsed JSON response
+        :return: if possible parsed JSON response, otherwise raw response
```
soxofaan, line 1590: this return doc is not completely consistent with the "auto_decode" feature

soxofaan (commented):
some minor finetune notes

soxofaan:
Practically ready to merge.

can you also add a CHANGELOG.md entry?

soxofaan:
thanks and congrats with the first PR  :tada:

Issue #528, opened 2024-01-17
use default chunk_size in download


soxofaan:
download currently uses `None` chunk_size, which apparently is not ideal at the moment (risk of loading download first in memory before dumping to disk)
see https://github.com/Open-EO/openeo-python-client/pull/524 and https://github.com/Open-EO/openeo-geopyspark-driver/issues/631

I think we can just start here by using a default of say 10MB. If there is need for making this configurable, we can add that later, but I would expect that a better default than `None` will cover most use cases

PR #535, opened 2024-02-06
changed default download chunk size to 10MB

VictorVerhaert:
issue #528
this affects response.iter_content in the following functions:
- openeo.rest.connection.Connection.download
- openeo.rest.job.Job.download
- openeo.rest.userfile.UserFile.download

in Job.download the chunk size can still be overwritten by the user

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 ### Changed
-- Changed default `chunk_size` of various `download` functions from None to 10MB. This improves the handling of large downloads and reduces memory usage.
+- Changed default `chunk_size` of various `download` functions from None to 10MB. This improves the handling of large downloads and reduces memory usage. ([#528](https://github.com/Open-EO/openeo-python-client/issues/528))
 - `Connection.execute()` and `DataCube.execute()` now have a `auto_decode` argument. If set to True (default) the response will be decoded as a JSON and throw an exception if this fails, if set to False the raw `requests.Response` object will be returned. ([#499](https://github.com/Open-EO/openeo-python-client/issues/499))
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 ### Changed
-- Changed default `chunk_size` of various `download` functions from None to 10MB. This should improve download speed of large files.
+- Changed default `chunk_size` of various `download` functions from None to 10MB. This improves the handling of large downloads and reduces memory usage.
 - `Connection.execute()` and `DataCube.execute()` now have a `auto_decode` argument. If set to True (default) the response will be decoded as a JSON and throw an exception if this fails, if set to False the raw `requests.Response` object will be returned. ([#499](https://github.com/Open-EO/openeo-python-client/issues/499))
 ### Removed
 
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -10,8 +10,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 ### Changed
+- Changed default `chunk_size` of various `download` functions from None to 10MB. This should improve download speed of large files.
```
soxofaan, line 13: I'm not sure it will improve the speed, I would just say that it improves handling of large downloads

openeo/rest/connection.py
```diff
@@ -1566,7 +1566,7 @@ def download(
 
         if outputfile is not None:
             with Path(outputfile).open(mode="wb") as f:
-                for chunk in response.iter_content(chunk_size=None):
+                for chunk in response.iter_content(chunk_size=10000000):
```
soxofaan, line 1569: create a constant DEFAULT_DOWNLOAD_CHUNK_SIZE in connection.py for this
and reuse it in the other places too

VictorVerhaert: This causes an ImportError due to circular imports. I think this should be set in a config file but I think that this functionality is not fully developed in openeo yet?

I can prevent the circular imports by defining the constant before the openeo imports in connection.py but this seems like bad coding to me

soxofaan: I'd define the constant in openeo/rest/__init__.py to circumvent circular import issues

openeo/rest/__init__.py
```diff
@@ -2,6 +2,9 @@
 
 from openeo import BaseOpenEoException
 
+# TODO: get from config file
+DEFAULT_DOWNLOAD_CHUNK_SIZE = 10000000  # 10MB
```
soxofaan, line 6: FYI in python you can make large numbers more readable with `_`
```suggestion
DEFAULT_DOWNLOAD_CHUNK_SIZE = 10_000_000  # 10MB
```

openeo/rest/job.py
```diff
@@ -351,14 +357,17 @@ def __repr__(self):
             n=self.name, t=self.metadata.get("type", "unknown"), h=self.href
         )
 
-    def download(self, target: Optional[Union[Path, str]] = None, chunk_size=None) -> Path:
+    def download(self, target: Optional[Union[Path, str]] = None, chunk_size: Optional[int] = None) -> Path:
```
soxofaan, line 360: I would change the default here directly:
```suggestion
    def download(self, target: Optional[Union[Path, str]] = None, chunk_size: int = DEFAULT_DOWNLOAD_CHUNK_SIZE) -> Path:
```

That simplifies implementation and is self-documenting.
And otherwise the user can't intentionally disable chunking anymore if they want that for some reason

openeo/rest/job.py
```diff
@@ -351,14 +357,17 @@ def __repr__(self):
...
         """
         Download asset to given location
 
         :param target: download target path. Can be an existing folder
             (in which case the filename advertised by backend will be used)
             or full file name. By default, the working directory will be used.
+        :param chunk_size: size of chunks to download (in bytes). If None, connection.DEFAULT_DOWNLOAD_CHUNK_SIZE will be used.
```
soxofaan, line 367:
```suggestion
        :param chunk_size: size of chunks to download (in bytes).
```
see above

openeo/rest/job.py
```diff
@@ -351,14 +357,17 @@ def __repr__(self):
...
         """
+        if not chunk_size:
+            chunk_size = DEFAULT_DOWNLOAD_CHUNK_SIZE
```
soxofaan, line 370: not necessary (see above)

Issue #464, opened 2023-08-24
CollectionMetadata ->  CubeMetadata?


soxofaan:
https://github.com/Open-EO/openeo-python-client/blob/cbd40533d6158c0288d31bd66f35db62e1304aa9/openeo/metadata.py#L198-L200

"CollectionMetadata" got its name when data cubes where still called "ImageCollection" (instead of "DataCube"). But now that is getting a bit confusing: in openEO, "collections" is the source data one loads with `load_collection`, while "cubes" are the intermediate data structure one works with. So "CollectionMetadata" should better be called "CubeMetadata". 

Also, CollectionMetadata contains some specific dimension metadata parsing logic (cube:dimensions/eo:bands), which is closely tied to the STAC description of collections. I think it would be better to separate this from the essential cube metadata logic.

soxofaan:
I think we should change it to:

- `CubeMetadata`: just the bare metadata essentials for a data cube: interface to interact with the cube dimensions and their labels (if available), maybe some resolution info too. Methods allow to construct a new `CubeMetadata` instance from an existing one (e.g. `reduce_dimension` eliminates a dimension)
- `CollectionMetadata` (subclass of `CubeMetadata`): interface to interact with all  original collection (STAC) metadata. Calling a "process" like `reduce_dimension` produces a `CubeMetadata`

By making `CollectionMetadata` a subclass of `CubeMetadata` we can stay backward compatible I think, because the interface of  `CollectionMetadata` would not really change

jdries:
ok, makes sense!

soxofaan:
Another reason for better abstraction: `load_stac` should produce a data cube with metadata (#425)  , but that is not the same as collection metadata.

VictorVerhaert:
https://github.com/Open-EO/openeo-python-client/blob/498c4c7676f46d4ed6e858178cc08b1189a7699f/openeo/metadata.py#L231-L242
Currently functionality does not change due to this function.
As this function checks the type and returns an object of the same class, functions like `reduce_dimension` (who use this function to update a metadata object) will still return a CollectionMetadata object if that was the original object.

Changing this to return a CubeMetadata changes functionality and fails some existing tests.

An alternative is to provide an optional argument to set the expected returned class, however this feels more like a quick fix and might be confusing in the future.

Issue #527, opened 2024-01-15
No metadata for datacube loaded with `load_stac`


VincentVerelst:
Datacubes that are loaded from a STAC catalog using `load_stac` don't contain any metadata.

soxofaan:
related to https://github.com/Open-EO/openeo-python-client/issues/464 and https://github.com/Open-EO/openeo-python-client/issues/425

jdries:
This is also a duplicate of:
https://github.com/Open-EO/openeo-python-client/issues/368
We'll probably just want to clearly deprecate load_result, so that we only have to fix this one.

soxofaan:
FYI: `load_result` is a non-experimental process in openeo-processes 1.0.0 (it moved to "proposals" in 1.1.0), so strictly speaking we can not easily deprecate it in general. But I agree that we probably should not bother (bug)fixing it.

VincentVerelst:
TODO: add unit testing

Issue #516, opened 2023-12-14
metadata is None after `aggregate_spatial`


VincentVerelst:
After applying `aggregate_spatial` to a Rastercube, the resulting Vectorcube does not contain any metadata anymore.

soxofaan:
This is not as straightforward as hoped, as some vector cubes details are not settled yet. But maybe there are some quick wins already possible without having to over-engineer a general solution, so:

what do you want to use the metadata for? Any particular fields, e.g. you just want the band names?

VincentVerelst:
Yes, having the band names would be the most important.

soxofaan:
related to https://github.com/Open-EO/openeo-python-client/issues/464  https://github.com/Open-EO/openeo-python-client/issues/425  https://github.com/Open-EO/openeo-python-client/issues/527

jdries:
I would say aggregate_spatial preserves band names when going from raster to vector, so can we just copy that part of the metadata? Maybe @VincentVerelst  can prepare a PR for that?

Issue #425, opened 2023-05-10
Add `load_stac`


soxofaan:
important item under the umbrella of #424 : support for new `load_stac` process (replacement of existing `load_result`)

soxofaan:
already added a very basic `Connection.load_stac` in [6b3c539](https://github.com/Open-EO/openeo-python-client/commit/6b3c539f0019e9620909e15e71ba97ba9b656aa6)

soxofaan:
next steps:
- also load basic metadata as `cube.metadata`  (related to #464)

soxofaan:
The basics are implemented. This one can be closed

PR #541, opened 2024-02-23
Issue464 cube metadata

VictorVerhaert:
from issue #464

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - Changed default `chunk_size` of various `download` functions from None to 10MB. This improves the handling of large downloads and reduces memory usage. ([#528](https://github.com/Open-EO/openeo-python-client/issues/528))
 - `Connection.execute()` and `DataCube.execute()` now have a `auto_decode` argument. If set to True (default) the response will be decoded as a JSON and throw an exception if this fails, if set to False the raw `requests.Response` object will be returned. ([#499](https://github.com/Open-EO/openeo-python-client/issues/499))
-- Created superclass `CubeMetadata` for `CollectionMetadata` containing methods to handle basic metadata like dimensions and their labels, without requiring STAC information. ([#464](https://github.com/Open-EO/openeo-python-client/issues/464))
+- Introduced superclass `CubeMetadata` for `CollectionMetadata` for essential metadata handling (just dimensions for now) without collection-specific STAC metadata parsing. ([#464](https://github.com/Open-EO/openeo-python-client/issues/464))
 ### Removed
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - Changed default `chunk_size` of various `download` functions from None to 10MB. This improves the handling of large downloads and reduces memory usage. ([#528](https://github.com/Open-EO/openeo-python-client/issues/528))
 - `Connection.execute()` and `DataCube.execute()` now have a `auto_decode` argument. If set to True (default) the response will be decoded as a JSON and throw an exception if this fails, if set to False the raw `requests.Response` object will be returned. ([#499](https://github.com/Open-EO/openeo-python-client/issues/499))
+- Created superclass `CubeMetadata` for `CollectionMetadata` containing methods to handle basic metadata like dimensions and their labels, without requiring STAC information. ([#464](https://github.com/Open-EO/openeo-python-client/issues/464))
 ### Removed
 
 ### Fixed
--- a/openeo/metadata.py
+++ b/openeo/metadata.py
@@ -213,10 +213,10 @@ class CubeMetadata:
         self._temporal_dimension = None
 
         if dimensions is not None:
-            self.set_dimensions(dimensions=dimensions)
+            self._set_dimensions(dimensions=dimensions)
 
-    def set_dimensions(self, dimensions: List[Dimension]):
-        if dimensions == None:
+    def _set_dimensions(self, dimensions: List[Dimension]):
+        if dimensions is None:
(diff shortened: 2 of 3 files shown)


VictorVerhaert:
https://github.com/Open-EO/openeo-python-client/blob/498c4c7676f46d4ed6e858178cc08b1189a7699f/openeo/metadata.py#L231-L242
Currently functionality does not change due to this function.
As this function checks the type and returns an object of the same class, functions like `reduce_dimension` (who use this function to update a metadata object) will still return a CollectionMetadata object if that was the original object.

Changing this to return a CubeMetadata changes functionality and fails some existing tests.

An alternative is to provide an optional argument to set the expected returned class, however this feels more like a quick fix and might be confusing in the future.

soxofaan (commented):
some initial comments

soxofaan:
> Currently functionality does not change due to this function.

I think it's fine to keep the original type of the metadata, that will indeed avoid breakage at other places.

Priority at the moment is having a base metadata class that is not implicitly tied to collection metadata (e.g. for https://github.com/Open-EO/openeo-python-client/issues/425 https://github.com/Open-EO/openeo-python-client/issues/516 https://github.com/Open-EO/openeo-python-client/issues/527 etc)

openeo/metadata.py
```diff
@@ -228,121 +217,37 @@ def __init__(self, metadata: dict, dimensions: List[Dimension] = None):
...
                 self._temporal_dimension = dim
 
     def __eq__(self, o: Any) -> bool:
-        return isinstance(o, CollectionMetadata) and self._dimensions == o._dimensions
+        return isinstance(o, type(self)) and self._dimensions == o._dimensions
 
     def _clone_and_update(
         self, metadata: dict = None, dimensions: List[Dimension] = None, **kwargs
-    ) -> CollectionMetadata:
+    ) -> CubeMetadata:  # python >= 3.11: -> Self to be more correct for subclasses
```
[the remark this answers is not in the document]
VictorVerhaert: removed the comment as this won't be relevant anytime soon

VictorVerhaert:
_orig_metadata has been removed from CubeMetadata.
This causes CollectionMetadata to override the _clone_and_update method in order to preserve the orig_metadata (CollectionMetadata cannot be initiated without metadata, as it should be imo). 

Unless we change _clone_and_update to always return a CubeMetadata this is the only clean solution I think.

CubeMetadata tests still need to be written.

openeo/metadata.py
```diff
@@ -199,149 +199,45 @@ def rename_labels(self, target, source) -> Dimension:
...
-        else:
-            self._dimensions = dimensions
+        self._dimensions = None
         self._band_dimension = None
         self._temporal_dimension = None
+
+        if dimensions is not None:
+            self.set_dimensions(dimensions=dimensions)
+
+    def set_dimensions(self, dimensions: List[Dimension]):
```
soxofaan, line 218: I don't think `set_dimensions` should be a public method. 
Ideally, the metadata classes should be immutable from the outside, so they should not have setters

openeo/metadata.py
```diff
@@ -199,149 +199,45 @@ def rename_labels(self, target, source) -> Dimension:
...
+        if dimensions == None:
```
soxofaan, line 219: FYI: python tip: equality to `None` should be done with `is None` instead of `== None`

openeo/metadata.py
soxofaan: I would also just keep this logic under `__init__`, which will make it easier to understand

soxofaan (commented):
some quick notes

openeo/metadata.py
VictorVerhaert: reason I took this out of __init__ is so that the init of CollectionMetadata could reuse this code after parsing dimensions from metadata. 
I added some todo's about type checking e.g. BandsDimension, that would need to be resolved on two places otherwise.

VictorVerhaert:
TODO:
- [ ] Write test cases specific for CubeMetadata

openeo/metadata.py
```diff
@@ -84,6 +84,10 @@ def __init__(self, name: str, extent: Union[Tuple[str, str], List[str]]):
     def rename(self, name) -> Dimension:
         return TemporalDimension(name=name, extent=self.extent)
 
+    def rename_labels(self, target, source) -> Dimension:
+        # TODO should we check if the extend has changed with the new labels?
```
soxofaan, line 88:
```suggestion
        # TODO should we check if the extent has changed with the new labels?
```

soxofaan (commented):
some more notes

about test_metadata.py: be careful about adding formatting changes everywhere, that makes it hard to to find the actual things that were changed

VictorVerhaert:
Because I changed the names of the metadata tests using CollectionMetadata to test_collectionmetadata, pre-commit refactored the whole tests causing a large diff.
I'll be more careful with this in the future.

I added an entry in the changelog and processed the last review.

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - Changed default `chunk_size` of various `download` functions from None to 10MB. This improves the handling of large downloads and reduces memory usage. ([#528](https://github.com/Open-EO/openeo-python-client/issues/528))
 - `Connection.execute()` and `DataCube.execute()` now have a `auto_decode` argument. If set to True (default) the response will be decoded as a JSON and throw an exception if this fails, if set to False the raw `requests.Response` object will be returned. ([#499](https://github.com/Open-EO/openeo-python-client/issues/499))
+- Created superclass `CubeMetadata` for `CollectionMetadata` containing methods to handle basic metadata like dimensions and their labels, without requiring STAC information. ([#464](https://github.com/Open-EO/openeo-python-client/issues/464))
```
soxofaan, line 15:
```suggestion
- Introduced superclass `CubeMetadata` for `CollectionMetadata` for essential metadata handling (just dimensions for now) without collection-specific STAC metadata parsing. ([#464](https://github.com/Open-EO/openeo-python-client/issues/464))
```

soxofaan (approved):

Issue #542, opened 2024-02-23
Pylance type check is not working


hoetmaaiers:
The load_connection function in VSCode returns me the types of the decorator. This can be solved by changing the type signature of the `openeo_process`.

Changing the current implementation https://github.com/Open-EO/openeo-python-client/blob/master/openeo/internal/documentation.py#L14

to this solves that problem:

```python
T = TypeVar('T')

def openeo_process(f: Optional[T] = None, process_id: Optional[str] = None, mode: Optional[str] = None) -> T:
```

Does this make sense to the OpenEO contributors? Is a PR a good idea for this?

@hghcomphys helped me figuring out this problem 🙏

soxofaan:
Sure, if this is something that improves the developer experience, we would be certainly open to something like this.

soxofaan:
done with #543 
thanks!

PR #543, opened 2024-02-26
Use generic type definition in the openeo_process decorator

hoetmaaiers:
This improves static type checkers ability to infer typing.

Related to: #542

### DIFF
--- a/openeo/internal/documentation.py
+++ b/openeo/internal/documentation.py
@@ -12,7 +12,7 @@ from typing import Callable, Optional, Tuple, TypeVar
 _process_registry = collections.defaultdict(list)
 
 
-T = TypeVar("T")
+T = TypeVar("T", bound=Callable)
 
 
 def openeo_process(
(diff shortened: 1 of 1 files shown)


openeo/internal/documentation.py
```diff
@@ -1,17 +1,23 @@
...
 from functools import partial
-from typing import Callable, Optional, Tuple
+from typing import Callable, Optional, Tuple, TypeVar
 
 # TODO: give this a proper public API?
 _process_registry = collections.defaultdict(list)
 
 
-def openeo_process(f: Optional[Callable] = None, process_id: Optional[str] = None, mode: Optional[str] = None):
+T = TypeVar("T")
```
soxofaan, line 15: I'm not that familiar with TypeVar details yet, but I think it makes sense to define it with `bound=Callable` to keep the spirit of the original typehint of openeo_process

hoetmaaiers: Good suggestion, I updated this.

PR #547, opened 2024-03-12
Issue441 automatic backoff retry on http429

jdries:
@soxofaan  looks like something we can merge easily as this is mostly about using functionality from requests?

### DIFF
--- a/openeo/rest/http.py
+++ b/openeo/rest/http.py
@@ -1,0 +1,39 @@
+from typing import Set
+
+import requests
+import requests.adapters
+
+MAX_RETRIES = 3
+
+
+def requests_with_retry(
+    total: int = MAX_RETRIES,
+    read: int = MAX_RETRIES,
+    other: int = MAX_RETRIES,
+    status: int = MAX_RETRIES,
+    backoff_factor: float = 1,
+    status_forcelist: Set[int] = frozenset([429, 500, 502, 503, 504]),
+    **kwargs,
+) -> requests.Session:
+    """
+    Create a `requests.Session` with automatic retrying
+
+    Inspiration and references:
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -28,6 +28,7 @@
     "flake8>=5.0.0",
     "time_machine",
     "pyproj>=3.2.0",  # Pyproj is an optional, best-effort runtime dependency
+    "re-assert",
```
soxofaan, line 31: we already have dirty_equals for these kind of things, wee should not add yet another utility here

tests/rest/test_http.py
```diff
@@ -0,0 +1,40 @@
...
+    """Simple test for retrying using an invalid domain."""
+    caplog.set_level(logging.DEBUG)
+
+    session = requests_with_retry(total=0)
+    with pytest.raises(requests.exceptions.ConnectionError, match="Max retries exceeded"):
+        _ = session.get("https://example.test")
+
+    assert caplog.messages == [
+        "Starting new HTTPS connection (1): example.test:443",
+    ]
```
soxofaan, line 40: there are only tests with retries but ending in failure, there are no tests about retries that end up in success

soxofaan (commented):
some quick points to further finetune

openeo/rest/http.py
```diff
@@ -0,0 +1,39 @@
+from typing import Set
+
+import requests
+import requests.adapters
+
+MAX_RETRIES = 3
```
soxofaan, line 6: default of max 3 retries might be a bit low in practice in combination with default backoff_factor: last attempt will be roughly just be 4 seconds after first attempt, which might be too quick to stop trying

soxofaan:
Because of conflicts: reoding this PR in #770

PR #548, opened 2024-03-12
Issue527 load stac metadata

VincentVerelst:


openeo/metadata.py
```diff
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import logging
+import pystac
```
soxofaan, line 4: if pystac is imported here at top level like this, it has to be declared as global dependency in setup.py,
(not just as test dependency like we did originally)

openeo/metadata.py
```diff
@@ -522,3 +523,61 @@ def _repr_html_(self):
...
+def metadata_from_stac(url: str) -> CubeMetadata:
+    """
+    Reads the band metadata a static STAC catalog or a STAC API Collection and returns it as a :py:class:`CubeMetadata`
+
+    :param url: The URL to a static STAC catalog (STAC Item, STAC Collection, or STAC Catalog) or a specific STAC API Collection
+    :return: A :py:class:`CubeMetadata` containing the DataCube band metadata from the url.
+    """
+
+    def get_band_names(asst: pystac.Asset) -> List[Band]:
+        return [Band(eo_band["name"]) for eo_band in asst.extra_fields["eo:bands"]]
```
soxofaan, line 537: `Band` also has `common_name` and `wavelength_um` fields, so I would also extract these if they are available

openeo/metadata.py
```diff
@@ -522,3 +523,61 @@ def _repr_html_(self):
...
+
+def metadata_from_stac(url: str) -> CubeMetadata:
+    """
+    Reads the band metadata a static STAC catalog or a STAC API Collection and returns it as a :py:class:`CubeMetadata`
+
+    :param url: The URL to a static STAC catalog (STAC Item, STAC Collection, or STAC Catalog) or a specific STAC API Collection
+    :return: A :py:class:`CubeMetadata` containing the DataCube band metadata from the url.
+    """
+
+    def get_band_names(asst: pystac.Asset) -> List[Band]:
```
soxofaan, line 536: this function is only used once below, but it looks like it can be used more

openeo/metadata.py
```diff
@@ -522,3 +523,61 @@ def _repr_html_(self):
...
+
+    elif isinstance(stac_object, pystac.Collection):
+        collection = stac_object
+        band_names = [Band(b["name"]) for b in collection.summaries.lists.get("eo:bands", [])]
+
+        # Summaries is not a required field in a STAC collection, so also check the assets
+        for itm in collection.get_items():
+            band_assets = {
+                asset_id: asset
+                for asset_id, asset in dict(sorted(itm.get_assets().items())).items()
```
soxofaan, line 566: why sorting here? isn't that going to mix up existing metadata orders?

openeo/metadata.py
```diff
@@ -522,3 +523,61 @@ def _repr_html_(self):
...
+
+    def get_band_names(asst: pystac.Asset) -> List[Band]:
+        return [Band(eo_band["name"]) for eo_band in asst.extra_fields["eo:bands"]]
+
+    def is_band_asset(asset: pystac.Asset) -> bool:
+        return "eo:bands" in asset.extra_fields
+
+    stac_object = pystac.read_file(href=url)
+
+    band_names = []
```
soxofaan, line 544: you use "band name" in several places here (which sound like strings), but you are actually handling `Band` objects, so that might be confusing

openeo/rest/connection.py
```diff
@@ -1361,6 +1368,10 @@ def load_stac(
                 prop: build_child_callback(pred, parent_parameters=["value"]) for prop, pred in properties.items()
             }
         cube = self.datacube_from_process(process_id="load_stac", **arguments)
+        try:
+            cube.metadata = metadata_from_stac(url)
+        except Exception:
+            _log.warning("Python client could not read band metadata from URL.")
```
soxofaan, line 1374: make sure to include info about the thrown exception
```suggestion
            _log.warning(f"Failed to extract cube metadata from STAC URL {url}", exc_info=True)
```

setup.py
```diff
@@ -30,6 +30,7 @@
     "flake8>=5.0.0",
     "time_machine",
     "pyproj>=3.2.0",  # Pyproj is an optional, best-effort runtime dependency
+    "pystac"
```
soxofaan, line 33: as mentioned above, this should be a general dependency, not just test dependency

tests/test_metadata.py
```diff
@@ -782,3 +784,50 @@ def filter_bbox(self, bbox):
     assert isinstance(new, MyCubeMetadata)
     assert orig.bbox is None
     assert new.bbox == (1, 2, 3, 4)
+
+
+collection_json = {
```
soxofaan, line 789: I would not define these inputs as top level variables, as we are probably going to add a lot of variations

soxofaan (commented):
some quick notes

openeo/metadata.py
VincentVerelst: This does not sort the band names, but rather the assets in the item. In the backend, the assets are also sorted before the band metadata is read, so I included it here as well, to make sure there is no mismatch between the backend and client implementation.
https://github.com/Open-EO/openeo-geopyspark-driver/blob/7caaa9aa74585bad3c42e054cfb5dfe95ed31b55/openeogeotrellis/backend.py#L1003

tests/test_metadata.py
VincentVerelst: Ok, so I'll just plug the dictionaries directly into the @pytest.mark.parametrize decorator?

openeo/metadata.py
soxofaan: The openeo client should be usable with other backends than ours, so we should avoid hardcoding geopyspark behavior like that. And even for the geopyspark backend it's questioned if we should do that kind of sorting.
e.g. see https://github.com/Open-EO/openeo-processes/issues/488

tests/test_metadata.py
soxofaan: indeed, I would start with that,

In a further phase you could also define a helper function to dynamically generate stac-like dicts for various (parameterized) tests

VincentVerelst:
@soxofaan , addressed all your initial comments.

clausmichele:
@soxofaan @VincentVerelst I just seen this PR. The same functionality has been implemented by me here using the load_stac implementation: https://github.com/Open-EO/openeo-python-client/blob/d64fc30decfcf94860553f1b32027f5d1183099d/openeo/local/connection.py#L116

but it's quite slow since it's building the datacube view of the queried Collection to get the metadata. It would be good to align and get the metadata in the same way, I will look into how to port it to the LocalConnection.

Just a comment: I don't like that the "bands" dimension has an hardcoded "bands" label. As experienced during the openeo-test-suite, this information should come from the datacube extension if it's available (cube.dimensions ...)
https://github.com/Open-EO/openeo-python-client/pull/548/files#diff-de2cebdb79abf5b65ca416029a8be7bed5e377d451383eef92cbec604d773122R583

soxofaan:
@clausmichele indeed, this implementation is certainly open for further improvements and aligning. We just wanted to get some initial stuff in place to support some use cases

PR #559, opened 2024-04-17
Read band metadata from item-asset extension in STAC collection if possible

VincentVerelst:
Updated `metadata_from_stac` such that if the item-asset extension is present in a STAC collection, the client reads the band metadata from there instead of looping over all the assets.

### DIFF
--- a/openeo/metadata.py
+++ b/openeo/metadata.py
@@ -546,32 +546,47 @@ def metadata_from_stac(url: str) -> CubeMetadata:
     def is_band_asset(asset: pystac.Asset) -> bool:
         return "eo:bands" in asset.extra_fields
 
+    def bands_from_collection(collection: pystac.Collection) -> List[Band]:
+        # First check if the item_asset extension is present, if so, we can use that to get the band metadata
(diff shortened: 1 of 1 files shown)


openeo/metadata.py
```diff
@@ -546,32 +546,47 @@ def get_band_names(bands: List[Band]) -> List[str]:
     def is_band_asset(asset: pystac.Asset) -> bool:
         return "eo:bands" in asset.extra_fields
 
+    def bands_from_collection(collection: pystac.Collection) -> List[Band]:
+        # First check if the item_asset extension is present, if so, we can use that to get the band metadata
+        if pystac.extensions.item_assets.ItemAssetsExtension.has_extension(collection):
```
soxofaan, line 551: does this work? `pystac.extensions.item_assets` is not imported yet I think

soxofaan (commented):
is it possible to provide some unit tests about this.
It's a bit hard to review this without some more concrete examples

soxofaan:
Support for item-assets was also introduced by PR #577 so this one can be closed

PR #562, opened 2024-05-06
update localprocessing docs

clausmichele:
Added a reference on how to properly clone the openeo-processes-dask repository if someone wants to contribute with new processes or perform some tests modifying the existing ones.

This is a common issue several users faced already and the idea to add a reference here as well came from here:
https://github.com/Open-EO/openeo-processes-dask/issues/243#issuecomment-2095284938

### DIFF
--- a/docs/cookbook/localprocessing.rst
+++ b/docs/cookbook/localprocessing.rst
@@ -23,6 +23,15 @@ Installation
 
    pip install openeo[localprocessing]
 
+Development
+-----------
+
+If you want to contribute implementing new processes in openeo-processes-dask, clone the repository with `--recurse-submodules` to also fetch the process specs:
+
+.. code:: bash
+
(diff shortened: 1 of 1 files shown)


soxofaan:
I'm not sure this belongs in the docs of the openeo python client, this is a development aspect of the openeo-processes-dask repo. In this sense this looks too much.

On the other hand, the current PR also incomplete: you should not only clone the repo, but also properly install it  (in editable mode) in  your virtual environment.

If we want to add this to the python client docs, I would also put this at the bottom of the page (it feels too niche to put these technicalities so high on the page)

clausmichele:
Thanks for the feedback, I will think about it!

soxofaan:
@clausmichele   you're still interested in this, or can we close this PR (for now)?

clausmichele:
Hi @soxofaan, I will close it since as you pointed out, it's not the right place to document that.

PR #564, opened 2024-05-26
add more tips to udf docs

jdries:


docs/udf.rst
```diff
@@ -78,6 +78,18 @@ in smaller chunks and process these chunks separately (e.g. on isolated workers)
 Consequently, it's important that your **UDF algorithm operates correctly
 in such a chunked processing context**.
 
+A very common mistake is to use index-based array indexing, rather than name based. The index based approach
+assumes that datacube dimension order is fixed, which is not guaranteed. Next to that, it also reduces the readability
+of your code. Label based indexing is a great feature of xarray, and should be used whenever possible.
```
soxofaan, line 83: I think it's worth to put a more concrete example here

docs/udf.rst
```diff
@@ -78,6 +78,18 @@ in smaller chunks and process these chunks separately (e.g. on isolated workers)
...
+
+As a rule of thumb, the UDF should preserve the dimensions, their order, and shape of the input
```
soxofaan, line 85: I think users should not worry about dimension order (as long as they use something like Xarray), so I would not mention "should preserve ... their order" here

jdries: remove 'their order'

jdries:
@soxofaan  addressed one comment, adding an example takes more time. I would merge this important update now, so that users at least have some extra guidance.

Issue #567, opened 2024-06-04
Reduce_temporal throws MetadataException after load_stac


VictorVerhaert:
When using load_stac, the python client tries to read metadata from the provided stac. The only metadata currently being read is bands, so no spatial or temporal metadata.

When the python client fails to read metadata from the stac, metadata is simply set to None. 
These lines check is metadata is present and have default behaviour otherwise:
https://github.com/Open-EO/openeo-python-client/blob/36ecc2bdb0feaf4b7876819946866bd651a18189/openeo/rest/datacube.py#L1350-L1359
and 
https://github.com/Open-EO/openeo-python-client/blob/36ecc2bdb0feaf4b7876819946866bd651a18189/openeo/rest/datacube.py#L1216

The problem arises however when metadata IS being read from the STAC. As only bands are read `cube.metadata` is not None, but it does not contain a temporal dimension resulting in a MetadataException originating for example from: 
https://github.com/Open-EO/openeo-python-client/blob/36ecc2bdb0feaf4b7876819946866bd651a18189/openeo/metadata.py#L271-L275

The solution is threefold I think:
1. Not only check if the metadata is not None but also whether it contains a temporal dimension
2. Implement reading a temporal dimension (and spatial) from stac metadata
3. Catch metadata related errors and throw them as warnings (Discussion) as this blocks the execution of a batch job while it would run fine.

soxofaan:
as discussed, a workaround would be to keep the openeo package below 0.29 for now

VictorVerhaert:
Another workaround is to add the following:
`cube.metadata = cube.metadata.add_dimension("t", label=None, type="temporal")`

soxofaan:
Started with a PR at #568

soxofaan:
I already merged an initial fix that avoids throwing the `MetadataException`, so at least that can already be included in the next release

next steps are properly detecting real properties of the temporal dimension

clausmichele:
@soxofaan @VictorVerhaert thhis method parses already correctly the dimensions, why not re using that? It seems we are trying to re-implement something that already exists: https://github.com/Open-EO/openeo-python-client/blob/3c6045b6f0215455609599a868340cc52cddf86c/openeo/metadata.py#L413

clausmichele:
[...]

PR #568, opened 2024-06-05
 fix for broken `reduce_temporal` after `load_stac`

soxofaan:
Issue #567 initial fix for broken `reduce_temporal` after `load_stac`


soxofaan:
todo:

- changelog entry
- detect real temporal dimension from stac metadata
- tests with real-world-like use case

soxofaan:
Included in master with 2574094b1b2e6c43de9cdbaad723af259b49b553

Issue #566, opened 2024-06-03
Add method to load the results of a previous job


VictorVerhaert:
Add a method to the openeo python client to directly load the result of a previous job. The python client in turn fetches the stac of the previous job and add a `load_stac` with this link to the PG.

The method should receive either a `Job` object of a job-id as argument.

soxofaan:
suggestion: use the method name `load_stac_from_job`, which is self-explaining a easy to discover (e.g will automatically be suggested by autocomplete functionality if you start typing "load_stac")

PR #572, opened 2024-06-11
Issue566 load previous job

VictorVerhaert:


soxofaan:
issue #566

openeo/rest/connection.py
```diff
@@ -1385,6 +1385,56 @@ def load_stac(
...
+        spatial_extent: Optional[Dict[str, float]] = None,
+        temporal_extent: Union[Sequence[InputDate], Parameter, str, None] = None,
+        bands: Optional[List[str]] = None,
+        properties: Optional[Dict[str, Union[str, PGNode, Callable]]] = None,
+    ) -> DataCube:
+        """
+        Wrapper for :py:meth:`load_stac` that loads the result of a previous job using the STAC collection of its results.
+
+        :param job: The batch job object that has successfully finished, or a job id. When a job id is provided,
+            a job object is created using self as connection.
```
soxofaan, line 1400: "using self as connection" won't make sense to the general user.

I would phrase it something like

```
:param job: a :py:class:`~openeo.rest.job.BatchJob` or job id pointing to a finished job. 
    Note that the  :py:class:`~openeo.rest.job.BatchJob` approach allows to point 
    to a batch job on a different back-end.
```

openeo/rest/connection.py
```diff
@@ -1385,6 +1385,56 @@ def load_stac(
...
+        :param spatial_extent: limit data to specified bounding box or polygons
+        :param temporal_extent: limit data to specified temporal interval.
+        :param bands: only add the specified bands
```
soxofaan, line 1403:
```suggestion
        :param bands: limit data to the specified bands
```

openeo/rest/connection.py
```diff
@@ -1385,6 +1385,56 @@ def load_stac(
...
+        :return: a :py:class:`DataCube`
```
soxofaan, line 1404: Only add a `:return:` doc when you document more than just the return type. 
The return type is already covered by the type annotation, and rendered in the docs, so a `:return:` line like this gives duplication there

openeo/rest/connection.py
```diff
@@ -1385,6 +1385,56 @@ def load_stac(
...
+            if len(canonical_links) == 0:
+                _log.warning("No canonical link found in job results metadata. Using job results URL instead.")
+                stac_link = job.get_results_metadata_url(full=True)
+            else:
+                if len(canonical_links) > 1:
+                    _log.warning(
+                        f"Multiple canonical links found in job results metadata: {canonical_links}. Picking first one."
+                    )
+                stac_link = canonical_links[0]
+        except OpenEoApiError as e:
```
soxofaan, line 1426: I think you should also catch exceptions from failing `link["href"]`  or `link["rel"]` attempts above

tests/rest/test_connection.py
```diff
@@ -2456,6 +2456,88 @@ def test_properties(self, con120):
             }
         }
 
+    def test_load_stac_from_job_canonical(self, con120, requests_mock):
+        requests_mock.get(
+            API_URL + "jobs/j0bi6/results",
+            json={
+                "links": [
+                    {
+                        "href": "https://wrong.link",
```
soxofaan, line 2465: FYI: .link is a valid top-level domain. In tests we preferably use TLDs that are reserved for testing (and will never be valid) e.g. [.test](https://en.wikipedia.org/wiki/.test)

tests/rest/test_connection.py
```diff
@@ -2456,6 +2456,88 @@ def test_properties(self, con120):
...
+        job = con120.job("j0bi6")
+        unsigned_link = job.get_results_metadata_url(full=True)
+
+        cube = con120.load_stac_from_job(job)
+        fg = cube.flat_graph()
+
+        assert fg == {
+            "loadstac1": {
+                "process_id": "load_stac",
+                "arguments": {"url": unsigned_link},
```
soxofaan, line 2508: I think your expected value should not depend on get_results_metadata_url here, but just explicitly state `"https://oeo.test/jobs/j0bi6/results"`

soxofaan (commented):
looks good, just some minor notes

tests/rest/test_connection.py
bossie: > TLDs that are reserved for testing (and will never be valid) e.g. [.test](https://en.wikipedia.org/wiki/.test)

I did not know this :open_mouth:

openeo/rest/connection.py
```diff
@@ -1385,6 +1385,56 @@ def load_stac(
             _log.warning(f"Failed to extract cube metadata from STAC URL {url}", exc_info=True)
         return cube
 
+    def load_stac_from_job(
+        self,
+        job: Union[BatchJob, str],
+        spatial_extent: Optional[Dict[str, float]] = None,
+        temporal_extent: Union[Sequence[InputDate], Parameter, str, None] = None,
```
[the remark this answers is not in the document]
VictorVerhaert: Added it, and as a consequence also to load_stac.

openeo/rest/connection.py
VictorVerhaert: When these fail, `canonical_links` will just me empty and the unsigned link will be used as well. I've added an extra test to cover this behaviour

soxofaan: No I mean that the link dictionaries could be missing "rel" or "href" fields. Or maybe the link list items might not even dictionaries. Or the whole "links" field might be missing

It would actually be safer to write it like this:
```python
canonical_links = [
    link["href"] 
    for link in job_results.get_metadata().get("links", []) 
    if link.get("rel") == "canonical" and "href" in link
]
```

openeo/rest/connection.py
```diff
@@ -1385,6 +1385,58 @@ def load_stac(
...
+    ) -> DataCube:
+        """
+        Wrapper for :py:meth:`load_stac` that loads the result of a previous job using the STAC collection of its results.
+
+        :param job: a :py:class:`~openeo.rest.job.BatchJob` or job id pointing to a finished job.
+            Note that the  :py:class:`~openeo.rest.job.BatchJob` approach allows to point
+            to a batch job on a different back-end.
+        :param spatial_extent: limit data to specified bounding box or polygons
+        :param temporal_extent: limit data to specified temporal interval.
+        :param bands: limit data to the specified bands
```
soxofaan, line 1404: can you also add the following version info to this doc block:

```
.. versionadded:: 0.30.0
```

PR #576, opened 2024-06-17
inference using URL

Pratichhya:
Updated the documentation to use URL based method to load the model

### DIFF
--- a/docs/machine_learning.rst
+++ b/docs/machine_learning.rst
@@ -89,7 +89,6 @@ Finally execute this whole training flow as a batch job::
     training_job = model.create_job()
     training_job.start_and_wait()
 
-
 Inference
 ----------
 
@@ -97,22 +96,29 @@ When the batch job finishes successfully, the trained model can then be used
 with the ``predict_random_forest`` process on the raster data cube
 (or another cube with the same band structure) to classify all the pixels.
 
+Thus, the training job results is inspected to fetch the trained models url::
+
+    from openeo.rest.job import JobResults
+
+    results : JobResults = training_job.get_results()
+    links = results.get_metadata()['links']
+    ml_model_metadata_url = [link for link in links if 'ml_model_metadata.json' in link['href']][0]['href']
+    print(ml_model_metadata_url)
+
+
+Next, load the model from the URL::
+
+    from openeo.rest.mlmodel import MlModel
+    saved_model = MlModel.load_ml_model(connection=connection, id=ml_model_metadata_url)
+
 Technically, the openEO ``predict_random_forest`` process has to be used as a reducer function
 inside a ``reduce_dimension`` call, but the openEO Python client library makes it
 a bit easier by providing a :py:meth:`~openeo.rest.datacube.DataCube.predict_random_forest` method
 directly on the :py:class:`~openeo.rest.datacube.DataCube` class, so that you can just do::
 
     predicted = cube.predict_random_forest(
-        model=training_job.job_id,
+        model = saved_model,
         dimension="bands"
     )
 
     predicted.download("predicted.GTiff")
-
-
-We specified the model here by batch job id (string),
-but it can also be specified in other ways:
-as :py:class:`~openeo.rest.job.BatchJob` instance,
-as URL to the corresponding STAC Item that implements the `ml-model` extension,
-or as :py:class:`~openeo.rest.mlmodel.MlModel` instance (e.g. loaded through
-:py:meth:`~openeo.rest.connection.Connection.load_ml_model`).


docs/machine_learning.rst
```diff
@@ -89,30 +89,36 @@ Finally execute this whole training flow as a batch job::
...
 Inference
 ----------
 
 When the batch job finishes successfully, the trained model can then be used
 with the ``predict_random_forest`` process on the raster data cube
 (or another cube with the same band structure) to classify all the pixels.
 
+Thus, the training job results is inspected to fetch the trained models url::
+
+    from openeo.rest.job import JobResults
```
soxofaan, line 101:
```suggestion
```
Importing this for just a type hint in static docs is not that useful

docs/machine_learning.rst
```diff
@@ -89,30 +89,36 @@ Finally execute this whole training flow as a batch job::
...
+
+    results : JobResults = training_job.get_results()
```
soxofaan, line 103:
```suggestion
    results = training_job.get_results()
```

docs/machine_learning.rst
```diff
@@ -89,30 +89,36 @@ Finally execute this whole training flow as a batch job::
...
+    links = results.get_metadata()['links']
+    ml_model_metadata_url = [link for link in links if 'ml_model_metadata.json' in link['href']][0]['href']
```
soxofaan, line 105: Why getting the model URL from the links? Isn't this an asset in the batch job results?

Also, this "ml_model_metadata.json" looks like a VITO-specific implementation detail. Can this be generalized. Or at least documented why you are looking for that string?

docs/machine_learning.rst
```diff
@@ -89,30 +89,36 @@ Finally execute this whole training flow as a batch job::
...
+    print(ml_model_metadata_url)
+
+
+Next, load the model from the URL::
+
+    from openeo.rest.mlmodel import MlModel
+    saved_model = MlModel.load_ml_model(connection=connection, id=ml_model_metadata_url)
```
soxofaan, line 112: Isn't it possible to do this with something like

```python
saved_model = connection.load_ml_model(ml_model_metadata_url)
```

docs/machine_learning.rst
```diff
@@ -89,30 +89,36 @@ Finally execute this whole training flow as a batch job::
...
+
 Technically, the openEO ``predict_random_forest`` process has to be used as a reducer function
 inside a ``reduce_dimension`` call, but the openEO Python client library makes it
 a bit easier by providing a :py:meth:`~openeo.rest.datacube.DataCube.predict_random_forest` method
 directly on the :py:class:`~openeo.rest.datacube.DataCube` class, so that you can just do::
 
     predicted = cube.predict_random_forest(
-        model=training_job.job_id,
+        model = saved_model,
```
soxofaan, line 120:
```suggestion
        model=saved_model,
```

docs/machine_learning.rst
```diff
@@ -89,30 +89,36 @@ Finally execute this whole training flow as a batch job::
...
 
     predicted.download("predicted.GTiff")
-
-
-We specified the model here by batch job id (string),
-but it can also be specified in other ways:
-as :py:class:`~openeo.rest.job.BatchJob` instance,
-as URL to the corresponding STAC Item that implements the `ml-model` extension,
-or as :py:class:`~openeo.rest.mlmodel.MlModel` instance (e.g. loaded through
-:py:meth:`~openeo.rest.connection.Connection.load_ml_model`).
```
soxofaan, line 118: Why are the docs about the other options (batch job id, batch job instance, ...) removed? Is it because this is not valid anymore? If so, then the docs in a couple of places should be updated too

docs/machine_learning.rst
Pratichhya: This is within the metadata indeed but different than the Assets's link. The link in the assets is as tar.gz . It throws a JSON error when trying to load it using that specific url and same is with the canonical link. 
ref: https://github.com/Open-EO/openeo-geopyspark-driver/issues/562

docs/machine_learning.rst
Pratichhya: yes it is indeed.

docs/machine_learning.rst
Pratichhya: It was indeed confusion on my part since the other methods, weren't working for me in the CDSE backends.

docs/machine_learning.rst
soxofaan: If I understand  `save_ml_model` correctly (https://github.com/Open-EO/openeo-processes/pull/441/files#diff-0aae18f348c05c0f26c5caee2dcb0ca4af218a55287c9cbf8a286fc35b8faae5) there should be a asset that is a STAC Item that is to be used with `load_ml_model`.

that `'ml_model_metadata.json'` reference in the current docs is highly implementation specific and should be avoided

docs/machine_learning.rst
```diff
@@ -89,30 +89,42 @@ Finally execute this whole training flow as a batch job::
...
 
-
 Inference
 ----------
 
 When the batch job finishes successfully, the trained model can then be used
 with the ``predict_random_forest`` process on the raster data cube
 (or another cube with the same band structure) to classify all the pixels.
 
+Thus, the training job results is inspected to fetch the trained models url::
```
soxofaan, line 99:
```suggestion
We inspect the result metadata of the training job to obtain the STAC Item URL of the trained model::
```

docs/machine_learning.rst
```diff
@@ -89,30 +89,42 @@ Finally execute this whole training flow as a batch job::
...
+    model = connection.load_ml_model(id=ml_model_metadata_url)
+
 Technically, the openEO ``predict_random_forest`` process has to be used as a reducer function
 inside a ``reduce_dimension`` call, but the openEO Python client library makes it
 a bit easier by providing a :py:meth:`~openeo.rest.datacube.DataCube.predict_random_forest` method
 directly on the :py:class:`~openeo.rest.datacube.DataCube` class, so that you can just do::
 
     predicted = cube.predict_random_forest(
-        model=training_job.job_id,
+        model=saved_model,
```
soxofaan, line 118:
```suggestion
        model=model,
```

docs/machine_learning.rst
```diff
@@ -89,30 +89,42 @@ Finally execute this whole training flow as a batch job::
...
     predicted.download("predicted.GTiff")
 
-
-We specified the model here by batch job id (string),
+We specified the model here by URL corresponding to the
+STAC Item that implements the ml-model extension,
 but it can also be specified in other ways:
 as :py:class:`~openeo.rest.job.BatchJob` instance,
-as URL to the corresponding STAC Item that implements the `ml-model` extension,
+as job_id of training job(string),
```
soxofaan, line 128:
```suggestion
as job_id of training job (string),
```

docs/machine_learning.rst
soxofaan: I guess we are missing some specification to allow proper extraction of the model's URL, so I started the discussion at https://github.com/Open-EO/openeo-processes/pull/441/files/b162040dea04acd12fad73166274bbff370ed29b#r1650820880

Pratichhya:
@soxofaan, are additional changes needed ?

docs/machine_learning.rst
```diff
@@ -89,30 +89,42 @@ Finally execute this whole training flow as a batch job::
...
+
+    results = training_job.get_results()
+    links = results.get_metadata()['links']
+    ml_model_metadata_url = [link for link in links if 'ml_model_metadata.json' in link['href']][0]['href']
+    print(ml_model_metadata_url)
+
+
+Next, load the model from the URL::
+
+    model = connection.load_ml_model(id=ml_model_metadata_url)
```
soxofaan, line 110: just do
```suggestion
    model = connection.load_ml_model(ml_model_metadata_url)
```

I think the `id` argument is not a very good name and we might change that in the future, so just avoid using it to keep these docs more robust

docs/machine_learning.rst
```diff
@@ -89,30 +89,42 @@ Finally execute this whole training flow as a batch job::
...
-        model=training_job.job_id,
+        model=model,
         dimension="bands"
     )
 
     predicted.download("predicted.GTiff")
 
-
-We specified the model here by batch job id (string),
+We specified the model here by URL corresponding to the
```
soxofaan, line 124: This is not correct, you specify the model with `model`, which is a `MlModel` instance, not a URL

docs/machine_learning.rst
soxofaan: We need a better solution for 
```
[link for link in links if 'ml_model_metadata.json' in link['href']][0]['href']
```

this construct is quite brittle and implementation specific. 
Also note that https://github.com/Open-EO/openeo-processes/pull/441/files/b162040dea04acd12fad73166274bbff370ed29b#r1650872242 explicitly says that this the ML Model extension is outdated and going to be replaced, so it is very likely this is going to break

soxofaan: This is apparently the next-gen ML extension
- https://github.com/crim-ca/mlm-extension

I think we better make sure we make this snippet future-proof

docs/machine_learning.rst
soxofaan: ref:
- https://github.com/Open-EO/openeo-python-driver/issues/306

Issue #571, opened 2024-06-11
JobManager: store joblist on http url as Parquet


VincentVerelst:
https://github.com/Open-EO/openeo-gfmap/issues/107

- [x] Support Parquet
- [x] Support URL

VincentVerelst:
@soxofaan , I've made a gist on how I'd implement URL support in the MultiBackendJobManager: https://gist.github.com/VincentVerelst/aef96f03b5b45a3506d8945730af8221
It would require a user to always provide a url_read_write_helper if they provide a URL. We could foresee maybe some default helpers for CSV and Parquet on Artifactory?

Could you already have a look and leave your comments, before I start the full implementation?

soxofaan:
As a next step I was more thinking like PR #595

PR #580, opened 2024-06-24
Job tracker storage 

VincentVerelst:
@soxofaan , for issue #571 I created a JobTrackerStorage to manage everything job tracker related in the MultiBackendJobManager. For now it only supports storing as CSV to PosixPath, so no extra functionality has been added. Would be great to have your feedback already on this before adding functionality for storing as Parquet and storing to URL as well.

### DIFF
--- a/openeo/extra/job_management.py
+++ b/openeo/extra/job_management.py
@@ -175,35 +175,7 @@ class MultiBackendJobManager:
         connection.session.mount("https://", HTTPAdapter(max_retries=retries))
         connection.session.mount("http://", HTTPAdapter(max_retries=retries))
 
-    def _normalize_df(self, df: pd.DataFrame) -> pd.DataFrame:
-        """Ensure we have the required columns and the expected type for the geometry column.
 
-        :param df: The dataframe to normalize.
-        :return: a new dataframe that is normalized.
-        """
-
-        # check for some required columns.
-        required_with_default = [
-            ("status", "not_started"),
-            ("id", None),
-            ("start_time", None),
-            ("cpu", None),
-            ("memory", None),
-            ("duration", None),
-            ("backend_name", None),
-        ]
-        new_columns = {
-            col: val for (col, val) in required_with_default if col not in df.columns
-        }
-        df = df.assign(**new_columns)
-        # Workaround for loading of geopandas "geometry" column.
-        if "geometry" in df.columns and df["geometry"].dtype.name != "geometry":
-            df["geometry"] = df["geometry"].apply(shapely.wkt.loads)
-        return df
-
-    def _persists(self, df, output_file):
-        df.to_csv(output_file, index=False)
-        _log.info(f"Wrote job metadata to {output_file.absolute()}")
 
     def run_jobs(
         self,
@@ -248,17 +220,9 @@ class MultiBackendJobManager:
         """
         # TODO: Defining start_jobs as a Protocol might make its usage more clear, and avoid complicated doctrings,
         #   but Protocols are only supported in Python 3.8 and higher.
-        # TODO: this resume functionality better fits outside of this function
-        #       (e.g. if `output_file` exists: `df` is fully discarded)
-        output_file = Path(output_file)
-        if output_file.exists() and output_file.is_file():
-            # Resume from existing CSV
-            _log.info(f"Resuming `run_jobs` from {output_file.absolute()}")
-            df = pd.read_csv(output_file)
-            status_histogram = df.groupby("status").size().to_dict()
-            _log.info(f"Status histogram: {status_histogram}")
 
-        df = self._normalize_df(df)
+        df = JobTrackerStorage().resume_df(df, output_file)
+        df = JobTrackerStorage().normalize_df(df)
 
         while (
             df[
@@ -273,7 +237,7 @@ class MultiBackendJobManager:
                 self._update_statuses(df)
             status_histogram = df.groupby("status").size().to_dict()
             _log.info(f"Status histogram: {status_histogram}")
-            self._persists(df, output_file)
+            JobTrackerStorage().persists(df, output_file)
 
             if len(df[df.status == "not_started"]) > 0:
                 # Check number of jobs running at each backend
@@ -293,7 +257,7 @@ class MultiBackendJobManager:
                         to_launch = df[df.status == "not_started"].iloc[0:to_add]
                         for i in to_launch.index:
                             self._launch_job(start_job, df, i, backend_name)
-                            self._persists(df, output_file)
+                            JobTrackerStorage().persists(df, output_file)
 
             time.sleep(self.poll_sleep)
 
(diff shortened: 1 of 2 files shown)


openeo/extra/job_management.py
```diff
@@ -458,3 +422,60 @@ def ignore_connection_errors(context: Optional[str] = None):
...
+    # TODO: add support to store as Parquet file as well as CSV.
+    # TODO: add support to store to both PosixPath and URL.
+    """
+    Helper to manage the storage of batch job metadata.
+    """
+
+    def __init__(self):
+        pass
+
+    def normalize_df(self, df: pd.DataFrame) -> pd.DataFrame:
```
soxofaan, line 437: I'm not sure the dataframe normalization is an aspect of the storage. It's not that a different storage class will require different columns/defaults. I think the normalization can stay part of the job manager

openeo/extra/job_management.py
```diff
@@ -458,3 +422,60 @@ def ignore_connection_errors(context: Optional[str] = None):
...
+            ("cpu", None),
+            ("memory", None),
+            ("duration", None),
+            ("backend_name", None),
+        ]
+        new_columns = {col: val for (col, val) in required_with_default if col not in df.columns}
+        df = df.assign(**new_columns)
+
+        return df
+    def resume_df(self, df: pd.DataFrame, output_file: Union[str, Path]) -> pd.DataFrame:
```
soxofaan, line 458: I think the `output_file` should go in the `__init__`. Otherwise you can't work with alternative storage implementations that don't work with files

openeo/extra/job_management.py
```diff
@@ -458,3 +422,60 @@ def ignore_connection_errors(context: Optional[str] = None):
...
+        output_file = Path(output_file)
+        if output_file.exists() and output_file.is_file():
+            # Resume from existing CSV
+            _log.info(f"Resuming `run_jobs` from {output_file.absolute()}")
+            df = pd.read_csv(output_file)
+            status_histogram = df.groupby("status").size().to_dict()
+            _log.info(f"Status histogram: {status_histogram}")
+        return df
+
+    def persists(self, df, output_file):
```
soxofaan, line 474: Same as above: `output_file` should not be an argument here

openeo/extra/job_management.py
```diff
@@ -458,3 +422,60 @@ def ignore_connection_errors(context: Optional[str] = None):
...
+
+
+class JobTrackerStorage:
+    # TODO: add support to store as Parquet file as well as CSV.
+    # TODO: add support to store to both PosixPath and URL.
+    """
+    Helper to manage the storage of batch job metadata.
+    """
+
+    def __init__(self):
```
soxofaan, line 434: As noted below/above,  `__init__` should take the target file name/path I think

openeo/extra/job_management.py
```diff
@@ -458,3 +422,60 @@ def ignore_connection_errors(context: Optional[str] = None):
         _log.warning(f"Ignoring connection error (context {context or 'n/a'}): {e}")
         # Back off a bit
         time.sleep(5)
+
+
+class JobTrackerStorage:
```
soxofaan, line 427: I don't have an alternative yet, but it might be risky to call this "tracker", which is already a thing at the (VITO) backend side

soxofaan (commented):
some quick notes

jdries:
Abstracting it in a separate class is the way to go.
I am however wondering if we should keep the abstraction where the full dataframe is always loaded, or of the 'storage tracker' can offer more granular functionality?

For instance methods like (please improve my bad naming):

get_jobs_by_status(max_jobs, status:List[String]) (returns jobs for given status list)
update_job_metadata(job_row) (writes updated job state back to the database)
commit() (persist modifications to file or database)

soxofaan:
> I am however wondering if we should keep the abstraction where the full dataframe is always loaded,

Indeed, but that involves a bigger refactor (or maybe even full rewrite) of the multi job manager I guess.
In this first PR we want to make a first step by abstracting out just the storage aspect, without fiddling too much with the rest

jdries:
@soxofaan  yes, fine to keep this one limited and then move on to the bigger story.
I mainly wanted to inform you to avoid that we put a lot of work in polishing this one, to only then replace it with something else afterwards.

openeo/extra/job_management.py
VincentVerelst: @soxofaan, JobDatabasePersistence? In line with Jeroen's APEx documentation: https://jdries-vito.quarto.pub/apex-design/upscaling.html#job-database-persistence

VincentVerelst:
@soxofaan , as discussed earlier, I've abstracted away the read/write helper for CSV and Parquet files and allowed the user to give a custom read/write helper (with the eye on adding support for URL's for which a user might need to provide a custom request method). 
An interim review from you would be helpful, before continuing to URL support.

openeo/extra/job_management.py
```diff
@@ -192,24 +192,17 @@ def _normalize_df(self, df: pd.DataFrame) -> pd.DataFrame:
...
             ("backend_name", None),
         ]
-        new_columns = {
-            col: val for (col, val) in required_with_default if col not in df.columns
-        }
+        new_columns = {col: val for (col, val) in required_with_default if col not in df.columns}
         df = df.assign(**new_columns)
-        # Workaround for loading of geopandas "geometry" column.
-        if "geometry" in df.columns and df["geometry"].dtype.name != "geometry":
-            df["geometry"] = df["geometry"].apply(shapely.wkt.loads)
```
soxofaan, line 201: This workaround seems to be removed by this PR. Is that intended?
I think it's still necessary for at least the CSV use case?

openeo/extra/job_management.py
```diff
@@ -192,24 +192,17 @@ def _normalize_df(self, df: pd.DataFrame) -> pd.DataFrame:
...
-        df.to_csv(output_file, index=False)
-        _log.info(f"Wrote job metadata to {output_file.absolute()}")
+        return df
 
     def run_jobs(
         self,
         df: pd.DataFrame,
         start_job: Callable[[], BatchJob],
         output_file: Union[str, Path],
+        read_write_helper=None,
```
soxofaan, line 205: In this first PR I would not provide this `read_write_helper` yet.
Let's first make sure that we get parquet support right, before committing to a new user facing API

openeo/extra/job_management.py
```diff
@@ -458,3 +447,86 @@ def ignore_connection_errors(context: Optional[str] = None):
...
+class _ParquetReadWrite:
+    """
+    Helper to read and write Parquet files with pandas.
+    """
+
+    def read_file(self, file: Path) -> pd.DataFrame:
+        return pd.read_parquet(file)
+
+    def write_file(self, df: pd.DataFrame, file: Path):
+        df.to_parquet(file, index=False)
```
soxofaan, line 532: How I was thinking to evolve this as discussed, with minimal changes to just introduce parquet support, but still open the door to evolve this further for more persistence formats

```python

    def run_jobs(
        self,
        df: pd.DataFrame,
        start_job: Callable[[], BatchJob],
        output_file: Union[str, Path],
    ):
        output_file = Path(output_file)

        if output_file.suffix.lower() == ".csv":
            job_db = _CsvJobDatabase(output_file)
        elif output_file.suffix.lower() == ".parquet":
            job_db = _ParquetJobDatabase(output_file)
        else:
            raise ValueError(f"Unsupported output file type: {output_file}")

        if output_file.exists() and output_file.is_file():
            # Resume from existing db
            _log.info(f"Resuming `run_jobs` from {output_file.absolute()}")
            df = job_db.read()
            status_histogram = df.groupby("status").size().to_dict()
            _log.info(f"Status histogram: {status_histogram}")

        df = self._normalize_df(df)
        ...

        while ...
            ...
            job_db.persist(df)

...


class _JobDatabaseInterface:
    def read(self) -> pd.DataFrame:
        raise NotImplementedError()

    def persist(self, df: pd.DataFrame):
        raise NotImplementedError()

class _CsvJobDatabase(_JobDatabaseInterface):
    def __init__(self, path: Union[str, Path]):
        self.path = Path(path)

    def read(self) -> pd.DataFrame:
        return pd.read_csv(self.path)

    def persist(self, df: pd.DataFrame):
        self.path.parent.mkdir(parents=True, exist_ok=True)
        df.to_csv(self.path)

class _ParquetJobDatabase(_JobDatabaseInterface):
    def __init__(self, path: Union[str, Path]):
        self.path = Path(path)

    def read(self) -> pd.DataFrame:
        return pd.read_parquet(self.path)

    def persist(self, df: pd.DataFrame):
        self.path.parent.mkdir(parents=True, exist_ok=True)
        df.to_parquet(self.path)

File: openeo/extra/job_management.py (hunk 6 above)
Comment (VincentVerelst) [reply openeo/extra/job_management.py:201]: It's indeed intended, I want to leave all data in the geometry column as a string. Otherwise we run into trouble when writing the pandas DF to Parquet. Is there a reason why it might be necessary for CSV? Leaving it as a string also seems to work fine.

Comment (soxofaan) [reply openeo/extra/job_management.py:201]: But is removing this going to break some use cases where users expect a real geometry in their `start_job` callable?

File: tests/extra/test_job_management.py [hunk 9]
```diff
@@ -16,7 +16,12 @@
 
 import openeo
 from openeo import BatchJob
-from openeo.extra.job_management import MAX_RETRIES, MultiBackendJobManager
+from openeo.extra.job_management import (
+    MAX_RETRIES,
+    MultiBackendJobManager,
+    _CsvJobDatabase,
+    _ParquetJobDatabase,
```
soxofaan, line 23: you import these _CsvJobDatabase and _ParquetJobDatabase but don't use them yet?

openeo/extra/job_management.py
VincentVerelst: Hmm yes probably. So I can add it to the read method of CsvJobDatabase, but not to the ParquetJobDatabase, or what do you propose?

soxofaan: indeed, I would reinstate it at least for the CSV use case.

With parquet, the geometry is maybe loaded correctly automatically?

VincentVerelst: Both for CSV and Parquet, pandas will parse the geometry as string. However, for writing, it's possible to write a real geometry to CSV, but not to parquet (you'd have to use geopandas instead of pandas)

soxofaan (commented):
just some minor things

PR #587, opened 2024-07-15
Add testing utilities

soxofaan:
this PR adds testing utilities under `openeo.testing`, in particular `openeo.testing.results` to easily have assert functions to compare job results against reference data

relate to ESA-APEx/apex_algorithms#5


openeo/internal/compat.py
soxofaan, line 1: do removal of this compat.py utility separately on master

soxofaan:
TODO: include in docs and clearly flag as experimental API for now

soxofaan:
- [ ] add changelog entry

PR #591, opened 2024-07-17
Issue567 stac metadata temporal dim

clausmichele:
Very simple fix to https://github.com/Open-EO/openeo-python-client/issues/567, which allows doing:

```python
local_conn = LocalConnection("./")

url = "https://stac.eurac.edu/collections/SENTINEL2_L2A_SAMPLE_2"
spatial_extent =  {"east": 11.40, "north": 46.52, "south": 46.46, "west": 11.25}
temporal_extent = ["2022-06-01", "2022-06-30"]
bands = ["B04"]
s2_cube = local_conn.load_stac(
    url=url,
    spatial_extent=spatial_extent,
    temporal_extent=temporal_extent,
    bands=bands,
)
s2_cube = s2_cube.reduce_dimension(dimension="time",reducer="mean")
s2_cube
```

which otherwise would end up with:

`ValueError: Invalid dimension 'time'. Should be one of ['bands', 't']`


openeo/metadata.py
```diff
@@ -540,6 +540,22 @@ def metadata_from_stac(url: str) -> CubeMetadata:
...
+            or {}
+        )
+        if not cube_dimensions:
+            complain("No cube:dimensions metadata")
+        dimensions = []
+        for name, info in cube_dimensions.items():
+            dim_type = info.get("type")
+            if dim_type == "temporal":
+                return TemporalDimension(name=name, extent=info.get("extent"))
+        return []
```
soxofaan, line 557: returning an empty list as fallback looks weird. Shouldn't this be return `None` instead?

openeo/metadata.py
```diff
@@ -540,6 +540,22 @@ def metadata_from_stac(url: str) -> CubeMetadata:
...
+    def get_temporal_metadata(spec: dict, complain: Callable[[str], None] = warnings.warn) -> TemporalDimension:
+        # Dimension info is in `cube:dimensions` (or 0.4-style `properties/cube:dimensions`)
+        cube_dimensions = (
+            deep_get(spec, "cube:dimensions", default=None)
+            or deep_get(spec, "properties", "cube:dimensions", default=None)
+            or {}
+        )
+        if not cube_dimensions:
+            complain("No cube:dimensions metadata")
+        dimensions = []
```
soxofaan, line 552: this variable `dimensions` is not used, right?

openeo/metadata.py
```diff
@@ -553,8 +569,8 @@ def get_band_names(bands: List[Band]) -> List[str]:
     def is_band_asset(asset: pystac.Asset) -> bool:
         return "eo:bands" in asset.extra_fields
 
+    temporal_dimension = []
```
soxofaan, line 572: maybe I'm misunderstanding something, but using an empty list as default value for a dimension object looks weird

openeo/metadata.py
```diff
@@ -540,6 +540,21 @@ def metadata_from_stac(url: str) -> CubeMetadata:
 
     # TODO move these nested functions and other logic to _StacMetadataParser
 
+    def get_temporal_metadata(spec: dict, complain: Callable[[str], None] = warnings.warn) -> TemporalDimension:
```
soxofaan, line 543: as noted, I would not define this helper to work on dictionaries, but with pystac objects directly, to leverage all its existing  validation and discovery functionality

openeo/metadata.py
```diff
@@ -582,6 +597,7 @@ def is_band_asset(asset: pystac.Asset) -> bool:
         if _PYSTAC_1_9_EXTENSION_INTERFACE and collection.ext.has("item_assets"):
             # TODO #575 support unordered band names and avoid conversion to a list.
             bands = list(_StacMetadataParser().get_bands_from_item_assets(collection.ext.item_assets))
+        temporal_dimension = get_temporal_metadata(collection.to_dict())
```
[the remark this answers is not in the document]
soxofaan: moreover this PR should also cover the `pystac.Item` above I think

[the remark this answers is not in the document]
soxofaan: And it should be pretty straightforward to add some test coverage, e.g. to existing tests at https://github.com/Open-EO/openeo-python-client/blob/7e297f6bc13db493aa5dcbe1c94c23d28cf83038/tests/test_metadata.py#L791-L843

clausmichele:
@soxofaan the tests are passing, it's a bit more cumbersome to keep the support for the old Python versions though.

clausmichele:
I would have some more updates to parse correctly the spatial and bands dimension names, should I include them here @soxofaan ?

soxofaan:
I'll do some quick tweaks to this PR and try to merge asap

soxofaan:
finetuned and merged in 35b5fc7b0cc844bf5f02b0e437f683f501ffe5ff

soxofaan:
thanks!

PR #595, opened 2024-07-19
Start supporting custom job databases in MultiBackendJobManager

soxofaan:
Another step for issue #571 : allow user to specify a custom job database in run_jobs

### DIFF
--- a/openeo/extra/job_management.py
+++ b/openeo/extra/job_management.py
@@ -38,6 +38,8 @@ class JobDatabaseInterface(metaclass=abc.ABCMeta):
     Interface for a database of job metadata to use with the :py:class:`MultiBackendJobManager`,
     allowing to regularly persist the job metadata while polling the job statuses
     and resume/restart the job tracking after it was interrupted.
+
+    .. versionadded:: 0.31.0
     """
 
     @abc.abstractmethod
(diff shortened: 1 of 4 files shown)


openeo/extra/job_management.py
```diff
@@ -30,6 +31,22 @@ class _Backend(NamedTuple):
...
+        """Is there existing data in the database?"""
+        return False
+
+    def read(self) -> pd.DataFrame:
+        """Read job data from the database ad pandas DataFrame."""
+        raise NotImplementedError()
+
+    def persist(self, df: pd.DataFrame):
+        """Store job data to the database."""
+        raise NotImplementedError()
```
VincentVerelst, line 47: Is there a reason not to make this a formal interface? As a user it might be beneficial to immediately get thrown an error if your custom class doesn't have all necessary methods.

soxofaan: Indeed, now that we are making this interface more public, that would be useful

openeo/extra/job_management.py
```diff
@@ -472,18 +512,14 @@ def ignore_connection_errors(context: Optional[str] = None):
...
-        raise NotImplementedError()
-
 
-class _CsvJobDatabase(_JobDatabaseInterface):
+class CsvJobDatabase(JobDatabaseInterface):
     def __init__(self, path: Union[str, Path]):
         self.path = Path(path)
 
+    def exists(self) -> bool:
+        return self.path.exists()
```
soxofaan, line 521: add `is_file()` too

openeo/extra/job_management.py
```diff
@@ -501,15 +537,19 @@ def read(self) -> pd.DataFrame:
...
         df.to_csv(self.path, index=False)
 
 
-class _ParquetJobDatabase(_JobDatabaseInterface):
+class ParquetJobDatabase(JobDatabaseInterface):
     def __init__(self, path: Union[str, Path]):
         self.path = Path(path)
 
+    def exists(self) -> bool:
+        return self.path.exists()
```
soxofaan, line 551: add `is_file()` too

openeo/extra/job_management.py
soxofaan: reconsidering this: if it exists and it's not a file, things should fail at "read" time (as early as possible), not at "persist" time (when valuable work might have been gone to waste). Keeping as-is mean former behavior, adding `is_file()` changes to latter behavior.

PR #596, opened 2024-07-19
Include cancelling long running jobs

HansVRP:
Included _check_and_stop_long_running_jobs.

Still needs to be covered in test_suite


openeo/extra/job_management.py
```diff
@@ -76,7 +76,7 @@ def start_job(
     """
 
     def __init__(
-        self, poll_sleep: int = 60, root_dir: Optional[Union[str, Path]] = "."
+        self, poll_sleep: int = 60, root_dir: Optional[Union[str, Path]] = ".", max_running_duration: Optional[int] = 720
```
soxofaan, line 79: I think in all other places we specify timeouts/delays in seconds, so it might be better for consistence to define `max_running_duration` in seconds too. Note that you can specify the default as `12 * 60 * 60` to make it more obvious that this is 12 hours

openeo/extra/job_management.py
```diff
@@ -202,6 +209,39 @@ def _normalize_df(self, df: pd.DataFrame) -> pd.DataFrame:
...
+
+            try:
+                job_id = df.loc[i, "id"]
+                backend_name = df.loc[i, "backend_name"]
+                con = self._get_connection(backend_name)
+                the_job = con.job(job_id)
+
+                job_time_created = the_job.describe()['created']
+                job_time_created = datetime.datetime.strptime(job_time_created, '%Y-%m-%dT%H:%M:%SZ')
+                time_difference = (datetime.datetime.now() - job_time_created)
```
soxofaan, line 229: to avoid timezone headaches, it's typically recommended to call now() as follows:
```
import timezone
...
         datetime.datetime.now(timezone.utc)
```


e.g see https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow

openeo/extra/job_management.py
```diff
@@ -202,6 +209,39 @@ def _normalize_df(self, df: pd.DataFrame) -> pd.DataFrame:
...
+
+                if (time_difference.total_seconds() > self.max_running_duration.total_seconds()):
```
soxofaan, line 231: I think you can simplify all this date math a bit by just doing something like

```
if datetime.datetime.now(timezone.utc) > job_time_created + self.max_running_duration:
```

openeo/extra/job_management.py
```diff
@@ -202,6 +209,39 @@ def _normalize_df(self, df: pd.DataFrame) -> pd.DataFrame:
         df = df.assign(**new_columns)
 
         return df
+    
+    def _check_and_stop_long_running_jobs(self, df: pd.DataFrame):
```
soxofaan, line 213: I'm wondering if it is necessary to define a separate method for this actually.

I think you can  do this cancelling in the existing for-loop of `_update_statuses`, instead of duplicating a bunch of stuff

soxofaan (commented):
some initial notes

openeo/extra/job_management.py
HansVRP: done

openeo/extra/job_management.py
HansVRP: done

openeo/extra/job_management.py
HansVRP: done

openeo/extra/job_management.py
soxofaan: I would do the canceling from within `_update_statuses`.

Note that `_update_statuses` is about more than updating statuses anyway: it triggers `on_job_done` and `on_job_error`, which might be non-trivial things. I think `_update_statuses` is a bit of misnomer at the moment

HansVRP: Sounds good, should I rename it to _update_jobs within this PR?

soxofaan: `_tracks_statuses`  makes more sense at the moment I guess

openeo/extra/job_management.py
```diff
@@ -437,15 +441,38 @@ def _update_statuses(self, df: pd.DataFrame):
...
                 job_metadata = the_job.describe()
-                _log.info(
-                    f"Status of job {job_id!r} (on backend {backend_name}) is {job_metadata['status']!r}"
-                )
+                _log.info(f"Status of job {job_id!r} (on backend {backend_name}) is {job_metadata['status']!r}")
+
+                if job_metadata["status"] == "running":
+
+                    timezone = datetime.timezone.utc
+                    job_time_created = datetime.datetime.strptime(the_job.describe()["created"], "%Y-%m-%dT%H:%M:%SZ")
```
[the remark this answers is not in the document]
HansVRP: as discussed, I do propose to move the logic of cancelling a job to a separate function.

openeo/extra/job_management.py
```diff
@@ -437,15 +441,38 @@ def _update_statuses(self, df: pd.DataFrame):
...
+                    job_time_created = job_time_created.replace(tzinfo=timezone)
+
+                    if datetime.datetime.now(timezone) > job_time_created + self.max_running_duration:
+                        try:
+                            _log.info(
+                                f"Cancelling job {job_id} on backend {backend_name} as it has been running for more than {self.max_running_duration} minutes"
+                            )
+                            the_job.stop_job()
+                            self.on_job_error(the_job, df.loc[i])
```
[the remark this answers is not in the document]
HansVRP: would this then be a replication of on_job_error except that you might want to log the info logs instead of the error logs? (I am adding the cancel logging to info)

openeo/extra/job_management.py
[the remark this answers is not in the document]
soxofaan: as discussed, it's unfortunate the openEO API does not include job start timestamp yet:
- https://github.com/Open-EO/openeo-api/issues/542

HansVRP:
@soxofaan I have worked on the comments above and also included 2 additional unit tests to test the behavior

openeo/extra/job_management.py
```diff
@@ -404,6 +415,32 @@ def on_job_error(self, job: BatchJob, row):
...
+
+    def _cancel_prolonged_job(self, job: BatchJob, row):
+        """Cancel the job if it has been running for too long."""
+        job_running_timestamp = rfc3339.parse_datetime(row["running_time"], with_timezone = True)
+        current_time = datetime.now(timezone.utc)
+        
+        if current_time > job_running_timestamp + self.max_running_duration:
+            try:
+                job.stop()
+                _log.info(
```
soxofaan, line 438: put the `_log.info()` before the `job.stop()`

openeo/extra/job_management.py
```diff
@@ -93,6 +98,10 @@ def __init__(
                 - get_job_dir
                 - get_error_log_path
                 - get_job_metadata_path
+
+        :param max_running_duration [seconds]:
```
soxofaan, line 102: About the name of this parameter: users will use it like:
```python
manager = MultiBackendJobManager(max_running_duration=1000)
```

This might give wrong impression that the whole manager will stop running after 1000 seconds.

proposal: name this parameter `cancel_job_after` or `cancel_running_job_after` (the latter leaves room to add `cancel_started_job_after` later if we want that)

soxofaan (commented):
some more quick notes :smile: 
(I haven't looked into tests yet)

openeo/extra/job_management.py
```diff
@@ -1,5 +1,5 @@
 import contextlib
-import datetime
+from datetime import datetime, timedelta, timezone
```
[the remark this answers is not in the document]
HansVRP: Fully agree

I made this revision since I was struggling with datetime for this exact reason in the unit tests. I wanted to mock the datetime.datetime module without mocking the datetime.timedelta module, and this was only possible when importing it as is. 

Maybe good to revisit this comment when you had had time to look at the tests as well.

openeo/extra/job_management.py
```diff
@@ -93,6 +98,10 @@ def __init__(
                 - get_job_dir
                 - get_error_log_path
                 - get_job_metadata_path
+
+        :param max_running_duration [seconds]:
+            A temporal limit for long running jobs to get automatically canceled.
+            The preset duration 12 hours. Can be set to None to disable
```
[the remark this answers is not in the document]
HansVRP: good catch

openeo/extra/job_management.py
HansVRP: Good suggestion

openeo/extra/job_management.py
```diff
@@ -404,6 +415,32 @@ def on_job_error(self, job: BatchJob, row):
...
 
+    def on_job_cancel(self, job: BatchJob, row):
+        """
+        Handles jobs that that were cancelled. Can be overridden to provide custom behaviour.
+
+        Default implementation does not do anything.
+
+        :param job: The job that has finished.
+        :param row: DataFrame row containing the job's metadata.
+        """
```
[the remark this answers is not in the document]
HansVRP: for some reason that statement was floating around elsewhere

openeo/extra/job_management.py
```diff
@@ -437,15 +471,25 @@ def _update_statuses(self, df: pd.DataFrame):
...
                 job_metadata = the_job.describe()
-                _log.info(
-                    f"Status of job {job_id!r} (on backend {backend_name}) is {job_metadata['status']!r}"
-                )
-                if job_metadata["status"] == "finished":
-                    self.on_job_done(the_job, df.loc[i])
+                _log.info(f"Status of job {job_id!r} (on backend {backend_name}) is {job_metadata['status']!r}")
+
+                if (df.loc[i, "status"] == "created" or df.loc[i, "status"] == "queued" or df.loc[i, "status"] == "started") and job_metadata["status"] == "running":
+                    df.loc[i, "running_time"] = datetime.now(timezone.utc).strftime('%Y-%m-%dT%H:%M:%S.%fZ')
```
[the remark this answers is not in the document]
HansVRP: I agree on the confusing. 

The idea was to remain consistent with the start_time field in pandas. Ideally I would go for start_running_timestamp, but I think this is confusing as well in relation to start_time.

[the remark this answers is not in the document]
HansVRP: maybe running_state_timestamp?

It could be usefull to revise start_time to a similar notation; however this is not backwards compatible

openeo/extra/job_management.py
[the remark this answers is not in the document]
soxofaan: regarding mocking datetime in tests, we started using the "time_machine" test utility, which takes care of all that hard-to-get-right date/time mocking stuff.
e.g. see https://github.com/Open-EO/openeo-python-client/blob/f7ee55223464bf67e45291b9df773d3749f4150f/tests/test_util.py#L297-L305

openeo/extra/job_management.py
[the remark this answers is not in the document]
soxofaan: I don't think it's necessary to include `status` (or `state`) in the name. I would also stick to `time`, like in existing `start_time`, instead of introducing `timestamp` 

So `start_running_time` or `running_start_time`. Maybe even `start_running` or `running_start` would work :smile:

tests/extra/test_job_management.py
```diff
@@ -455,7 +447,65 @@ def start_job(row, connection_provider, connection, **kwargs):
         assert set(result.status) == {"running"}
         assert set(result.backend_name) == {"foo"}
 
-
+    @mock.patch('openeo.extra.job_management.datetime', autospec=True)
+    @mock.patch('openeo.extra.job_management.timezone', autospec=True)
```
soxofaan, line 451: as mentioned higher, I would leverage the "time_machine" test helper/fixture instead of these convoluted datetime mocks

e.g. see pytest-specific docs at https://github.com/adamchainz/time-machine?tab=readme-ov-file#pytest-plugin

openeo/extra/job_management.py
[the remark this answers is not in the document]
HansVRP: sure, however I want to avoid using rfc3999.utcnow as it creates a string. For the comparisson I actually need a timedelta object. 

I will take a look if this is also an option through the rfc package

openeo/extra/job_management.py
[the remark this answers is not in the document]
HansVRP: by moving to rcf3339, I was able to avoid the datetime ambiguity

openeo/extra/job_management.py
```diff
@@ -189,14 +199,15 @@ def _normalize_df(self, df: pd.DataFrame) -> pd.DataFrame:
         required_with_default = [
             ("status", "not_started"),
             ("id", None),
-            ("start_time", None),
+            ("start_time", None),
```
soxofaan, line 202:
```suggestion
            ("start_time", None),
```
apparently you added a trailing space here

openeo/extra/job_management.py
```diff
@@ -189,14 +199,15 @@ def _normalize_df(self, df: pd.DataFrame) -> pd.DataFrame:
...
+            ("running_start_time", None), 
             # TODO: columns "cpu", "memory", "duration" are not referenced directly
             #       within MultiBackendJobManager making it confusing to claim they are required.
-            #       However, they are through assumptions about job "usage" metadata in `_update_statuses`.
+            #       However, they are through assumptions about job "usage" metadata in `_track_statuses`.
             ("cpu", None),
             ("memory", None),
             ("duration", None),
-            ("backend_name", None),
+            ("backend_name", None)
```
soxofaan, line 210:
```suggestion
            ("backend_name", None),
```
I would not remove this trailing comma (e.g. code formatters like [black](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#trailing-commas) will put it back anyway)

openeo/extra/job_management.py
```diff
@@ -404,6 +412,34 @@ def on_job_error(self, job: BatchJob, row):
             self.ensure_job_dir_exists(job.job_id)
             error_log_path.write_text(json.dumps(error_logs, indent=2))
 
+    def on_job_cancel(self, job: BatchJob, row):
+        """
+        Handles jobs that that were cancelled. Can be overridden to provide custom behaviour.
```
soxofaan, line 417:
```suggestion
        Handle a job that was cancelled. Can be overridden to provide custom behaviour.
```

openeo/extra/job_management.py
```diff
@@ -404,6 +412,34 @@ def on_job_error(self, job: BatchJob, row):
...
+
+        Default implementation does not do anything.
+
+        :param job: The job that has finished.
```
soxofaan, line 421:
```suggestion
        :param job: The job that was canceled.
```

openeo/extra/job_management.py
```diff
@@ -404,6 +412,34 @@ def on_job_error(self, job: BatchJob, row):
...
+
+    def _cancel_prolonged_job(self, job: BatchJob, row):
+        """Cancel the job if it has been running for too long."""
+        job_running_start_time = rfc3339.parse_datetime(row["running_start_time"], with_timezone=True)
+        current_time = rfc3339.parse_datetime(rfc3339.utcnow(), with_timezone=True)
+
+        
+        if current_time > job_running_start_time + self.cancel_running_job_after:
+            try:
+                print(str(self.cancel_running_job_after))
```
soxofaan, line 434: this print is probably a debug left-over

```suggestion
```

openeo/extra/job_management.py
```diff
@@ -404,6 +412,34 @@ def on_job_error(self, job: BatchJob, row):
...
+                _log.info(
+                f"Cancelling job {job.job_id} as it has been running for more than {str(self.cancel_running_job_after)}"
```
soxofaan, line 436:
```suggestion
                f"Cancelling job {job.job_id} as it has been running for more than {self.cancel_running_job_after}"
```

the `{}` construct normally does the right thing automatically without explicit "str", but I might be mistaken about timedelta. Feel free to ignore

openeo/extra/job_management.py
```diff
@@ -437,15 +470,29 @@ def _update_statuses(self, df: pd.DataFrame):
...
+                if previous_status in {"created", "queued", "started"} and new_status == "running":
+                    df.loc[i, "running_start_time"] = rfc3339.utcnow() 
+
+                if self.cancel_running_job_after and job_metadata["status"] == "running":
+                    self._cancel_prolonged_job(the_job, df.loc[i])
+                    
                 if df.loc[i, "status"] != "error" and job_metadata["status"] == "error":
                     self.on_job_error(the_job, df.loc[i])
 
+                if job_metadata["status"] == "finished":
```
soxofaan, line 488:
```suggestion
                if new_status == "finished":
```

openeo/extra/job_management.py
```diff
@@ -437,15 +470,29 @@ def _update_statuses(self, df: pd.DataFrame):
...
+                    self.on_job_done(the_job, df.loc[i])
+
+                if job_metadata["status"] == "canceled":
```
soxofaan, line 491:
```suggestion
                if new_status == "canceled":
```

openeo/extra/job_management.py
```diff
@@ -437,15 +470,29 @@ def _update_statuses(self, df: pd.DataFrame):
...
+
+                _log.info(f"Status of job {job_id!r} (on backend {backend_name}) is {new_status!r}")
+
+                if previous_status in {"created", "queued", "started"} and new_status == "running":
+                    df.loc[i, "running_start_time"] = rfc3339.utcnow() 
+
+                if self.cancel_running_job_after and job_metadata["status"] == "running":
+                    self._cancel_prolonged_job(the_job, df.loc[i])
+                    
                 if df.loc[i, "status"] != "error" and job_metadata["status"] == "error":
```
soxofaan, line 485:
```suggestion
                if previous_status != "error" and new_status == "error":
```

openeo/extra/job_management.py
```diff
@@ -437,15 +470,29 @@ def _update_statuses(self, df: pd.DataFrame):
...
+                previous_status = df.loc[i, "status"]  
+                new_status = job_metadata["status"] 
+
+                _log.info(f"Status of job {job_id!r} (on backend {backend_name}) is {new_status!r}")
+
+                if previous_status in {"created", "queued", "started"} and new_status == "running":
+                    df.loc[i, "running_start_time"] = rfc3339.utcnow() 
+
+                if self.cancel_running_job_after and job_metadata["status"] == "running":
+                    self._cancel_prolonged_job(the_job, df.loc[i])
```
soxofaan, line 483: I would move this canceling after all the other status checks. 

Because now it allows _cancel_prolonged_job to change the status in `df.loc[i]`, and because of `previous_status = df.loc[i, "status"]` above, things might get confusing/inconsistent very easily

tests/extra/test_job_management.py
```diff
@@ -455,7 +448,55 @@ def start_job(row, connection_provider, connection, **kwargs):
...
+        # Initialize manager with the max_running_duration as seconds
+        max_running_duration_seconds = 12 * 60 * 60  # 12 hours
+        manager = MultiBackendJobManager(cancel_running_job_after=max_running_duration_seconds)
+        
+        # set up timestamps
+        job_running_timestamp = rfc3339.parse_datetime(row["running_start_time"], with_timezone=True)
+        future_time = job_running_timestamp + datetime.timedelta(seconds=max_running_duration_seconds) + datetime.timedelta(seconds=1)
+
+        with time_machine.travel(future_time, tick=False):
+            manager._cancel_prolonged_job(job, row)
```
soxofaan, line 469: I think this test should not call the private `_cancel_prolonged_job`. Running the public `run_jobs` would be ideal, but this involves an "infinite" while loop, so that is not ideal. 
I'd call at least `_update_statuses` to have better coverage of the actual flow

tests/extra/test_job_management.py
```diff
@@ -455,7 +448,55 @@ def start_job(row, connection_provider, connection, **kwargs):
...
+        # Initialize manager with the max_running_duration as seconds
+        max_running_duration_seconds = 12 * 60 * 60  # 12 hours
+        manager = MultiBackendJobManager(cancel_running_job_after=max_running_duration_seconds)
+        
+        # set up timestamps
+        job_running_timestamp = rfc3339.parse_datetime(row["running_start_time"], with_timezone=True)
+        future_time = job_running_timestamp + datetime.timedelta(seconds=max_running_duration_seconds) - datetime.timedelta(seconds=1)
+        
+        with time_machine.travel(future_time, tick=False):
+            manager._cancel_prolonged_job(job, row)
```
soxofaan, line 494: see above

soxofaan (commented):
some more notes

openeo/extra/job_management.py
HansVRP: I propose to place it right above: if new_status == "canceled":

That way the manager reacts properly on the cancelled job within the same 'loop'

tests/extra/test_job_management.py
HansVRP: struggling here to make it work. Given that you actually need to run the job to create an output csv. I was trying to access the job metapath but it also doesn't exist yet when I try to run my test.

HansVRP: I believe I found a way!

jdries:
Can we merge now? We have some other changes coming up as well on this one.

HansVRP:
@soxofaan The issues seems to be resolved

soxofaan:
squashed and merged in 712a6faf95f8aa8abc77fdfaa2bd17fdd00bc8d8 :+1:

PR #602, opened 2024-08-14
Add Json default for serialising Path objects

EmileSonneveld:
After this MR, `batch_job.py` in `openeo-geopyspark-driver` needs to be updated


EmileSonneveld:
@soxofaan Is it ok to use the utility class here for tools needed in `openeo-geopyspark-driver`? I saw some other utils used there and followed example
https://github.com/Open-EO/openeo-geopyspark-driver/pull/836

PR #607, opened 2024-08-21
extend job database interface to allow more granular querying on status

jdries:
This proposal tries to support very large job databases in the job manager. It adjusts the job database interface to allow retrieving jobs with a specific status, and putting a limit on the maximum amount of jobs to return.

### DIFF
--- a/openeo/extra/job_management.py
+++ b/openeo/extra/job_management.py
@@ -75,11 +75,11 @@ class JobDatabaseInterface(metaclass=abc.ABCMeta):
         ...
 
     @abc.abstractmethod
-    def get_by_status(self, include: List[str], max=None) -> pd.DataFrame:
+    def get_by_status(self, statuses: List[str], max=None) -> pd.DataFrame:
         """
         Returns a dataframe with jobs, filtered by status.
 
-        :param include: List of statuses to include.
+        :param statuses: List of statuses to include.
         :param max: Maximum number of jobs to return.
 
         :return: DataFrame with jobs filtered by status.
@@ -370,11 +370,11 @@ class MultiBackendJobManager:
             with ignore_connection_errors(context="get statuses"):
                 self._track_statuses(job_db)
 
-            not_started = job_db.get_by_status(include=["not_started"],max=200)
+            not_started = job_db.get_by_status(statuses=["not_started"],max=200)
 
             if len(not_started) > 0:
                 # Check number of jobs running at each backend
-                running = job_db.get_by_status(include=["created","queued","running"])
+                running = job_db.get_by_status(statuses=["created","queued","running"])
                 per_backend = running.groupby("backend_name").size().to_dict()
                 _log.info(f"Running per backend: {per_backend}")
                 for backend_name in self.backends:
@@ -531,7 +531,7 @@ class MultiBackendJobManager:
         Tracks status (and stats) of running jobs (in place).
         Optionally cancels jobs when running too long.
         """
-        active = job_db.get_by_status(include=["created", "queued", "running"])
+        active = job_db.get_by_status(statuses=["created", "queued", "running"])
         for i in active.index:
             job_id = active.loc[i, "id"]
             backend_name = active.loc[i, "backend_name"]
@@ -618,17 +618,17 @@ class FullDataFrameJobDatabase(JobDatabaseInterface):
 
 
 
-    def get_by_status(self, include, max=None) -> pd.DataFrame:
+    def get_by_status(self, statuses, max=None) -> pd.DataFrame:
         """
         Returns a dataframe with jobs, filtered by status.
 
-        :param include: List of statuses to include.
+        :param statuses: List of statuses to include.
         :param max: Maximum number of jobs to return.
 
         :return: DataFrame with jobs filtered by status.
         """
         df = self.df
-        filtered = df[df.status.isin(include)]
+        filtered = df[df.status.isin(statuses)]
         return filtered.head(max) if max is not None else filtered
 
     def _merge_into_df(self, df: pd.DataFrame):
--- a/openeo/extra/job_management.py
+++ b/openeo/extra/job_management.py
@@ -6,7 +6,7 @@ import logging
 import time
 import warnings
 from pathlib import Path
-from typing import Callable, Dict, NamedTuple, Optional, Union
+from typing import Callable, Dict, NamedTuple, Optional, Union, List
 
 import pandas as pd
 import requests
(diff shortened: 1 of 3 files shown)


openeo/extra/job_management.py
```diff
@@ -325,33 +346,28 @@ def run_jobs(
             # Resume from existing db
             _log.info(f"Resuming `run_jobs` from existing {job_db}")
             df = job_db.read()
```
VictorVerhaert, line 348: This overwrites the df argument of the function, causing a lot of confusion to new job manager users. I would suggest adding a new argument to the function (something like `overlap_resolver` or `df_mode`, although I am not final at all on this name) which allows the user to state that if a job_db already exists the df assed can either be ignored (`ignore`), new lines added to the job_db (`append`) or all lines in the df should be overwritten in de job_db (`overwrite`).

This allows a user to add new jobs to an existing job_db, allows a user to overwrite for example failed jobs (perhaps with new job arguments like more memory), and the default value could be the ignore option to remain backwards compatible.

Lastly I would make the df argument optional with an exception being thrown when there is no job_db and no df, as currently the df is simply not used when a job_db is present but still required.

jdries: valid comment, but note that it applies to existing behaviour.
I have some other ideas to further refactoring job manager, but may need to keep this PR small to get it through review.

jdries:
@soxofaan @HansVRP  This is a next step in jobmanager evolution.

openeo/extra/job_management.py
```diff
@@ -337,43 +358,39 @@ def run_jobs(
...
-            status_histogram = df.groupby("status").size().to_dict()
-            _log.info(f"Status histogram: {status_histogram}")
-
-        df = self._normalize_df(df)
+            #df = job_db.read()
+            #status_histogram = df.groupby("status").size().to_dict()
+            #_log.info(f"Status histogram: {status_histogram}")
+        else:
+            df = self._normalize_df(df)
+            job_db.persist(df)
```
soxofaan, line 366: I like that this PR decouples the dataframe aspects from `run_jobs` and I think we can even take it a small step further by making the `df` argument optional  now:
- allow it to be `None` in function signature and docs
- only do the normalize and persist if it is defined

openeo/extra/job_management.py
```diff
@@ -59,11 +59,32 @@ def read(self) -> pd.DataFrame:
...
         """
         Store job data to the database.
+        The provided dataframe may contain partial information, which is merged into the larger database.
 
         :param df: job data to store.
         """
         ...
 
+    @abc.abstractmethod
+    def has_unfinished_jobs(self) -> bool:
```
soxofaan, line 69: I don't think this has to be an exposed interface: it can just be done with `get_by_status` and the right status list. 
Note that `MultiBackendJobManager` is in charge of all the statuses it sets and expects, so for proper decoupling, a `JobDatabaseInterface` should not have to worry what statuses mean "unfinished"

openeo/extra/job_management.py
```diff
@@ -59,11 +59,32 @@ def read(self) -> pd.DataFrame:
...
+        """
+        Check if there are still unfinished jobs in the database.
+
+        :return: True if there are unfinished jobs, False otherwise.
+        """
+        ...
+
+    @abc.abstractmethod
+    def get_by_status(self, include, max=None) -> pd.DataFrame:
```
soxofaan, line 78:
```suggestion
    def get_by_status(self, include: List[str], max=None) -> pd.DataFrame:
```

openeo/extra/job_management.py
```diff
@@ -337,43 +358,39 @@ def run_jobs(
...
             # Resume from existing db
             _log.info(f"Resuming `run_jobs` from existing {job_db}")
-            df = job_db.read()
-            status_histogram = df.groupby("status").size().to_dict()
-            _log.info(f"Status histogram: {status_histogram}")
-
-        df = self._normalize_df(df)
+            #df = job_db.read()
+            #status_histogram = df.groupby("status").size().to_dict()
+            #_log.info(f"Status histogram: {status_histogram}")
```
soxofaan, line 363: I'd just remove these lines, no need to keep then commented out

openeo/extra/job_management.py
```diff
@@ -337,43 +358,39 @@ def run_jobs(
...
 
             with ignore_connection_errors(context="get statuses"):
-                self._track_statuses(df)
-            status_histogram = df.groupby("status").size().to_dict()
-            _log.info(f"Status histogram: {status_histogram}")
-            job_db.persist(df)
+                self._track_statuses(job_db)
+            #TODO do we still want to support this? Would require a 'count_by_status'?
+            #status_histogram = df.groupby("status").size().to_dict()
+            #_log.info(f"Status histogram: {status_histogram}")
```
soxofaan, line 377: maybe define a `get_status() -> str` method  on `JobDatabaseInterface` to move that responsibility out of run_jobs?

openeo/extra/job_management.py
soxofaan: I'm also a bit confused by name "include", is that because you anticipate to add an "exclude" list later?

openeo/extra/job_management.py
```diff
@@ -515,16 +532,16 @@ def ensure_job_dir_exists(self, job_id: str) -> Path:
...
             job_dir.mkdir(parents=True)
 
-    def _track_statuses(self, df: pd.DataFrame):
+    def _track_statuses(self, job_db: JobDatabaseInterface):
         """
         Tracks status (and stats) of running jobs (in place).
         Optionally cancels jobs when running too long.
         """
-        active = df.loc[(df.status == "created") | (df.status == "queued") | (df.status == "running")]
+        active = self.job_db.get_by_status(include=["created", "queued", "running"])
```
soxofaan, line 540:
```suggestion
        active = job_db.get_by_status(include=["created", "queued", "running"])
```

openeo/extra/job_management.py
jdries: yes, although a job database could optimize this implementation rather than having to retrieve the full list of unfinished jobs?
We could also go for something like 'has_jobs_with_status'
or 'count_by_status'?

openeo/extra/job_management.py
jdries: done

openeo/extra/job_management.py
jdries: I now opted to move it into the file based job db itself.
If a job db is backed by an actual database, it probably makes more sense to monitor progress in something more visual.

openeo/extra/job_management.py
jdries: Proposed: count_by_status

openeo/extra/job_management.py
jdries: include -> statuses

jdries: done

soxofaan:
minor todo left: 
- [ ] changelog entry

PR #614, opened 2024-09-05
use threading to run jobmanager loop

jdries:
I propose to allow running the jobmanager cycle in a separate thread, allowing for a more clean option to interrupt it.


soxofaan (commented):
Overall I'm not sure we should do this with asyncio, as everything (HTTP requests, file IO) are just classic blocking calls currently, so there is not much to gain. 

I think in this case working with Python `threading` will be a bit simpler

openeo/extra/job_management.py
```diff
@@ -252,6 +255,43 @@ def _normalize_df(self, df: pd.DataFrame) -> pd.DataFrame:
 
         return df
 
+    def start_job_thread(self,start_job: Callable[[], BatchJob],
```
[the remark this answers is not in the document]
jdries: it's now converted to use an actual 'Thread' object, so the confusion is gone?

openeo/extra/job_management.py
```diff
@@ -355,28 +395,30 @@ def run_jobs(
...
-                _log.info(f"Running per backend: {per_backend}")
-                for backend_name in self.backends:
-                    backend_load = per_backend.get(backend_name, 0)
-                    if backend_load < self.backends[backend_name].parallel_jobs:
-                        to_add = self.backends[backend_name].parallel_jobs - backend_load
-                        to_launch = df[df.status == "not_started"].iloc[0:to_add]
-                        for i in to_launch.index:
-                            self._launch_job(start_job, df, i, backend_name)
-                            job_db.persist(df)
+            asyncio.run(self._job_update_loop(df, job_db, start_job))
```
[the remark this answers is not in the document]
jdries: asyncio removed

openeo/extra/job_management.py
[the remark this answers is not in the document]
jdries: @soxofaan if this is fine now, we can merge and continue with the other PR's

openeo/extra/job_management.py
```diff
@@ -273,6 +277,36 @@ def _normalize_df(self, df: pd.DataFrame) -> pd.DataFrame:
...
+
+        self._stop = False
+
+
+
+        def run_loop():
+            while (sum(job_db.count_by_status(statuses=["not_started","created","queued","running"]).values()) > 0 and not self._stop
+            ):
+                self._job_update_loop(df, job_db, start_job)
+
```
soxofaan, line 298: Shouldn't there be as sleep here as well?

openeo/extra/job_management.py
```diff
@@ -273,6 +277,36 @@ def _normalize_df(self, df: pd.DataFrame) -> pd.DataFrame:
...
+
+
+        self._timer = Thread(target = run_loop)
+        self._timer.start()
+
+    def stop_job_thread(self, force_timeout_seconds = 30):
+        self._stop = True
+        if(self._timer is not None):
+            self._timer.join(force_timeout_seconds)
```
soxofaan, line 307: This "force" here is misleading I think, It won't force the killing of the thread, it's just a timeout to wait for it 
the return here won't be a guarantee that the thread has finished.
At least I would print a warning when the thread is still alive (`.is_alive()`) before returning

tests/extra/test_job_management.py
```diff
@@ -85,6 +86,69 @@ def sleep_mock(self):
...
+                "geometry": ["POINT (1 2)"] * 5,
+            }
+        )
+        output_file = tmp_path / "jobs.csv"
+
+        def start_job(row, connection, **kwargs):
+            year = int(row["year"])
+            return BatchJob(job_id=f"job-{year}", connection=connection)
+
+        df = manager._normalize_df(df)
```
soxofaan, line 133: shouldn't this normalize_df be handled automatically in the manager?

jdries: we lack a good mechanism to initialize a job db correctly, we'll have to come up with something

jdries: @soxofaan  I introduced initialize_job_db to address this issue.

soxofaan:
FYI: I pushed some tweaks (some automatic code style cleanups), but more importantly some tweaks to the sleep/wait logic  in 95a4ec7. Feel free to revert

tests/extra/test_job_management.py
```diff
@@ -85,6 +80,70 @@ def sleep_mock(self):
...
+            }
+        )
+        output_file = tmp_path / "jobs.csv"
+
+        def start_job(row, connection, **kwargs):
+            year = int(row["year"])
+            return BatchJob(job_id=f"job-{year}", connection=connection)
+
+        job_db = CsvJobDatabase(output_file)
+        manager.initialize_job_db(job_db, df)
```
soxofaan, line 128: https://github.com/Open-EO/openeo-python-client/pull/614#discussion_r1768324368:
> @soxofaan I introduced initialize_job_db to address this issue.

I see, but I'd think that this initialization is not something the user should be bothered with,
can't this just be done automatically like in `run_jobs`:
https://github.com/Open-EO/openeo-python-client/blob/0c2fdc1b11f7abc8684937af7c877926c6f2000a/openeo/extra/job_management.py#L358-L363

soxofaan:
remove the `initialize_job_db` concern from the scope of this PR and merged in ffa7be2ce080d3e517a72d54c98a9ea363d9b478

Issue #520, opened 2023-12-18
Context parameter in openeo.UDF.from_file() causes confusion


JeroenVerstraelen:
A user let us know that the following code does not pass the context to his UDF:
```
cube = cube.apply_neighborhood(
    openeo.UDF.from_file("test_udf.py"),
    size=[
        {"dimension": "x", "unit": "px", "value": 32},
        {"dimension": "y", "unit": "px", "value": 32}
    ],
    overlap=[],
    context={"test_variable": "Definitely not None"}
)
```

The solution was to pass the context in the from_file method instead:
```
cube = cube.apply_neighborhood(
    openeo.UDF.from_file("test_udf.py", context={"test_variable": "Definitely not None"}),
    size=[
        {"dimension": "x", "unit": "px", "value": 32},
        {"dimension": "y", "unit": "px", "value": 32}
    ],
    overlap=[],
)
```

We should either:
- Make it clear that the context parameter should not be used when an UDF object is passed
- Merge the two context parameters in the process graph

soxofaan:
I'd like to point out that this is not only a client issue, but a less obvious part of openeo process graphs with run_udf.

There are multiple ways to get "configuration" or `context` into a UDF:

Directly in `run_udf`:
```
 {
   "process_id": "run_udf"
    "arguments": {
      ...
      "context": {"color": "red"}     
```

or through the parent process, e.g. `apply`
```
  {
    "process_id": "apply",
    "arguments": {
      ...
      "process": {
        ...
        "runudf1": {
          "process_id": "run_udf",
          "arguments": {
            "context": {"from_parameter": "context"}
            ...
        .....
      "context": {"color": "red"}  
```
note that the `context` from `apply` has to be passed through explicitly by `run_udf` with `{"from_parameter": "context"}`.
If you forget that passing through you get in the situation mentioned above 
> the following code does not pass the context to his UDF ...


The first solution (direct `context` in `run_udf`) is obviously the easiest, e.g. as proposed above too 
> solution was to pass the context in the from_file method instead


However, the solution with passing through the context from "upstream" is still important to cover, e.g. in UDPs or services where the context or parts of the context come from (UDP)  parameters.

soxofaan:
regardless, we can at least throw warnings client side when it looks like the user might have forgotten something to get the context  inside the UDF properly

soxofaan:
an alternative solution is to use `{"from_parameter": "context"}` by default in `run_udf` if there is a context set by the parent process.

[...]

PR #617, opened 2024-09-12
Documenting "inspect" and context awareness udf.rst

daviddkovacs:
Hi, I made a quick description on how to use the "inspect" function to print within the UDFs.

moreover, I tried to give a brief explanation on how users can provide their own local side variables to be used within the UDF.

### DIFF
--- a/docs/udf.rst
+++ b/docs/udf.rst
@@ -365,7 +365,7 @@ See the example below:
     """,context={"from_parameter": "context"}) # the UDF is now context aware 
 
 In the example above, the user stores a preferred value of `0.0001`, which can be passed to the UDF and used by the function.
-Later, this value is accessed by calling `context["key"]` within the UDF.
+Later, this value is accessed by calling `context["factor"]` within the UDF.
 
 Example: ``apply_dimension`` with a UDF
 ========================================
--- a/docs/udf.rst
+++ b/docs/udf.rst
@@ -317,7 +317,55 @@ To invoke a UDF like this, the apply_neighborhood method is most suitable:
             {'dimension': 'y', 'value': 128, 'unit': 'px'}
         ], overlap=[])
 
+Inspecting variables within UDF
+========================================
 
+To print and inspect variables that are within the UDF, users can use `inspect(data=[], message="")` function. 
+This will print the data that is supplied within, and show it with the message within the logs.
+
+.. code-block:: python
+    :linenos:
+    :caption: ``Inspecting UDFs``
+    :emphasize-lines: 7
+
+    # Create a UDF object from inline source code.
+    udf = openeo.UDF("""
+    import xarray
+
+    def apply_datacube(cube: xarray.DataArray, context: dict) -> xarray.DataArray:
+        cube.values = 0.0001 * cube.values  
+        inspect(data=[type(cube.values)], message="The dtype of cube.values")
+        return cube
+    """)
+
+In the above example, the `inspect` function is used to retrieve the datatype of `cube.values`. Once the job logs are opened in the Web Editor, the result will appear
+under the supplied message. This case it will be shown that `Data: <class 'numpy.ndarray'>`
+ 
+Passing user defined variables to UDF
+========================================
+
+In order to pass variables and values that are used throughout the user side of script, these need to be put in the `context` dictionary.
+Once, these variables are defined within `context` dictionary, the UDF needs to be made context aware, by adding `context={"from_parameter": "context"}` at the end of your UDF.
+See the example below:
+
+.. code-block:: python
+    :linenos:
+    :caption: ``Passing user defined values``
+    :emphasize-lines: 10
+
+    context = {"factor": 0.0001} 
+
+    # Create a UDF object from inline source code.
+    udf = openeo.UDF("""
+    import xarray
+
+    def apply_datacube(cube: xarray.DataArray, context: dict) -> xarray.DataArray:
+        cube.values = context["factor"] * cube.values  # Accessing the value stored in the context dictionary by the "factor" key.
+        return cube
+    """,context={"from_parameter": "context"}) # the UDF is now context aware 
+
+In the example above, the user stores a preferred value of `0.0001`, which can be passed to the UDF and used by the function.
+Later, this value is accessed by calling `context["key"]` within the UDF.
 
 Example: ``apply_dimension`` with a UDF
 ========================================
--- a/docs/udf.rst
+++ b/docs/udf.rst
@@ -351,9 +351,7 @@ See the example below:
 .. code-block:: python
     :linenos:
     :caption: ``Passing user defined values``
-    :emphasize-lines: 10
-
-    context = {"factor": 0.0001} 
+    :emphasize-lines: 8
 
     # Create a UDF object from inline source code.
     udf = openeo.UDF("""
@@ -364,8 +362,13 @@ See the example below:
         return cube
     """,context={"from_parameter": "context"}) # the UDF is now context aware 
 
-In the example above, the user stores a preferred value of `0.0001`, which can be passed to the UDF and used by the function.
+    user_variable = {"factor": 0.0001}
+    cube = cube.apply(udf, context = user_variable)
+
+In the example above, the user stores a preferred value of `0.0001` in the `user_variable` dictionary, 
+which can be passed to the UDF and used by the function.
 Later, this value is accessed by calling `context["factor"]` within the UDF.
+The parent UDF is called with the user's custom dictionary with `.apply(udf, context = user_variable)`.
 
 Example: ``apply_dimension`` with a UDF
 ========================================
--- a/docs/udf.rst
+++ b/docs/udf.rst
@@ -365,7 +365,7 @@ See the example below:
     user_variable = {"factor": 0.0001}
     cube = cube.apply(udf, context = user_variable)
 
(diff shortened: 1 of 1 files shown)


jdries:
Nice, thanks!
For the bit about inspecting variables, we may want to integrate it with the existing explanation about logging?
https://open-eo.github.io/openeo-python-client/udf.html#logging-from-a-udf
Or if somehow the existing explanation was hard to find, suggestions for improvement are welcome.

soxofaan:
Hi, thanks for taking the time to contribute!

About the part on `inspect` in UDF, I agree that it might be better to integrate that with the existing `inspect` docs at https://open-eo.github.io/openeo-python-client/udf.html#logging-from-a-udf  or make them easier to discover (e.g. move it up or sprinkle some references around).

About the part on passing through the context: that is indeed a valid topic to document better. Also see the discussion at https://github.com/Open-EO/openeo-python-client/issues/520
A thing that should be added to the current PR is how to pass the context to the "parent" process of the UDF

docs/udf.rst
```diff
@@ -317,7 +317,55 @@ To invoke a UDF like this, the apply_neighborhood method is most suitable:
...
+    :emphasize-lines: 10
+
+    context = {"factor": 0.0001} 
+
+    # Create a UDF object from inline source code.
+    udf = openeo.UDF("""
+    import xarray
+
+    def apply_datacube(cube: xarray.DataArray, context: dict) -> xarray.DataArray:
+        cube.values = context["factor"] * cube.values  # Accessing the value stored in the context dictionary by the "factor" key.
```
soxofaan, line 363: Avoid long lines like this, these don't render properly in the docs. For example put comments on preceding line

docs/udf.rst
```diff
@@ -317,7 +317,55 @@ To invoke a UDF like this, the apply_neighborhood method is most suitable:
...
+        return cube
+    """,context={"from_parameter": "context"}) # the UDF is now context aware 
 
+In the example above, the user stores a preferred value of `0.0001`, which can be passed to the UDF and used by the function.
+Later, this value is accessed by calling `context["key"]` within the UDF.
```
soxofaan, line 368:
```suggestion
Later, this value is accessed by calling `context["factor"]` within the UDF.
```

docs/udf.rst
```diff
@@ -317,7 +317,55 @@ To invoke a UDF like this, the apply_neighborhood method is most suitable:
...
+    # Create a UDF object from inline source code.
+    udf = openeo.UDF("""
+    import xarray
+
+    def apply_datacube(cube: xarray.DataArray, context: dict) -> xarray.DataArray:
+        cube.values = context["factor"] * cube.values  # Accessing the value stored in the context dictionary by the "factor" key.
+        return cube
+    """,context={"from_parameter": "context"}) # the UDF is now context aware 
 
+In the example above, the user stores a preferred value of `0.0001`, which can be passed to the UDF and used by the function.
```
soxofaan, line 367: In the current snippet there is a missing link to make this passing through actually happen: pass the python `context` variable to the parent `apply_...` process

daviddkovacs:
> Hi, thanks for taking the time to contribute!
> 
> About the part on `inspect` in UDF, I agree that it might be better to integrate that with the existing `inspect` docs at https://open-eo.github.io/openeo-python-client/udf.html#logging-from-a-udf or make them easier to discover (e.g. move it up or sprinkle some references around).
> 
> About the part on passing through the context: that is indeed a valid topic to document better. Also see the discussion at #520 A thing that should be added to the current PR is how to pass the context to the "parent" process of the UDF

Yes, indeed it is better to integrate into the existing docs.
I am looking into passing the context ot the parent process of the UDF (i.e. the part NOT in apply_datacube function right ?), however I didnt manage to do that, nor found anything on it. Can you indicate me where it is shown, and I'll integrate it in the PR

soxofaan:
good point, it's not easy to find succint examples on properly using context in UDF.

We have some unit test coverage in the VITO backend on this for example at https://github.com/Open-EO/openeo-geopyspark-driver/blob/cdd731ce6d684eba894beff7c8ac78266ddf12b0/tests/test_api_result.py#L718-L889, but that's probably a bit cryptic.


I think there are two use cases to document:

1. directly passing `context` to `run_udf`

```python
udf = openeo.UDF(
    "...", 
    context={"factor": 12.34},
)
cube = cube.apply(udf)
```

2.  passing `context` to parent process, and pass it through in `run_udf`

```python
udf = openeo.UDF(
    "...", 
    context={"from_parameter": "context"},
)
cube = cube.apply(udf, context={"factor": 12.34})
```

Both of these patterns have their usefulness .The first is simpler to reason about. The second is the approach to take when the context comes from "higher up", e.g. UDP parameters

soxofaan:
That being said, I think the python client should make it simpler to get that second usage pattern right. I made a ticket for that:
- https://github.com/Open-EO/openeo-python-client/issues/618

daviddkovacs (commented):
Added proper description, shorter lines and passing "context" to parent udf

docs/udf.rst
```diff
@@ -317,7 +317,58 @@ To invoke a UDF like this, the apply_neighborhood method is most suitable:
             {'dimension': 'y', 'value': 128, 'unit': 'px'}
         ], overlap=[])
 
+Inspecting variables within UDF
```
soxofaan, line 320: This section is still redundant given the existing "Logging from a UDF" section at https://github.com/Open-EO/openeo-python-client/blob/21922f1d8afb1ab9b1588fbfea56640717083e47/docs/udf.rst?plain=1#L616-L645

I'd propose to finetune the existing docs if that is necessary

docs/udf.rst
```diff
@@ -317,7 +317,58 @@ To invoke a UDF like this, the apply_neighborhood method is most suitable:
...
+        return cube
+    """)
+
+In the above example, the `inspect` function is used to retrieve the datatype of `cube.values`. Once the job logs are opened in the Web Editor, the result will appear
+under the supplied message. This case it will be shown that `Data: <class 'numpy.ndarray'>`
+ 
+Passing user defined variables to UDF
+========================================
+
+In order to pass variables and values that are used throughout the user side of script, these need to be put in the `context` dictionary.
```
soxofaan, line 347: > used throughout the user side of script

I'm not sure what you mean here, e.g. with "user side". And what "script" are you referring to? The script that build the process graph, or the UDF script?

docs/udf.rst
```diff
@@ -317,7 +317,58 @@ To invoke a UDF like this, the apply_neighborhood method is most suitable:
...
+Once, these variables are defined within `context` dictionary, the UDF needs to be made context aware, by adding `context={"from_parameter": "context"}` at the end of your UDF.
```
soxofaan, line 348:
```suggestion
Once these variables are defined within `context` dictionary, the UDF needs to be made context aware, by adding `context={"from_parameter": "context"}` at the end of your UDF.
```

> variables are defined within `context` dictionary

This is a bit confusing, because in your example you define them in a `user_variable` dictionary

docs/udf.rst
```diff
@@ -317,7 +317,58 @@ To invoke a UDF like this, the apply_neighborhood method is most suitable:
...
+
+    def apply_datacube(cube: xarray.DataArray, context: dict) -> xarray.DataArray:
+        cube.values = context["factor"] * cube.values  # Accessing the value stored in the context dictionary by the "factor" key.
+        return cube
+    """,context={"from_parameter": "context"}) # the UDF is now context aware 
+
+    user_variable = {"factor": 0.0001}
+    cube = cube.apply(udf, context = user_variable)
 
+In the example above, the user stores a preferred value of `0.0001` in the `user_variable` dictionary,
```
soxofaan, line 368:
```suggestion
In the example above, the user stores a preferred value of ``0.0001`` in the ``user_variable`` dictionary, 
```
(the docs are in restructured text format, where inline literals are fenced with double backticks, unlike Markdow where it is single backticks)

docs/udf.rst
```diff
@@ -317,7 +317,58 @@ To invoke a UDF like this, the apply_neighborhood method is most suitable:
...
+        return cube
+    """,context={"from_parameter": "context"}) # the UDF is now context aware 
+
+    user_variable = {"factor": 0.0001}
+    cube = cube.apply(udf, context = user_variable)
 
+In the example above, the user stores a preferred value of `0.0001` in the `user_variable` dictionary, 
+which can be passed to the UDF and used by the function.
+Later, this value is accessed by calling `context["factor"]` within the UDF.
+The parent UDF is called with the user's custom dictionary with `.apply(udf, context = user_variable)`.
```
soxofaan, line 371: FYI: The UDF is not the parent here, `apply` is the parent.

the hierarchical  flow is `apply` -> `run_udf` -> your UDF

soxofaan (commented):
some more notes

docs/udf.rst
soxofaan: I'm still a bit confused what you mean:
> where the user runs the UDF.

the user does not run the UDF user side, it's the backend that executes the UDF backend-side

daviddkovacs: Sorry, the script where the user **defines** the UDF script.

docs/udf.rst
daviddkovacs: ok you are right, later I will try to finetune the existing ones

docs/udf.rst
daviddkovacs: true, how should I define it?

PR #619, opened 2024-09-15
add job database implementation that uses stac

jdries:
To solve still:

- [x] how to detect authentication settings
- [ ] use aggregate api -> won't do, that api is still on dev and not clear if it works
- [x] filtering by status doesn't work yet
- [x] avoid dependency on external stac builder library


openeo/extra/job_management.py
```diff
@@ -70,14 +70,6 @@ def exists(self) -> bool:
         """Does the job database already exist, to read job data from?"""
         ...
 
-    @abc.abstractmethod
-    def read(self) -> pd.DataFrame:
```
soxofaan, line 74: I don't think we can already remove this as a required method to implement: 
it's still being called from `start_job_thread`: 
https://github.com/Open-EO/openeo-python-client/blob/47172cdb479756f0a5c1d9f53f0005b1a9675929/openeo/extra/job_management.py#L367

however, It seems to be unused there, so we could actually remove this.

openeo/extra/job_management.py
```diff
@@ -112,6 +104,20 @@ def get_by_status(self, statuses: List[str], max=None) -> pd.DataFrame:
         """
         ...
 
+    @abc.abstractmethod
+    def initialize_from_df(self, df: pd.DataFrame, on_exists: str = "error") -> "JobDatabaseInterface":
```
soxofaan, line 108: I'm not sure this method must be  a required part of  the  `JobDatabaseInterface` interface. There are still some conceptual issues with `initialize_from_df` at the moment (e.g. see #667), so we might want to avoid painting ourselves in the corner here.

 `initialize_from_df` is not something that should be called automatically from within the job manager. It is explicitly intended for end users to call explicitly on their job database. So in that sense, it should not be part of the  `JobDatabaseInterface` interface, which is a contract between the developer of the database  and job manager, not a contract between user and job manager.

openeo/extra/stac_job_db.py
```diff
@@ -0,0 +1,303 @@
+import concurrent
```
soxofaan, line 1: I think we should start making `job_manager` a package instead of module, so that files are organised like

```
- openeo
    - extra
        - job_management
            - __init__.py
            - stac_job_db.py
```

openeo/extra/stac_job_db.py
```diff
@@ -0,0 +1,303 @@
...
+        self._auth = auth
+        self.has_geometry = has_geometry
+        self.geometry_column = geometry_column
+        self.base_url = stac_root_url
+        self.bulk_size = 500
+
+
+
+    def exists(self) -> bool:
+        return len([c.id for c in self.client.get_collections() if c.id == self.collection_id]) > 0
```
soxofaan, line 59: FYI: this is a bit simpler and cheaper:
```suggestion
        return any(c.id == self.collection_id for c in self.client.get_collections())
```

openeo/extra/stac_job_db.py
```diff
@@ -0,0 +1,303 @@
...
+
+        :param item: STAC Item to be converted.
+        :return: pandas.Series
+        """
+        item_dict = item.to_dict()
+        item_id = item_dict["id"]
+        dt = item_dict["properties"]["datetime"]
+        item_dict["datetime"] = pystac.utils.str_to_datetime(dt)
+
+        return pd.Series(item_dict["properties"], name=item_id)
```
soxofaan, line 115: you only use `item_dict["properties"]`  here so the line above that sets `item_dict["datetime"]` has no use?

openeo/extra/stac_job_db.py
```diff
@@ -0,0 +1,303 @@
...
+import logging
+from typing import Iterable, List, Union
+
+import geopandas as gpd
+import numpy as np
+import pandas as pd
+import pystac
+import requests
+from pystac import Collection, Item
+from pystac_client import Client
```
soxofaan, line 12: as discussed, pystac_client is an optional dependency at the moment.

you should document that at https://github.com/Open-EO/openeo-python-client/blob/47172cdb479756f0a5c1d9f53f0005b1a9675929/docs/installation.rst?plain=1#L82-L94

tests/extra/test_stac_jobdb.py
```diff
@@ -0,0 +1,391 @@
...
+
+
+@pytest.fixture
+def mock_auth():
+    return MagicMock(spec=AuthBase)
+
+
+@pytest.fixture
+def mock_stac_api_job_database(mock_auth) -> STACAPIJobDatabase:
+    return STACAPIJobDatabase(collection_id="test_id", stac_root_url="http://fake-stac-api", auth=mock_auth)
```
soxofaan, line 24: for dummy/fake  URLs, use the [`.test` TLD](https://en.wikipedia.org/wiki/.test), which is designed for especially for test situations

soxofaan (commented):
this is quite a large PR and I didn't go through all of it already, just some initial comments

openeo/extra/job_management.py
soxofaan: with https://github.com/Open-EO/openeo-python-client/commit/29d9b160f1aa4d9ca50fe6f94a920aff9a411b0b (in master branch)  I now moved `read` from `JobDatabaseInterface` to `FullDataFrameJobDatabase`

openeo/extra/stac_job_db.py
soxofaan: With a811bffb99e0c7071fdd68e1350505d2f419d95f  I made `openeo.extra.job_management` a package now, so you can move this new module to `openeo.extra.job_management.stac_job_db` now

soxofaan: I also merge master now in this feature branch to resolve all conflicts.

so make sure to pull first before continuing on this feature branch

openeo/extra/job_management/stac_job_db.py
```diff
@@ -0,0 +1,300 @@
+import concurrent
```
soxofaan, line 1:
```suggestion
import concurrent.futures
```

openeo/extra/job_management/stac_job_db.py
```diff
@@ -0,0 +1,300 @@
...
+        item_dict["id"] = series.name
+        item = pystac.Item.from_dict(item_dict)
+        if self.has_geometry:
+            item.bbox = shape(series[self.geometry_column]).bounds
+        else:
+            item.bbox = None
+        return item
+
+    def count_by_status(self, statuses: Iterable[str] = ()) -> dict:
+        items = self.get_by_status(statuses, max=200)
```
soxofaan, line 154: `statuses: Iterable[str]` only guarantees that you can iterate through it once.
You are using `statuses` twice in this function, so the second time it might be an exhausted/empty collection.
It's best, to first make a copy of the statuses that you can use twice, e.g.
```
statuses = set(statuses)
```

openeo/extra/job_management/stac_job_db.py
```diff
@@ -0,0 +1,300 @@
...
+        self._upload_items_bulk(self.collection_id, all_items)
+
+    def _prepare_item(self, item: Item, collection_id: str):
+        item.collection_id = collection_id
+
+        if not item.get_links(pystac.RelType.COLLECTION):
+            item.add_link(pystac.Link(rel=pystac.RelType.COLLECTION, target=item.collection_id))
+
+    def _ingest_bulk(self, items: Iterable[Item]) -> dict:
+        collection_id = items[0].collection_id
```
soxofaan, line 209: because of `items: Iterable[Item]`, it's not really valid to do `items[0]` (`Iterable` only allows to do a for loop, once, not random access with e.g. `[0]`).
In this case, I think you should just use type annotation `List[Item]`

openeo/extra/job_management/stac_job_db.py
```diff
@@ -0,0 +1,300 @@
...
+            filter=status_filter,
+            max_items=max,
+        )
+
+        series = [self.series_from(item) for item in search_results.items()]
+
+        df = pd.DataFrame(series)
+        if len(series) == 0:
+            # TODO: What if default columns are overwritten by the user?
+            df = MultiBackendJobManager._normalize_df(
```
soxofaan, line 178: Using private `MultiBackendJobManager._normalize_df` from `get_by_status` looks a bit problematic.

It's also used from `initialize_from_df`, which is ok at the moment, but subject to change (see https://github.com/Open-EO/openeo-python-client/issues/667)

Using it in another context than `initialize_from_df`, like here, seems to indicate that we have to rethink all this initialization business. Not sure yet what to do instead

openeo/extra/job_management/stac_job_db.py
```diff
@@ -0,0 +1,300 @@
...
+import datetime
+import logging
+from typing import Iterable, List, Union
+
+import geopandas as gpd
+import numpy as np
+import pandas as pd
+import pystac
+import requests
+from pystac import Collection, Item
```
soxofaan, line 11: I guess it's a matter of taste, but I would avoid importing generic names like `Collection` and `Item` in the global namespace like this.
I would just use `pystac.Collection()` and the like in the code, which is a bit more typing work, but a lot clearer reading-wise

(same with `Client` from `pystac_client`)

soxofaan: Extra advantage of `import foo` over `from foo import bar`: it reduces risk on circular import issues

soxofaan (commented):
some more notes

openeo/extra/job_management/stac_job_db.py
VincentVerelst: Indeed I was hesitant to use it here first, but I see no other way to return an empty dataframe here that has all columns required by the MultiBackendJobManager

soxofaan:
I'm a bit short on time to review this deeply, let alone try running some use cases. It's good that half of the PR is unit tests
The PR also just adds a new class and doesn't touch existing code paths. So maybe we can just merge this as is

soxofaan (approved):

PR #631, opened 2024-09-25
add apply_vectorcube udf

JeroenVerstraelen:


soxofaan:
This new signature should also be added to https://github.com/Open-EO/openeo-python-client/blob/master/openeo/udf/udf_signatures.py

openeo/udf/run_code.py
```diff
@@ -196,6 +195,37 @@ def run_udf_code(code: str, data: UdfData) -> UdfData:
...
+                        c=len(data.get_feature_collection_list())
+                    )
+                )
+            if len(data.get_datacube_list()) != 1:
+                raise ValueError(
+                    "The provided UDF expects exactly one datacube, but {c} were provided.".format(
+                        c=len(data.get_datacube_list())
+                    )
+                )
+            # TODO: geopandas is optional dependency.
```
soxofaan, line 222: > TODO: geopandas is optional dependency

what does this refer to?

openeo/udf/run_code.py
```diff
@@ -196,6 +195,37 @@ def run_udf_code(code: str, data: UdfData) -> UdfData:
...
+            input_geoms = data.get_feature_collection_list()[0].data
+            input_cube = data.get_datacube_list()[0].get_array()
+            result_geoms, result_cube = func(input_geoms, input_cube)
```
soxofaan, line 225: I would make argument names more explicit:
```suggestion
            result_geoms, result_cube = func(geometries=input_geoms, cube=input_cube)
```

soxofaan:
I would also add a changelog entry

soxofaan:
also consider adding  a test at https://github.com/Open-EO/openeo-python-client/blob/master/tests/udf/test_run_code.py

openeo/udf/run_code.py
JeroenVerstraelen: It refers to the fact that the annotations check for geometries requires geopandas as a dependency. Because it is a geopandas geodataframe.
 
https://github.com/Open-EO/openeo-python-client/blob/58f0205db63efbc860b99477cc5e5dea6069224b/openeo/udf/feature_collection.py#L13

soxofaan: note that you can also do the annotation check for geopandas through a string representation, without actually depending on `import geopandas`: e.g. see the annotation checks for `openeo_udf.api.datacube.DataCube`: which refers to the legacy `openeo_udf` project

JeroenVerstraelen: I believe that checking for an annotation string for an external library might be too brittle. Especially since users could write the annotation as for example "gpd.GeoDataFrame". I have tested it and `_get_annotation_str` will not match it in those cases.

openeo/_version.py
```diff
@@ -1 +1 @@
-__version__ = "0.32.0a1"
+__version__ = "0.33.0a1"
```
soxofaan, line 1: bumping the major/minor/patch level of the version is done at time of doing a release, don't change those from a PR.
You can bump the alpha level however

```suggestion
__version__ = "0.32.0a2"
```

openeo/udf/run_code.py
soxofaan:
```python
import geopandas as gpd
import inspect

def foo(bar: gpd.GeoDataFrame):
    ...

for name, param in inspect.signature(foo).parameters.items():
    print(name, param.annotation)
```
gives 
```
bar <class 'geopandas.geodataframe.GeoDataFrame'>
```
So the annotation does not depend on how geopandas was imported

openeo/udf/udf_signatures.py
```diff
@@ -85,3 +86,16 @@ def apply_metadata(metadata: CollectionMetadata, context: dict) -> CollectionMet
 
     """
     pass
+
+
+def apply_vectorcube(geometries: any, cube: xarray.DataArray, context: dict) -> (any, xarray.DataArray):
```
soxofaan, line 91: `any` is a function, I guess you meant  `typing.Any` here.

But regardless, `Any` is way to liberal, I would at least put a GeoDataFrame annotation in string format

soxofaan (commented):
some quick notes

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `MultiBackendJobManager`: add `cancel_running_job_after` option to automatically cancel jobs that are running for too long ([#590](https://github.com/Open-EO/openeo-python-client/issues/590))
 - Added `openeo.api.process.Parameter` helper to easily create a "spatial_extent" UDP parameter
 - Wrap OIDC token request failure in more descriptive `OidcException` (related to [#624](https://github.com/Open-EO/openeo-python-client/issues/624))
+- `apply_vectorcube` udf is now supported for vector cubes ([Open-EO/openeo-geopyspark-driver#856]https://github.com/Open-EO/openeo-geopyspark-driver/issues/856)
```
soxofaan, line 16:
```suggestion
- Add support for `apply_vectorcube` UDF signature in `run_udf_code` ([Open-EO/openeo-geopyspark-driver#856]https://github.com/Open-EO/openeo-geopyspark-driver/issues/856)
```

openeo/udf/run_code.py
JeroenVerstraelen: I added an annotation check for the geometries parameter.

openeo/udf/udf_signatures.py
JeroenVerstraelen: Added the annotation in string format.

Issue #635, opened 2024-09-27
Better API for database initialization with job manager


soxofaan:
Spin-off from PR #614: how to have a clean API to initialize the job database (`_normalize_df`: add necessary columns)?

soxofaan:
Done by #636

PR #636, opened 2024-09-27
 cleaner job manager API for initializing the job database

soxofaan:
brainstorm PR for #635 to have cleaner job db initialization


jdries:
Already looks good, we could to a factory method like:
 `job_db = create_job_db(output_file,df)`
And then 'guess' the tye of db by looking at file extension or else make it more explicit.
The difference with:
` job_db = CsvJobDatabase(output_file).initialize_from_df(df)`
is however minimal, so this version is also just fine for me.

jdries (approved):

openeo/extra/job_management.py
```diff
@@ -253,8 +258,11 @@ def _make_resilient(connection):
         connection.session.mount("https://", HTTPAdapter(max_retries=retries))
         connection.session.mount("http://", HTTPAdapter(max_retries=retries))
 
-    def _normalize_df(self, df: pd.DataFrame) -> pd.DataFrame:
```
soxofaan, line 256: users are apparently using this private method `_normalize_df`, so we might want to be careful to remove/change it

openeo/extra/job_management.py
```diff
@@ -691,6 +699,25 @@ def __init__(self):
         super().__init__()
         self._df = None
 
+    def initialize_from_df(self, df: pd.DataFrame):
```
soxofaan, line 702: also have a mode/option so that users can also call it when the db already exists, but instead of exception:
- skip initialization
- or merge data,
- or ...

soxofaan:
I already moved some bits of this PR to master to make this PR more to the point

openeo/extra/job_management.py
soxofaan: done

soxofaan:
> we could to a factory method like:
> `job_db = create_job_db(output_file,df)`
> And then 'guess' the tye of db by looking at file extension or else make it more explicit.

I now also added `create_job_db` for the "guess" type from filename extension

soxofaan:
To summarize what this PR introduces:

Simplest usage would be this factory:

```python
from openeo.extra.job_management import create_job_db

job_db = create_job_db(path="path/to/jobs.csv", df=df)
```

Which will create a CSV based job database (guessed from extension), initialized from the given dataframe, and automatically normalized.

If the file already exists, this will raise an exception (which is best default behavior I think). If user is fine with skipping initialization when file already exists, they can do:
```python
job_db = create_job_db(path="path/to/jobs.csv", df=df, on_exists="skip")
```


Users that want a bit more control over what is happening can be a bit more explicit as follows:
```python

job_db = ParquetJobDatabase(path=...)

if not job_db.exists() or some_other_condition():
    df = get_df()
    job_db.initialize_from_df()
```

PR #654, opened 2024-11-04
fix broken multibackend use

DeRooBert:
index for processing started always at index=0. If running on different backend this submitted the same jobs but to different backends.


soxofaan:
great, nice catch. This actually fixes a weird unit test issue I still had to figure out.

merged in 1fef13212a5dc1a09c69e579ce6b3f0609022949

Issue #656, opened 2024-11-08
job manager: "FutureWarning: Setting an item of incompatible dtype" warning


soxofaan:
reported by @mbuchhorn:

with usage of multibackend job manager (v 0.34.0 and probably earlier)  he started to see warnings like these:
```
openeo\extra\job_management.py:539: FutureWarning: Setting an item of incompatible dtype is deprecated and will raise an error in a future version of pandas. Value 'cdse' has dtype incompatible with float64, please explicitly cast to a compatible dtype first.
  df.loc[i, "backend_name"] = backend_name
openeo\extra\job_management.py:557: FutureWarning: Setting an item of incompatible dtype is deprecated and will raise an error in a future version of pandas. Value '2024-11-07T14:02:24Z' has dtype incompatible with float64, please explicitly cast to a compatible dtype first.
  df.loc[i, "start_time"] = rfc3339.utcnow()
openeo\extra\job_management.py:559: FutureWarning: Setting an item of incompatible dtype is deprecated and will raise an error in a future version of pandas. Value 'j-241107dc8572430887c7600e3d0cb112' has dtype incompatible with float64, please explicitly cast to a compatible dtype first.
  df.loc[i, "id"] = job.job_id
```

pvbouwel:
Output I had was very similar just a few more warnings:
```

.../lib/python3.10/site-packages/openeo/extra/job_management.py:539: FutureWarning: Setting an item of incompatible dtype is deprecated and will raise an error in a future version of pandas. Value '<backend_name>' has dtype incompatible with float64, please explicitly cast to a compatible dtype first.
  df.loc[i, "backend_name"] = backend_name

.../lib/python3.10/site-packages/openeo/extra/job_management.py:557: FutureWarning: Setting an item of incompatible dtype is deprecated and will raise an error in a future version of pandas. Value '2024-11-05T15:20:45Z' has dtype incompatible with float64, please explicitly cast to a compatible dtype first.
  df.loc[i, "start_time"] = rfc3339.utcnow()

.../lib/python3.10/site-packages/openeo/extra/job_management.py:559: FutureWarning: Setting an item of incompatible dtype is deprecated and will raise an error in a future version of pandas. Value 'j-2411059f6bb2454c979736e5cdd3c481' has dtype incompatible with float64, please explicitly cast to a compatible dtype first.
  df.loc[i, "id"] = job.job_id

[...]

Issue #641, opened 2024-10-04
SettingWithCopyWarning in job manager


soxofaan:
reported by Joris C:

> Not sure whether it's due to a change in the job manager, or whether it was there already, but I'm getting a warning on the line here:
https://github.com/Open-EO/openeo-python-client/blob/master/openeo/extra/job_management.py#L650
> SettingWithCopyWarning: 
> A value is trying to be set on a copy of a slice from a DataFrame.
> Try using .loc[row_indexer,col_indexer] = value instead
 > See the caveats in the documentation: https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#returning-a-view-versus-a-copy
>  active.loc[i, "running_start_time"] = rfc3339.utcnow()

soxofaan:
I already looked into this issue, but I could not yet reproduce this warning from using the job manager. 

I can reproduce it with some simple pandas usage like
```python
df = pandas.DataFrame({"foo": range(10)})
active = df[df.foo % 3 == 0]
active.loc[3, "foo"] = 666
```

but I would need a more practical usage example of the job manager that triggers this

soxofaan:
I could somehow reproduce the problem and 803fa2b seems to fix it. Unfortunately I didn't find a way to capture it in a unit test (because I could only reproduce it by hacks I don't want to keep in the code, and I couldn't wrap my head around the exact conditions that trigger the warning), so I'll just merge this and close this ticket. Will be part of next release (0.35.0). Feel free to reopen when problem persists

PR #662, opened 2024-11-14
job manager tests: fail on any warning

soxofaan:
Attempt to harden the job manager tests: fail hard on any warning (except our own deprecation warnings)

related to #641, #656

### DIFF
--- a/tests/extra/test_job_management.py
+++ b/tests/extra/test_job_management.py
@@ -38,6 +38,13 @@ from openeo.extra.job_management import (
 from openeo.rest._testing import OPENEO_BACKEND, DummyBackend, build_capabilities
 from openeo.util import rfc3339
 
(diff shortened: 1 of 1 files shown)


tests/extra/test_job_management.py
```diff
@@ -77,6 +77,8 @@ def sleep_mock():
         yield sleep
 
 
+@pytest.mark.filterwarnings("default:.*`output_file` argument is deprecated.*:DeprecationWarning")
+@pytest.mark.filterwarnings("error")
```
soxofaan, line 81: Apparently it is also possible to set these markers at module level, (instead of only on this test class), with the `pytestmark` module global var according to https://docs.pytest.org/en/stable/how-to/capture-warnings.html#pytest-mark-filterwarnings

Issue #588, opened 2024-07-16
JobManager: include cost of job in resulting file


VictorVerhaert:
Currently only the usage stats are included in the metadata overview, but not the cost.

PR #664, opened 2024-11-19
added costs as a column to tracking databases

VictorVerhaert:
#588

### DIFF
--- a/openeo/extra/job_management.py
+++ b/openeo/extra/job_management.py
@@ -207,6 +207,7 @@ class MultiBackendJobManager:
         "cpu": _ColumnProperties(dtype="str"),
         "memory": _ColumnProperties(dtype="str"),
         "duration": _ColumnProperties(dtype="str"),
+        "costs": _ColumnProperties(dtype="str"),
     }
 
     def __init__(
@@ -744,6 +745,8 @@ class MultiBackendJobManager:
                 for key in job_metadata.get("usage", {}).keys():
                     if key in active.columns:
                         active.loc[i, key] = _format_usage_stat(job_metadata, key)
+                if "costs" in job_metadata.keys():
+                    active.loc[i, "costs"] = job_metadata.get("costs")
 
             except OpenEoApiError as e:
                 # TODO: inspect status code and e.g. differentiate between 4xx/5xx
--- a/openeo/extra/job_management.py
+++ b/openeo/extra/job_management.py
@@ -207,7 +207,7 @@ class MultiBackendJobManager:
         "cpu": _ColumnProperties(dtype="str"),
         "memory": _ColumnProperties(dtype="str"),
         "duration": _ColumnProperties(dtype="str"),
-        "costs": _ColumnProperties(dtype="str"),
+        "costs": _ColumnProperties(dtype="float64"),
     }
 
     def __init__(
--- a/openeo/rest/_testing.py
+++ b/openeo/rest/_testing.py
@@ -235,10 +235,7 @@ class DummyBackend:
             result["usage"] = {
                 "cpu": {"unit": "cpu-seconds", "value": 30.0},
                 "duration": {"unit": "seconds", "value": 55},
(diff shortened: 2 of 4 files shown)


soxofaan (commented):
quick note

openeo/extra/job_management.py
```diff
@@ -207,6 +207,7 @@ def start_job(
         "cpu": _ColumnProperties(dtype="str"),
         "memory": _ColumnProperties(dtype="str"),
         "duration": _ColumnProperties(dtype="str"),
+        "costs": _ColumnProperties(dtype="str"),
```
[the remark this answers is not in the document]
VictorVerhaert: I tried integer first, however an integer column does not allow empty values. Float64 might allow this, will give it a try!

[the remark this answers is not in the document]
soxofaan: yes, the "number" specification in JSON schema means int+float, so I would go for float64 in pandas

openeo/rest/_testing.py
```diff
@@ -225,11 +225,22 @@ def _handle_get_job(self, request, context):
...
         }
+        if self.batch_jobs[job_id]["status"] == "finished":  # HACK some realistic values for a small job
+            result["costs"] = 4
+            result["usage"] = {
+                "cpu": {"unit": "cpu-seconds", "value": 30.0},
+                "duration": {"unit": "seconds", "value": 55},
+                "input_pixel": {"unit": "mega-pixel", "value": 6.0},
+                "max_executor_memory": {"unit": "gb", "value": 0.5},
+                "memory": {"unit": "mb-seconds", "value": 150000.0},
+                "network_received": {"unit": "b", "value": 200000},
```
soxofaan, line 241: Note: `input_pixel`, `max_executor_memory`, `network_received` are specific from VITO backend. I would not include these in this default dummy implementation.  I would just stick to standard `cpu`, `memory` and `duration`

tests/extra/test_job_management.py
```diff
@@ -174,6 +174,8 @@ def test_basic(self, tmp_path, job_manager, job_manager_root_dir, sleep_mock):
             ("job-2022", "finished", "foo"),
         ]
 
+        assert not pd.read_csv(job_db_path)[["cpu", "memory", "duration", "costs"]].isnull().any().any()
```
soxofaan, line 177: this looks quite cryptic. Can't you assert with real  values. All values are predictable, right?

openeo/rest/_testing.py
VictorVerhaert: I get these usage metrics on CDSE as well, however as now they are not used by the job manager i'll remove them

tests/extra/test_job_management.py
VictorVerhaert: The values are predictable however also well hidden in the _testing.py file (which isn't in the testing folder but still rests under the src of openeo)https://github.com/Open-EO/openeo-python-client/blob/34d42b8267aab2810784d5d61596389b1d0a1a19/openeo/rest/_testing.py#L34

This is why I chose to simply check that the values are not none.
Would it help to add a comment `# check that the columns "cpu", "memory", "duration" and "costs" of finished jobs are not empty` or do you prefer to check on the actual values and add a comment referencing where they are defined? (last option does hold the danger that when the _testing.py file is moved this comment won't be updated)

Third option of course would be to immediately move the _testing.py file in this PR

soxofaan: I'd prefer to check for actual values (especially because the job manager also does some formatting of the usage values, which is not really covered by the tests I think).
The values are indeed somewhat hidden, but you can improve the situation by using values that are more obviously dummy values and easier to find, e.g.:

```
     "cpu": {"unit": "cpu-seconds", "value": 1234.5},
     "duration": {"unit": "seconds", "value": 2345},
     "memory": {"unit": "mb-seconds", "value": 56789},
```

PR #674, opened 2024-11-27
Drop read_vector

soxofaan:
For #104 , version 2 of #672


soxofaan:
this might be breaking change for some old use cases (hopefully not). Provide docs on how to still work with `read_vector` when really necessary?

Issue #505, opened 2023-11-15
Option to disable printing error logs on failed job


soxofaan:
We currently automatically print error logs when a job failed from `start_and_wait` or related.
Sometimes it's not necessary or useful to do this, so there should be an option to disable that feature

https://github.com/Open-EO/openeo-python-client/blob/bab3030e8899496e3a7665cb6e9d4cc37aca185d/openeo/rest/job.py#L308

soxofaan:
merged #687

PR #687, opened 2024-12-16
issue #505 option to disable error logs in batch jobs

ElienVandermaesenVITO:
issue #505

### DIFF
--- a/openeo/rest/datacube.py
+++ b/openeo/rest/datacube.py
@@ -2502,6 +2502,9 @@ class DataCube(_ProcessGraphAbstraction):
 
         .. versionadded:: 0.36.0
             Added argument ``additional``.
+
+        .. versionchanged:: 0.37.0
+            Added argument ``show_error_logs``.
         """
         # TODO: start showing deprecation warnings about these inconsistent argument names
         if "format" in format_options and not out_format:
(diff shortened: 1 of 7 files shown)


openeo/rest/datacube.py
```diff
@@ -2477,6 +2477,7 @@ def execute_batch(
         job_options: Optional[dict] = None,
         validate: Optional[bool] = None,
         auto_add_save_result: bool = True,
+        log_error=True,
```
soxofaan, line 2480: I think `log_error` is not ideal as a name here, as we always log errors. This feature is just about automatically showing them or not. So I would name this `show_error_logs` instead.

Also, add a type hint, e.g. `show_error_logs: bool = True` for documentation purposes

openeo/rest/datacube.py
```diff
@@ -2494,6 +2495,7 @@ def execute_batch(
         :param validate: Optional toggle to enable/prevent validation of the process graphs before execution
             (overruling the connection's ``auto_validate`` setting).
         :param auto_add_save_result: Automatically add a ``save_result`` node to the process graph if there is none yet.
+        :param log_error: whether to print error logs
```
soxofaan, line 2498:
```suggestion
        :param log_error: whether to automatically print error logs when the batch job failed.
```

soxofaan:
(rebased and) merged in c161f08e7d1cac14be60ebbbd87a6b917c8f58a3

PR #688, opened 2024-12-17
add name based getter for dimension

jdries:


openeo/metadata.py
```diff
@@ -379,6 +394,11 @@ def add_dimension(self, name: str, label: Union[str, float], type: str = None) -
             dim = Dimension(type=type or "other", name=name)
         return self._clone_and_update(dimensions=self._dimensions + [dim])
 
+    def add_dimension(self, dimension: Dimension) -> CubeMetadata:
```
soxofaan, line 397: this overwrites the existing `add_dimension` method 
(which makes some tests fail)

Issue #254, opened 2021-10-19
Clear capabilities cache on log in


soxofaan:
from https://github.com/Open-EO/openeo-api/pull/417#issuecomment-946780643

`Connection`  caches a couple of capabilties (file formats, processes, ...).
This cache should be cleared when auth changes (login, switch account)

soxofaan:
learned from going through https://github.com/Open-EO/openeo-python-client/issues/424 and https://github.com/Open-EO/openeo-api/releases/tag/1.2.0: not only capabilities cache should be cleared on auth status change, but also collection listings

soxofaan:
fixed with #691

PR #691, opened 2024-12-19
issue #254 Clear cache when authenticate

ElienVandermaesenVITO:
#254


tests/rest/test_connection.py
```diff
@@ -513,6 +513,86 @@ def test_capabilities_caching(requests_mock):
...
+            {"methods": ["GET"], "path": "/credentials/basic"},
+            {"methods": ["GET"], "path": "/account/status"},
+        ],
+    }
+    assert get_capabilities_mock.call_count == 2
+
+
+
+def test_capabilities_caching_after_authenticate_oidc(requests_mock):
+    requests_mock.get(API_URL, json={"api_version": "1.0.0"})
```
soxofaan, line 553: this requests_mock.get call is unnecessary (is overwritten a couple of lines lower)

tests/rest/test_connection.py
```diff
@@ -513,6 +513,86 @@ def test_capabilities_caching(requests_mock):
     assert con.capabilities().api_version() == "1.0.0"
     assert m.call_count == 1
 
+def test_capabilities_caching_after_authenticate_basic(requests_mock):
+    user, pwd = "john262", "J0hndo3"
+
+    def get_capabilities(request, context):
+        endpoints = BASIC_ENDPOINTS
```
soxofaan, line 520: FYI: the tests are failing because we are using BASIC_ENDPOINTS here, and appending directly to that, which causes leakage between tests.

So the fix for this is
```suggestion
        endpoints = BASIC_ENDPOINTS.copy()
```
(in both places)

soxofaan:
FYI: the test failures are due to #700 , which is already fixed in master branch, so I would merge the master branch in this feature branch to address that

soxofaan:
(rebased and) merged in 315ce7a11317cea82ac45c4f4ed9cba99657b95d

Issue #689, opened 2024-12-17
ensure parent dir on `cube.download()`


soxofaan:
synchronous download requires that parent dir already exists. Client should create it automatically.

(we do this already with batch job result download for example)

PR #696, opened 2025-01-02
issue #689 create directory in connection.download if it doesn't exist

ElienVandermaesenVITO:
#689


soxofaan:
Looks good and merged

(failing tests where because of #700)

Issue #693, opened 2024-12-20
sar_backscatter: get coefficients from schema


jdries:
OpenEO backends publish their process definitions (schemas), which in many cases follow the official schema, but sometimes not.
The current python client never uses the actual backend schemas, and instead makes assumptions, which leads to user confusion. 
It is relatively easy to: 

- check if a certain process exists before using it
- check if a certain process argument exists
- check what values are available

One example is the list of supported backscatter coefficients. I suggest to implement metadata lookup, allowing us to improve this check:
https://github.com/Open-EO/openeo-python-client/blob/ec66b318ce44ee27c0c7c1a9cfa32efcdcd7f0f8/openeo/rest/datacube.py#L2728

Suggestion is to use existing 'describe_process', but add a caching mechanism. Also important to check the case where the process doesn't exist.

https://github.com/Open-EO/openeo-python-client/blob/ec66b318ce44ee27c0c7c1a9cfa32efcdcd7f0f8/openeo/rest/connection.py#L1051

PR #698, opened 2025-01-06
issue #693 add search for schema and find parameters from the backend

ElienVandermaesenVITO:
#693


openeo/rest/connection.py
```diff
@@ -1065,6 +1065,37 @@ def describe_process(self, id: str, namespace: Optional[str] = None) -> dict:
...
 
+    """
+        Loads all available processes of the back end.
+
+        :param namespace: The namespace for which to list processes.
+
+        :return: Dictionary of All available processes of the back end.
+    """
+
+    def dict_processes(self, namespace: Optional[str] = None) -> dict:
```
soxofaan, line 1076: I don't think it's necessary to implement this method. We already have `list_processes` and `describe_process`, which should cover most/all needs (including the one in this PR)

openeo/rest/connection.py
```diff
@@ -1065,6 +1065,37 @@ def describe_process(self, id: str, namespace: Optional[str] = None) -> dict:
...
+        Returns schema of the parameter of the process from the back end.
+
+        :param process_id: The id of the process.
+        :param parameter_id: The id of the parameter.
+        :param namespace: The namespace of the process.
+
+        :return: schema of the parameter in the process.
+    """
+
+    def get_schema(self, process_id: str, parameter_id: str, namespace: Optional[str] = None) -> Union[dict, list]:
```
soxofaan, line 1093: This function name is too generic for a Connection method: there are more kinds of schema's in openEO than process parameter schema's

openeo/rest/datacube.py
```diff
@@ -2724,9 +2724,8 @@ def sar_backscatter(
         .. versionadded:: 0.4.9
         .. versionchanged:: 0.4.10 replace `orthorectify` and `rtc` arguments with `coefficient`.
         """
-        coefficient_options = [
-            "beta0", "sigma0-ellipsoid", "sigma0-terrain", "gamma0-ellipsoid", "gamma0-terrain", None
-        ]
+        schema = self.connection.get_schema("sar_backscatter", "coefficient")
+        coefficient_options = search_list_for_dict_key(schema, "enum")
```
soxofaan, line 2728: There should be fallbacks for several situations:
- there is no connection (`self.connection` is None)
- the backend does not provide an enum with possible  coefficient values

ElienVandermaesenVITO:
The test fail because of No mock address: GET https://oeo.test/processes. Do I have to change the test and add the address or is there another way to solve it?

soxofaan:
this feature indeed impacts existing tests

```
FAILED tests/rest/datacube/test_datacube100.py::test_sar_backscatter_defaults - requests_mock.exceptions.NoMockAddress: No mock address: GET https://oeo.test/processes
FAILED tests/rest/datacube/test_datacube100.py::test_sar_backscatter_custom - requests_mock.exceptions.NoMockAddress: No mock address: GET https://oeo.test/processes
FAILED tests/rest/datacube/test_datacube100.py::test_sar_backscatter_coefficient_none - requests_mock.exceptions.NoMockAddress: No mock address: GET https://oeo.test/processes
FAILED tests/rest/datacube/test_datacube100.py::test_sar_backscatter_coefficient_invalid - requests_mock.exceptions.NoMockAddress: No mock address: GET https://oeo.test/processes
```

doing `cube.sar_backscatter()` now triggers a `GET /processes` request, so you should indeed add a mock for that in these tests

openeo/util.py
```diff
@@ -687,3 +687,22 @@ def normalize_crs(crs: Any, *, use_pyproj: bool = True) -> Union[None, int, str]
             raise ValueError(f"Can not normalize CRS data {type(crs)}")
 
     return crs
+
+
+def search_list_for_dict_key(lst: list, key: str) -> Union[dict, list]:
```
ElienVandermaesenVITO, line 692: Moved this function to util.py, because it couldn't be imported from connection.py due to a circular import issue. Can it stay here or should it be moved elsewhere

openeo/rest/connection.py
```diff
@@ -1066,6 +1067,28 @@ def describe_process(self, id: str, namespace: Optional[str] = None) -> dict:
...
+        """
+        processes = self.list_processes(namespace)
+        for process in processes:
+            if process["id"] == process_id:
+                schema = Process.from_dict(process)
+                for parameter in schema.parameters:
+                    if parameter.name == parameter_id:
+                        return parameter.schema.schema
+                raise OpenEoClientException("Parameter does not exist.")
+        raise OpenEoClientException("Process does not exist.")
```
soxofaan, line 1090: with the above suggestions, this whole get_schema_from_process_parameter would become roughly this oneliner `Process.from_dict(self.describe_process(process_id)).get_parameter(parameter_name)`. 
Which would be reason to not define `get_schema_from_process_parameter` and just use that construct where you need it

openeo/rest/connection.py
```diff
@@ -1066,6 +1067,28 @@ def describe_process(self, id: str, namespace: Optional[str] = None) -> dict:
...
+        :return: schema of the parameter in the process.
+        """
+        processes = self.list_processes(namespace)
+        for process in processes:
+            if process["id"] == process_id:
+                schema = Process.from_dict(process)
+                for parameter in schema.parameters:
+                    if parameter.name == parameter_id:
+                        return parameter.schema.schema
+                raise OpenEoClientException("Parameter does not exist.")
```
soxofaan, line 1089: FYI: if you raise an exception saying that "Something does not exist", you better add actual name you've been looking up to make the error message more user-friendly and actionable

openeo/rest/datacube.py
```diff
@@ -2732,9 +2739,10 @@ def sar_backscatter(
         .. versionadded:: 0.4.9
         .. versionchanged:: 0.4.10 replace `orthorectify` and `rtc` arguments with `coefficient`.
         """
-        coefficient_options = [
-            "beta0", "sigma0-ellipsoid", "sigma0-terrain", "gamma0-ellipsoid", "gamma0-terrain", None
-        ]
+        coefficient_options = [None]
+        if self.connection:
+            schema = self.connection.get_schema_from_process_parameter("sar_backscatter", "coefficient")
+            coefficient_options += search_list_for_dict_key(schema, "enum")
```
soxofaan, line 2745: I think it's risky to put this new logic in the main code path here. I would put a try-except around it because this could still go wrong in various places (both in metadata provided by backend and client side assumptions). When something goes wrong we should show a warning about this, but fallback on the original hard-coded coefficient list.

soxofaan (commented):
some quick notes

openeo/internal/processes/parse.py
```diff
@@ -127,6 +127,12 @@ def from_json_file(cls, path: Union[str, Path]) -> Process:
         with Path(path).open("r") as f:
             return cls.from_json(f.read())
 
+    def get_parameter(self, parameter_id: str) -> Union[dict, list]:
```
soxofaan, line 130: can you add a test for this new method?

soxofaan: Also to stay closer to the openEO spec, I would just use `name` as argument here. "Parameter Id" is not really a thing in the spec

openeo/rest/connection.py
```diff
@@ -38,6 +38,7 @@
 from openeo.internal.graph_building import FlatGraphableMixin, PGNode, as_flat_graph
 from openeo.internal.jupyter import VisualDict, VisualList
 from openeo.internal.processes.builder import ProcessBuilderBase
+from openeo.internal.processes.parse import Process
```
soxofaan, line 41: this adds an unused import I think

openeo/rest/datacube.py
```diff
@@ -2732,9 +2740,19 @@ def sar_backscatter(
...
         .. versionchanged:: 0.4.10 replace `orthorectify` and `rtc` arguments with `coefficient`.
         """
-        coefficient_options = [
-            "beta0", "sigma0-ellipsoid", "sigma0-terrain", "gamma0-ellipsoid", "gamma0-terrain", None
-        ]
+        try:
+            schema = Process.from_dict(self.connection.describe_process("sar_backscatter")).get_parameter("coefficient")
+            coefficient_options = search_list_for_dict_key(schema, "enum") + [None]
+        except Exception:
+            log.warning(f"Failed to extract option for coefficient in sar_backscatter")
```
soxofaan, line 2747:
```suggestion
            log.warning(f"Failed to extract coefficient options for process `sar_backscatter`: {e}")
```

openeo/internal/processes/parse.py
```diff
@@ -127,6 +127,12 @@ def from_json_file(cls, path: Union[str, Path]) -> Process:
...
+        for parameter in self.parameters:
+            if parameter.name == parameter_id:
+                return parameter.schema.schema
+        raise LookupError(f"Expected parameter {parameter_id} not found.")
```
soxofaan, line 134:
```suggestion
        raise LookupError(f"Parameter {parameter_id!r} not found.")
```

openeo/rest/datacube.py
```diff
@@ -2732,9 +2740,19 @@ def sar_backscatter(
         .. versionadded:: 0.4.9
         .. versionchanged:: 0.4.10 replace `orthorectify` and `rtc` arguments with `coefficient`.
         """
-        coefficient_options = [
-            "beta0", "sigma0-ellipsoid", "sigma0-terrain", "gamma0-ellipsoid", "gamma0-terrain", None
-        ]
+        try:
+            schema = Process.from_dict(self.connection.describe_process("sar_backscatter")).get_parameter("coefficient")
+            coefficient_options = search_list_for_dict_key(schema, "enum") + [None]
+        except Exception:
```
soxofaan, line 2746:
```suggestion
        except Exception as e:
```

openeo/internal/processes/parse.py
```diff
@@ -127,6 +127,12 @@ def from_json_file(cls, path: Union[str, Path]) -> Process:
         with Path(path).open("r") as f:
             return cls.from_json(f.read())
 
+    def get_parameter(self, parameter_id: str) -> Union[dict, list]:
+        for parameter in self.parameters:
+            if parameter.name == parameter_id:
+                return parameter.schema.schema
```
soxofaan, line 133: I would return the whole parameter instead of just its schema to be more generic
```suggestion
                return parameter
```

openeo/util.py
soxofaan: instead of this generic utility, I would define a method `get_enum_options()` on the `Schema` object

tests/rest/datacube/test_datacube100.py
```diff
@@ -2744,7 +2744,36 @@ def test_print_json_file_path(con100, tmp_path, path_factory):
...
+                            ],
+                            "type": "string",
+                        },
+                    ],
+                },
+            ],
+            "returns": {"description": "incremented value", "schema": {"type": "integer"}},
+        }
+    ]
+    requests_mock.get(API_URL + "/processes", json={"processes": processes})
```
soxofaan, line 2776: I would not add these request_mock to the existing tests, because these tests have to keep working without change.

Instead, I would add new tests where a custom sar_backscatter is mocked.

openeo/rest/datacube.py
```diff
@@ -2767,9 +2768,19 @@ def sar_backscatter(
         .. versionadded:: 0.4.9
         .. versionchanged:: 0.4.10 replace `orthorectify` and `rtc` arguments with `coefficient`.
         """
-        coefficient_options = [
-            "beta0", "sigma0-ellipsoid", "sigma0-terrain", "gamma0-ellipsoid", "gamma0-terrain", None
-        ]
+        try:
+            schema = Process.from_dict(self.connection.describe_process("sar_backscatter")).get_parameter("coefficient").schema
+            coefficient_options = schema.get_enum_options("enum") + [None]
```
soxofaan, line 2773: On top of getting the options, we should also get the default value from the schema. It is now hardcoded to "gamma0-terrain" (above), but that should also be dynamic

tests/rest/datacube/test_datacube100.py
```diff
@@ -2828,6 +2828,53 @@ def test_sar_backscatter_coefficient_invalid(con100):
...
+def test_sar_backscatter_check_coefficient_backend(con100, requests_mock):
+    requests_mock.get(API_URL, json={"api_version": "1.0.0"})
+    processes = [
+        {
+            "id": "sar_backscatter",
+            "description": "Computes backscatter from SAR input",
+            "summary": "Computes backscatter from SAR input",
+            "parameters": [
+                {
+                    "default": "gamma0-terrain",
```
soxofaan, line 2840: This test should also check that this default value is picked

soxofaan (commented):
a couple of additional notes

openeo/rest/datacube.py
```diff
@@ -2725,15 +2725,15 @@ def ard_normalized_radar_backscatter(
...
-            coefficient: Union[str, None] = "gamma0-terrain",
-            elevation_model: Union[str, None] = None,
-            mask: bool = False,
-            contributing_area: bool = False,
-            local_incidence_angle: bool = False,
-            ellipsoid_incidence_angle: bool = False,
-            noise_removal: bool = True,
-            options: Optional[dict] = None
+        self,
+        coefficient: Union[str, None] = "default",
```
soxofaan, line 2729: I would not use string "default"  here as default value. It's a bit arbitrary (e.g. "auto" would also be an option) and not a valid value for the coefficient in line with the official spec. We've already experienced that deviating from the official openEO spec with these kind of things makes things confusing for the user.
That being said, it makes sense to consider pushing some kind of "auto"/"default" option into the official spec, but that is quite out of scope of this PR, so we need a different solution at the moment.

The problem here is indeed that you need a special value to detect that the user has not specified anything. Normally we use `None` for that, but in this case, `None` has an actual meaning "Non-normalized backscatter", which we don't want to default to
https://github.com/Open-EO/openeo-processes/blob/6141771129d58d2a292db3b91d80c812c49f4e52/proposals/sar_backscatter.json#L47-L50

The alternative is a special ["sentinel" object](https://python-patterns.guide/python/sentinel-object/#sentinel-objects), which luckily is already available in this scope, so I would just use that one for now.
https://github.com/Open-EO/openeo-python-client/blob/026881df787de007205e110f765260aeaf2e9876/openeo/rest/datacube.py#L80

openeo/rest/datacube.py
```diff
@@ -2769,8 +2769,13 @@ def sar_backscatter(
         .. versionchanged:: 0.4.10 replace `orthorectify` and `rtc` arguments with `coefficient`.
         """
         try:
-            schema = Process.from_dict(self.connection.describe_process("sar_backscatter")).get_parameter("coefficient").schema
-            coefficient_options = schema.get_enum_options("enum") + [None]
+            parameter = Process.from_dict(self.connection.describe_process("sar_backscatter")).get_parameter(
+                "coefficient"
+            )
+            schema = parameter.schema
+            coefficient_options = schema.get_enum_options() + [None]
```
soxofaan, line 2776: actually, I think that `get_enum_options` could detect that `None` is also a possible value from https://github.com/Open-EO/openeo-processes/blob/6141771129d58d2a292db3b91d80c812c49f4e52/proposals/sar_backscatter.json#L36-L51

soxofaan (commented):
almost there :smiley:

soxofaan:
Can you also add changelog entry?

PR #708, opened 2025-01-15
Issue633 threaded download

HansVRP:
Somewhat of a preemptive PR. 

I tried to thread of the downloading in a similar style as the threaded MPJM. 

Do we want to unit/integration test to see if the downloads are in fact running in a threaded manner?

### DIFF
--- a/openeo/extra/job_management/__init__.py
+++ b/openeo/extra/job_management/__init__.py
@@ -21,7 +21,7 @@ from typing import (
     Optional,
     Union,
 )
-import os
+
 import numpy
 import pandas as pd
 import requests
@@ -492,7 +492,7 @@ class MultiBackendJobManager:
         # TODO: support user-provided `stats`
         stats = collections.defaultdict(int)
 
-        while sum(job_db.count_by_status(statuses=["not_started", "created", "queued", "running", "downloading"]).values()) > 0:
+        while sum(job_db.count_by_status(statuses=["not_started", "created", "queued", "running"]).values()) > 0:
             self._job_update_loop(job_db=job_db, start_job=start_job, stats=stats)
             stats["run_jobs loop"] += 1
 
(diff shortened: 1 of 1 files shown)


HansVRP:
Secondly I want to be cautious on how this potentially interacts with export workspace.

openeo/extra/job_management/__init__.py
```diff
@@ -606,27 +606,43 @@ def _launch_job(self, start_job, df, i, backend_name, stats: Optional[dict] = No
...
+                _log.info(f"Starting download for job {job.job_id} to directory {job_dir}")
+                job.get_results().download_files(target=job_dir)
+                _log.info(f"Successfully downloaded job {job.job_id} results to {job_dir}")
+            except Exception as e:
+                _log.error(f"Error downloading job {job.job_id}: {e}")
+
+        # Start the download in a separate thread
+        _log.info(f"Starting download thread for job {job.job_id}")
+        downloader = Thread(target=download_task, daemon=True)
+        downloader.start()
```
soxofaan, line 644: you should keep track of the Thread objects, to properly `.join()` them when they are done later

openeo/extra/job_management/__init__.py
```diff
@@ -606,27 +606,43 @@ def _launch_job(self, start_job, df, i, backend_name, stats: Optional[dict] = No
...
+        # Save metadata
+        _log.info(f"Saving metadata for job {job.job_id} to {metadata_path}")
         with metadata_path.open("w", encoding="utf-8") as f:
             json.dump(job_metadata, f, ensure_ascii=False)
 
+        # Define download logic inline
+        def download_task():
+            try:
+                _log.info(f"Starting download for job {job.job_id} to directory {job_dir}")
+                job.get_results().download_files(target=job_dir)
```
soxofaan, line 636: you use `job` here as non-local variable from the `on_job_done` closure. I wonder how thread-safe this is.

A batch job object on it's own is not that complex except for the `connection` inside it: that's quite a complex object with a lot of state and risk of being thread-unsafe.  For robust threaded downloading I would  try to minimize the data structures you are sharing from the main thread to the download thread. E.g. just the backend URL (a string), the job id (a string) and a valid access token (a string) might be enough. Or even better: just pass the download URLs if they are signed (and don't require a working access token)

soxofaan (commented):
some quick notes

PR #711, opened 2025-01-22
pull job action away from track_statuses

HansVRP:


HansVRP:
@soxofaan does this design match with what we discussed yesterday?

openeo/extra/job_management/__init__.py
```diff
@@ -206,6 +206,7 @@ def __init__(
         poll_sleep: int = 60,
         root_dir: Optional[Union[str, Path]] = ".",
         *,
+        download: Optional[bool] = True,
```
soxofaan, line 209: I think this feature to disable download can be a different feature request and PR

openeo/extra/job_management/__init__.py
```diff
@@ -619,14 +631,16 @@ def on_job_done(self, job: BatchJob, row):
 
         job_metadata = job.describe()
         job_dir = self.get_job_dir(job.job_id)
+        self.ensure_job_dir_exists(job.job_id)
         metadata_path = self.get_job_metadata_path(job.job_id)
 
-        self.ensure_job_dir_exists(job.job_id)
-        job.get_results().download_files(target=job_dir)
+        if self._download:
```
soxofaan, line 637: FYI: I think this `if` can be moved to the top of this function: if you don't want to download, nothing has to happen here

openeo/extra/job_management/__init__.py
```diff
@@ -722,20 +741,20 @@ def _track_statuses(self, job_db: JobDatabaseInterface, stats: Optional[dict] =
...
-                    self.on_job_error(the_job, active.loc[i])
-
-                if previous_status in {"created", "queued"} and new_status == "running":
-                    stats["job started running"] += 1
-                    active.loc[i, "running_start_time"] = rfc3339.utcnow()
+                    jobs_error.append((the_job, active.loc[i]))
 
                 if new_status == "canceled":
                     stats["job canceled"] += 1
                     self.on_job_cancel(the_job, active.loc[i])
```
soxofaan, line 752: jobs_canceled is not handled here

openeo/extra/job_management/__init__.py
```diff
@@ -722,20 +741,20 @@ def _track_statuses(self, job_db: JobDatabaseInterface, stats: Optional[dict] =
...
 
+                if previous_status in {"created", "queued"} and new_status == "running":
```
soxofaan, line 754: did you move this code block for a reason? I'm not sure that's necessary

openeo/extra/job_management/__init__.py
```diff
@@ -540,6 +542,16 @@ def _job_update_loop(
                         stats["job_db persist"] += 1
                         total_added += 1
 
+        # Act on jobs
+        for job, metadata in jobs_error:
```
soxofaan, line 546: `metadata` might be a bit confusing. I think we use something like `row` elsewhere

soxofaan (commented):
some quick notes

openeo/extra/job_management/__init__.py
HansVRP: yes? it looks a small enough modification to immediately include as well?

openeo/extra/job_management/__init__.py
HansVRP: for readibility I wanted to combine the appendign actions together

openeo/extra/job_management/__init__.py
soxofaan: Yes, but we might want to think this  through a bit more. 
This introduces a new argument that is highly correlated to the existing `root_dir` argument.
E.g. we could also let user disable auto-download by setting `root_dir=False`.
Or we could also incorporate the request to have custom download paths (based on some template).

soxofaan: I'm also bit hesitant about the shortness of `download`, maybe it should be `auto_download` or `download_on_done`

HansVRP:
@soxofaan failing unit tests seem to be unrelated?

soxofaan:
yes failing tests are already failed on master (#715 ), so I would merge master in your branch

HansVRP:
@soxofaan there we go

jdries:
@soxofaan  @HansVRP  best to push this one forward, as it's quite urgent for lcfm. We'll also need to test it in a real world case to know if it actually solves the issue.

HansVRP:
From my end, it's ready to merge. Uncertain if it will make a difference for LCFM. There we would need to pre-create a queue of created jobs

openeo/extra/job_management/__init__.py
```diff
@@ -540,6 +540,16 @@ def _job_update_loop(
...
 
+        # Act on jobs
+        for job, row in jobs_error:
+            self.on_job_error(job, row)
+
+        for job, row in jobs_cancel:
+            self.on_job_cancel(job, row)
+
+        for job, row in jobs_done:
+            self.on_job_done(job, row)
```
soxofaan, line 551: I would do the `on_job_done` before the `on_job_error`. It's probably more important to collect finished results than to collect the failures

soxofaan:
squashed and merged in 6dbf2f054b50349426a45a2ac2df012e2d1433e2

PR #712, opened 2025-01-22
Track `resample_spatial` in cube metadata

### LINKED ISSUE #690: resample_(cube_)spatial: support metadata transform
The cube metadata classes currently do not support resampling the spatial dimension. As a consequence, this metadata is incorrect and misaligned with the actual state on the server.
Server-side, the same problem exists in our 'dry-run' setup: no resampling, so incorrect metadata.

By implementing this in this repo, we can solve both problems at once.
For the user, better client metadata for cubes gives another means to detect issues early.

soxofaan:
Closes #690


soxofaan:
- [ ] add changelog entry

Issue #668, opened 2024-11-26
 Implement support for the Federation Extension


m-mohr:
In the EOEPCA+ project, we made an evaluation of the federation-related functionality available through the openEO Python Client.
The Federation Extension can be found here: https://github.com/Open-EO/openeo-api/blob/draft/extensions/federation/README.md

The following issues have been identified for the programatic access in Python:

- [x] `GET /` - Show federation backends (federation) -> possible but undocumented https://github.com/Open-EO/openeo-python-client/issues/610
- `federation:missing` can't be accessed through the corresponding functions for the following endpoints:
  - [ ] `GET /collections`
  - [ ] `GET /processes`
  - [ ] `GET /process_graphs`
  - [ ] `GET /files`
  - [ ] `GET /jobs`
  - [ ] `GET /jobs/:id/logs`
  - [ ] `GET /services`
  - [ ] `GET /services/:id/logs`
- `federation:backends` can't be accessed through the corresponding functions for the following endpoints:
  - [ ] `POST /validation`

The following issues have been identified for the programatic access through the UI interface in notebooks such as Jupyter:
  - [ ] `GET /collections`
  - [ ] `GET /processes`
  - [ ] `GET /file_formats`
  - [ ] `GET /process_graphs`
  - [ ] `GET /files`
  - [ ] `GET /jobs`
  - [ ] `GET /jobs/:id/logs` (missing upstream in Vue Components)
  - [ ] `GET /services`
  - [ ] `GET /services/:id/logs` (missing upstream in Vue Components)
- `federation:backends` can't be accessed through the corresponding functions for the following endpoints:
  - [ ] `POST /validation`
  - [ ] `GET /udf_runtimes`  per version (missing upstream in Vue Components)
  - [ ] `GET /services/{id}` (missing upstream in Vue Components)

Related issue for the upstream issues in Vue Components: https://github.com/Open-EO/openeo-vue-components/issues/97

soxofaan:
merge #716  with first bunch of additions to support federation extension

soxofaan:
Some initial docs are at https://open-eo.github.io/openeo-python-client/federation-extension.html

soxofaan:
planned to cover in  current sprint: 

> `federation:backends` can't be accessed through the corresponding functions for the following endpoints: `POST /validation`

soxofaan:
most important parts of "programatic access in Python" are now covered I think.
Going to unassign me for now

[...]

PR #716, opened 2025-01-28
Add support for federation extension

soxofaan:
Addresses #668


soxofaan:
some more todo's

- [ ] changelog entry
- [ ] doc (e.g. in cookbook) to document how to access new federation extension data

Issue #707, opened 2025-01-15
`STACAPIJobDatabase.get_by_status()` does not always return `pandas.DataFrame` with monotonic increasing index


VincentVerelst:
https://github.com/Open-EO/openeo-python-client/blob/c1589a842161cccc27cf466c0c2b043da2eba906/openeo/extra/job_management/stac_job_db.py#L181

The order of the indices `pandas.DataFrame` in `STACAPIJobDatabase.get_by_status()` is the same as the order of indices in the STAC collection. This is not necessarily monotic increasing. However, the `MultiBackendJobManager` requires monotonic increasing indices in the DataFrame, e.g.: https://github.com/Open-EO/openeo-python-client/blob/c1589a842161cccc27cf466c0c2b043da2eba906/openeo/extra/job_management/__init__.py#L539

PR #734, opened 2025-02-18
STACAPIJobDatabase.get_by_status() sorts df on index #707

VincentVerelst:
Small bug fix in the `STACAPIJobDatabase`.


soxofaan:
ok good to know. But you still have a problem if the item ids are non-consecutive, don't start at zero or when they are non-numeric?

isn't it better to _not use_ the id as index in the dataframe, but just use a default/automatic auto-increment index?

openeo/extra/job_management/stac_job_db.py
```diff
@@ -151,13 +150,13 @@ def count_by_status(self, statuses: Iterable[str] = ()) -> dict:
         if isinstance(statuses, str):
             statuses = {statuses}
         statuses = set(statuses)
-        items = self.get_by_status(statuses, max=200)
```
soxofaan, line 154: is removing this explicit max related to the bugfix?

VincentVerelst: No it isn't, just another thing I noticed :)

VincentVerelst:
Yes, if `STACAPIJobDatabase.get_by_status()` just returns a `pandas.DataFrame` with reset indices, I think we're also fine for the `MultibackendJobManager`. But we somehow have to keep track of the original item ID as well. Except for the index, I don't see another unique identifier that we could use.

VincentVerelst:
That could work; thanks for the pointer!

openeo/extra/job_management/stac_job_db.py
```diff
@@ -92,6 +94,9 @@ def initialize_from_df(self, df: pd.DataFrame, *, on_exists: str = "error"):
                 raise ValueError(f"Invalid on_exists={on_exists!r}")
 
         df = MultiBackendJobManager._normalize_df(df)
+        # If the user doesn't specify the item_id column, we will use the index.
+        if "item_id" not in df.columns:
+            df = df.reset_index(names=["item_id"])
```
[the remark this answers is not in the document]
VincentVerelst: ok, added that

soxofaan:
I'm not sure who is already using this feature, but does it make sense to list this change in the changelog?

VincentVerelst:
Ok, updated the changelog. Should I merge, or wait for you @soxofaan ?

soxofaan:
ouch you should have merged master before adding to the changelog (which is outdated in your branch)

VincentVerelst:
ow, I don't see any merge conflicts, however? But for future reference: first merge to master and only then update changelog?

soxofaan:
> But for future reference: first merge to master and only then update changelog

no the other way around: merge master into your feature branch

VincentVerelst:
ah I see, I don't get any merge conflicts, but the log will be added to the wrong version. Will remember that!

VincentVerelst:
ok, merged master into this branch. The log was immediately in the correct place in the changelog, no merge conflicts, so I guess I got lucky

PR #736, opened 2025-02-19
Hv issue719 job manager threaded job start

HansVRP:


HansVRP:
@soxofaan maybe good to rediscuss from this point (local unit tests are passing)

openeo/extra/job_management/__init__.py
```diff
@@ -364,19 +371,30 @@ def run_loop():
...
                 stats["run_jobs loop"] += 1
 
+                # Show current stats and sleep
                 _log.info(f"Job status histogram: {job_db.count_by_status()}. Run stats: {dict(stats)}")
-                # Do sequence of micro-sleeps to allow for quick thread exit
-                for _ in range(int(max(1, self.poll_sleep))):
-                    time.sleep(1)
-                    if self._stop_thread:
-                        break
+                time.sleep(self.poll_sleep)
```
soxofaan, line 384: Why replacing the microsleeps with a long sleep here? Note that the microsleeps are to allow quick exit and avoid a seemingly unresponsive UI

openeo/extra/job_management/__init__.py
```diff
@@ -517,11 +542,7 @@ def _job_update_loop(
 
         stats = stats if stats is not None else collections.defaultdict(int)
 
-        with ignore_connection_errors(context="get statuses"):
-            jobs_done, jobs_error, jobs_cancel = self._track_statuses(job_db, stats=stats)
-            stats["track_statuses"] += 1
-
-        not_started = job_db.get_by_status(statuses=["not_started"], max=200).copy()
+        not_started = job_db.get_by_status(statuses=["not_started"], max=50).copy()
```
soxofaan, line 545: why changing the `max=200` to 50 here?

openeo/extra/job_management/__init__.py
```diff
@@ -535,13 +556,23 @@ def _job_update_loop(
...
 
         # Act on jobs
+        # TODO: move this back closer to the `_track_statuses` call above, once job done/error handling is also handled in threads?
+
+        while not self._result_queue.empty():
+            full_frame = job_db.get_by_status(statuses=["not_started", "created", "queued"]).copy()
+            self._process_result_queue(job_db, full_frame, stats)
+
+        with ignore_connection_errors(context="get statuses"):
+            jobs_done, jobs_error, jobs_cancel = self._track_statuses(job_db, stats=stats)
```
soxofaan, line 573: I don't think we can move _track_statuses to the bottom here. Doesn't most of the stuff above depend on _track_statuses?

openeo/extra/job_management/__init__.py
soxofaan: Also another problem here: looking for "not_started" will also return jobs that are currently being started in the worker thread, so I think we still need this "queued_for_start" status

openeo/extra/job_management/__init__.py
```diff
@@ -535,13 +556,23 @@ def _job_update_loop(
...
+
+                        stats["job launch"] += 1
                         stats["job_db persist"] += 1
                         total_added += 1
 
         # Act on jobs
+        # TODO: move this back closer to the `_track_statuses` call above, once job done/error handling is also handled in threads?
+
+        while not self._result_queue.empty():
+            full_frame = job_db.get_by_status(statuses=["not_started", "created", "queued"]).copy()
```
soxofaan, line 569: checking for  status created or queued is not necessary here I think? At this point we should only check for job that are not started, or being started from the thread ("queued_for_start")?

soxofaan (commented):
some quick notes

openeo/extra/job_management/__init__.py
HansVRP: I mainly need it since the not_started at this point was outdated (and I wasen't using a different status just yet. )

Introducing it, would make things easier!

openeo/extra/job_management/__init__.py
HansVRP: should switch back to 200, 

the additional status will indeed make things easier

soxofaan:
by the way, I don't think you had to create a new PR, you just could have continued on https://github.com/Open-EO/openeo-python-client/pull/730 by pushing to its feature branch

openeo/extra/job_management/__init__.py
soxofaan: then the tests have to be updated I think. 
Now it looks strange to do a bunch of  things with jobs _before_ checking their current status

openeo/extra/job_management/__init__.py
HansVRP: odd, properly copied it over from the main branch

HansVRP:
worked on your initial comments @soxofaan 

currently checking why some unit tests are not passing

HansVRP:
@soxofaan I do not think the failing unit test in 3.11 are due to my changes? 

I did a first small integration test and the results were good! a single thread is already very performant: 

image

@VictorVerhaert will perform a larger scale test for LCFM

openeo/extra/job_management/__init__.py
```diff
@@ -364,19 +371,30 @@ def run_loop():
...
+
+            worker_thread.stop_event.set()
+            worker_thread.join()
+
+            _log.info(f"Job status histogram: {job_db.count_by_status()}. Run stats: {dict(stats)}")
+            # Do sequence of micro-sleeps to allow for quick thread exit
+            for _ in range(int(max(1, self.poll_sleep))):
+                time.sleep(1)
+                if self._stop_thread:
+                    break
```
soxofaan, line 395: I'm still not sure I understand these changes. The batch of microsleeps still has to be inside the while loop I think (instead of that `sleep(1)` I think.

openeo/extra/job_management/__init__.py
```diff
@@ -493,15 +511,22 @@ def run_jobs(
...
+        worker_thread.start()
+
+        while sum(job_db.count_by_status(statuses=["not_started", "created", "queued", "queued_for_start", "running"]).values()) > 0:
             self._job_update_loop(job_db=job_db, start_job=start_job, stats=stats)
             stats["run_jobs loop"] += 1
 
             # Show current stats and sleep
             _log.info(f"Job status histogram: {job_db.count_by_status()}. Run stats: {dict(stats)}")
-            time.sleep(self.poll_sleep)
+            time.sleep(1)
```
soxofaan, line 524: why `sleep(1)`? or is this temp code?

openeo/extra/job_management/__init__.py
```diff
@@ -535,13 +560,18 @@ def _job_update_loop(
                     to_add = self.backends[backend_name].parallel_jobs - backend_load
                     for i in not_started.index[total_added : total_added + to_add]:
                         self._launch_job(start_job, df=not_started, i=i, backend_name=backend_name, stats=stats)
-                        stats["job launch"] += 1
-
                         job_db.persist(not_started.loc[i : i + 1])
+
+                        stats["job launch"] += 1
```
soxofaan, line 565: I would not move stat updates like this. Keep them as close as possible to the actual event they are about.

openeo/extra/job_management/__init__.py
```diff
@@ -535,13 +560,18 @@ def _job_update_loop(
...
                         stats["job_db persist"] += 1
                         total_added += 1
 
         # Act on jobs
+        # TODO: move this back closer to the `_track_statuses` call above, once job done/error handling is also handled in threads?
```
soxofaan, line 570: This todo is about `for job, row in jobs_done ...`  not about `_process_result_queue`

openeo/extra/job_management/__init__.py
```diff
@@ -552,6 +582,35 @@ def _job_update_loop(
...
 
+    def _process_result_queue(self, stats: Optional[dict] = None):
+        """
+        Process results from the result queue, update job statuses, and persist changes.
+
+        :param job_db: The job database interface to persist changes.
+        :param not_started: DataFrame containing jobs that are not yet started.
+        :param stats: Dictionary to track statistics.
+        """
+        while not self._result_queue.empty():
```
soxofaan, line 593: Is it intentional to have two nested `while not self._result_queue.empty()` loops?

openeo/extra/job_management/__init__.py
```diff
@@ -552,6 +582,35 @@ def _job_update_loop(
...
+            _log.info('Result queue waiting for tasks')
```
soxofaan, line 594: This log message might be confusing. At this point there is a result available in the queue, and nothing to wait on. Also we're not waiting here for tasks, but the outcomes from starting jobs

tests/extra/job_management/test_job_management.py
```diff
@@ -1761,3 +1768,88 @@ def test_with_job_manager_parameter_column_map(
...
+        result_queue = queue.Queue()
+        return _JobManagerWorkerThread(work_queue, result_queue)
+
+    def test_successful_job_start(self, worker):
+        # Setup
+        mock_conn = mock.MagicMock()
+        mock_job = mock.MagicMock()
+        mock_job.start.return_value = None
+        mock_job.status.return_value = "running"
+        mock_conn.job.return_value = mock_job
```
soxofaan, line 1786: Isn't it possible to use `requests_mock` to mock the openEO API instead of mocking our own internal implementation (which risks not testing what we actually want to test) ?

tests/extra/job_management/test_job_management.py
```diff
@@ -1761,3 +1768,88 @@ def test_with_job_manager_parameter_column_map(
...
+            worker.start()
+            sleep(1)
+            worker.stop_event.set()
+            worker.join()
+            
+            # Verify
+            mock_connect.assert_called_with(url="https://backend.test")
+            mock_conn.authenticate_bearer_token.assert_called_with(bearer_token="bearer_token")
+            assert worker.result_queue.qsize() == 1
+            result = worker.result_queue.get()
```
soxofaan, line 1805: FYI: I would always set a `get` timeout or use `get_nowait()` (both in the implementation and tests) to avoid hanging indefinitely

soxofaan:
> @soxofaan I do not think the failing unit test in 3.11 are due to my changes?

indeed, that problem has been fixed on master by now

openeo/extra/job_management/__init__.py
HansVRP: temp code indeed, reversed it

soxofaan:
> when running at full capacity you will not already 'precreate jobs' which can instantly start running

I'm not sure that pre-creating jobs instead of on the fly like we currently do will make that big of a difference as the time to create a job is usually negligible compared to the required time to start a job.  Or do you experience different timings?

HansVRP:
@soxofaan adjusted the unit tests and resolved the small bugs. 

I also proposed to split off the _JobManagerWorkerThread

to allow it in the future to also contain alternative threads for downloading, jobworkerthreadpools, .....

openeo/extra/job_management/__init__.py
```diff
@@ -364,19 +373,30 @@ def run_loop():
             # TODO: support user-provided `stats`
             stats = collections.defaultdict(int)
 
-            while (
-                sum(job_db.count_by_status(statuses=["not_started", "created", "queued", "running"]).values()) > 0
-                and not self._stop_thread
-            ):
-                self._job_update_loop(job_db=job_db, start_job=start_job)
+            # TODO: multiple workers instead of a single one? Work with thread pool?
+            worker_thread = _JobManagerWorkerThread(work_queue=self._work_queue, result_queue=self._result_queue)
```
soxofaan, line 377: In my initial sketch I just set this up with a single worker thread, but as we are aiming for bigger use cases, I think we should already go for a pool of workers from the start

tests/extra/job_management/test_thread_worker.py
```diff
@@ -0,0 +1,204 @@
...
+        
+        assert fresh_worker.result_queue.qsize() == 1
+        work_type, (jid, success, status) = fresh_worker.result_queue.get()
+        assert success is True
+        assert status == "finished"
+
+    def test_start_job_failure(self, fresh_worker, monkeypatch):
+        def mock_connect(*args, **kwargs):
+            raise ConnectionError("Backend unreachable")
+        monkeypatch.setattr(openeo, "connect", mock_connect)
```
soxofaan, line 162: I would mock the http response here like in the other tests instead of monkeypaching openeo.connect

tests/extra/job_management/test_thread_worker.py
```diff
@@ -0,0 +1,204 @@
...
+    def fresh_worker(self):
+        work_queue = queue.Queue()
+        result_queue = queue.Queue()
+        worker = _JobManagerWorkerThread(work_queue, result_queue)
+        yield worker
+        if worker.is_alive():
+            worker.stop_event.set()
+            worker.join(timeout=1)
+
+    def enqueue_and_wait(self, worker, work_item, timeout=0.2):
```
soxofaan, line 88: I'm a bit torn by this helper: it combines two things that should be independent:
- starting and stopping the thread/worker
- putting work on the work queu

soxofaan (commented):
some more notes

tests/extra/job_management/test_thread_worker.py
```diff
@@ -0,0 +1,204 @@
...
+    return dummy
+
+
+@pytest.fixture
+def sleep_mock():
+    with mock.patch("time.sleep") as sleep:
+        yield sleep
+
+
+class TestJobManagerWorkerThread:
```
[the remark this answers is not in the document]
HansVRP: I can move the 'test_run_jobs_collects_worker_results' to the job manager tests?

tests/extra/job_management/test_thread_worker.py
HansVRP: mainly wrote it to not repeat myself, I can go back to the standard code if prefered

tests/extra/job_management/test_job_management.py
```diff
@@ -336,15 +338,13 @@ def start_worker_thread():
 
         thread = threading.Thread(target=start_worker_thread, name="Worker process", daemon=True)
 
-        timeout_sec = 5.0
```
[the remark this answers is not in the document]
HansVRP: it used to fail ues, but just checked and was able to revert it back to 5 and have the test pass

tests/extra/job_management/test_job_management.py
```diff
@@ -365,6 +365,10 @@ def start_worker_thread():
             for filename in ["job-results.json", f"job_{job_id}.json", "result.data"]
         }
 
+        assert is_done_file.exists(), (
+            "MultiBackendJobManager did not finish on its own and was killed. " + "Infinite loop is probable."
```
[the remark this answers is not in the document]
HansVRP: I will move it back, I believe the test were failing and I wanted to check if the other assert was failing as well

tests/extra/job_management/test_job_management.py
```diff
@@ -590,14 +595,15 @@ def get_status(job_id, current_status):
...
         with mock.patch.object(openeo.extra.job_management.time, "sleep", new=lambda s: time_machine.shift(60 * 60)):
             job_manager.run_jobs(df=df, start_job=self._create_year_job, job_db=job_db_path)
 
         final_df = CsvJobDatabase(job_db_path).read()
-        assert final_df.iloc[0].to_dict() == dirty_equals.IsPartialDict(
-            id="job-2024", status=expected_status, running_start_time="2024-09-01T10:00:00Z"
-        )
+        print(final_df.iloc[0].to_dict())
+        assert dirty_equals.IsPartialDict(id="job-2024", status=expected_status
+                ) == final_df.iloc[0].to_dict()
```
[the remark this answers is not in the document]
HansVRP: The issue here is that, as the amount of polling has increased, we add additional time shifts. So the running start time is artificially delayed by 2 hours 

I did not see it as a problem that we now poll more often, and the status did remain the same?

tests/extra/job_management/test_job_management.py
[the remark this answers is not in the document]
HansVRP: reversed it to 5, used to fail earlier in development

tests/extra/job_management/test_job_management.py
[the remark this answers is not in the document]
HansVRP: merely for testing, moved it back

tests/extra/job_management/test_thread_worker.py
HansVRP: I mainly did not want to repeat myself throughout the tests

tests/extra/job_management/test_thread_worker.py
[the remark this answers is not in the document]
HansVRP: I added in the multibackend job manager test, to have a bit bigger test coverage. I can move this one back to the standard multibackend job manager tests?

tests/extra/job_management/test_job_management.py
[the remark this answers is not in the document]
HansVRP: merely testing

tests/extra/job_management/test_thread_worker.py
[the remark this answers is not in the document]
soxofaan: the new threaded worker feature is a hard, non-optional built-in in MultiBackendJobManager now, and an important feature, so it should be covered in test_job_management. 

test_thread_worker should be just about the thread worker (which has no dependencies on the job manager)

tests/extra/job_management/test_thread_worker.py
```diff
@@ -0,0 +1,204 @@
...
+
+        def spy_process_result_queue(stats=None):
+            try:
+                work_result = job_manager._result_queue.get_nowait()
+                captured_results.append(work_result)
+            except queue.Empty:
+                pass
+            original_process_result_queue(stats)
+
+        job_manager._process_result_queue = spy_process_result_queue
```
soxofaan, line 109: you are not only spying here, but also stealing a result from the result queue. This might cause inconsistencies, or a test stuck in infinite waiting loop

HansVRP:
@soxofaan as discussed, removed the queues from the design and using the internal threadpool queues.

soxofaan (commented):
I'm a bit in a hurry, but here is some initial feedback

openeo/extra/job_management/thread_worker.py
```diff
@@ -0,0 +1,84 @@
...
+        except Exception as e:
+            _log.exception(f"Job {job_id} failed: {e}")
+            return (job_id, False, str(e))
+
+    def process_futures(self, stats: Dict[str, int]) -> None:
+        """Process completed futures and update stats."""
+        completed = []
+        _log.info(f"Processing {len(self.futures)} futures")
+        for future in concurrent.futures.as_completed(self.futures):
+            job_id = self.futures[future]
```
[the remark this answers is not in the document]
HansVRP: I used a dictionary for the futures, such that I also have a job ID in case  an unknown worktype has been sent in. Now I can provide a better error message which points towards the relevant job ID as well

openeo/extra/job_management/thread_worker.py
```diff
@@ -0,0 +1,84 @@
...
+                raise ValueError(f"Unknown work type: {work_type}")
+        except Exception as e:
+            _log.exception(f"Job {job_id} failed: {e}")
+            return (job_id, False, str(e))
+
+    def process_futures(self, stats: Dict[str, int]) -> None:
+        """Process completed futures and update stats."""
+        completed = []
+        _log.info(f"Processing {len(self.futures)} futures")
+        for future in concurrent.futures.as_completed(self.futures):
```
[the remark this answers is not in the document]
HansVRP: I see two options here: 

1) the simplest solution would be to run the process completed futures outside of the job_update loop. 
and put it in run_jobs. 

This might however lead to an unclear code structure as on_job_done ect are handles by tje job update loop. 

2) Set a batch size for the process_completed_features, such that it always process a max amount of futures prior to moving on. 

This for me introduces unnecessary complexity as it adds additional parameters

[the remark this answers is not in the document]
soxofaan: `concurrent.futures.wait` seems to make most sense here

openeo/extra/job_management/thread_worker.py
```diff
@@ -0,0 +1,183 @@
...
+    def submit_work(self, work_type: str, work_args: Tuple[Any, ...]) -> None:
+        """
+        Submit a work item to the thread pool for processing.
+
+        Args:
+            work_type (str): The type of work to be performed.
+            work_args (Tuple[Any, ...]): A tuple of arguments required to process the work item.
+        """
+        # Validate work type and arguments
+        self._validate_work_args(work_type, work_args)
```
soxofaan, line 48: With adding things like validation, the logic around a works item will get more and more scattered in various places. I think it makes sense to encapsulate all task aspects in a dedicated class, to replace the ad-hoc tuples. This class can then even contain the actual implementation of the work (e.g. starting a job in this case), which allows to decouple the thread pool aspects from the specifics of a task.

e.g. very roughly (using dataclasses to compactly define a "named tuple"):

```
import dataclasses

@dataclasses.dataclass(frozen=True)
class JobStartTask:
    root_url: str
    bearer_token: str
    job_id: str

    # possibly an __init__ here, e.g. to do validation if that is necessary

   def execute(self):
       # create connection and start job ...
```

openeo/extra/job_management/thread_worker.py
```diff
@@ -0,0 +1,183 @@
...
+                    stats["job start failed"] += 1
+                    _log.error(f"Job {job_id} failed: {data}")
+            except Exception as e:
+                _log.exception(f"Unexpected error processing future: {e}")
+            completed.append(future)
+
+        _log.info(f"Processed {len(completed)} jobs")
+        # Remove processed futures
+        for future in completed:
+            self._futures.remove(future)
```
soxofaan, line 135: can't this be done directly in the previous for loop ?

HansVRP:
@soxofaan not sure why these test are failing, but I think the code is at a god point to reevaluate.

Uncertain wheter we need to post process after shutting down the worker pool, given the 

while (
            sum(
                job_db.count_by_status(
                    statuses=["not_started", "created", "queued", "queued_for_start", "running"]
                ).values()
                
loop. 

As long if we have not started or queued for start states (on which the postprocessing would touch, we remain in the loop)

soxofaan:
> Uncertain wheter we need to post process after shutting down the worker pool, given the

good point, however, that probably only works out now with doing the start in a side thread. Once we add threaded result downloading, that `while(sum(...))`  is not going to guarantee that all the (side) work is done yet.

openeo/extra/job_management/_thread_worker.py
```diff
@@ -0,0 +1,95 @@
...
+        try:
+            conn = openeo.connect(self.root_url)
+            if self.bearer_token:
+                conn.authenticate_bearer_token(self.bearer_token)
+            job = conn.job(self.job_id)
+            job.start()
+            status = job.status()
+            return (self.job_id, True, status)
+        except Exception as e:
+            return (self.job_id, False, str(e))
```
soxofaan, line 40: I'm not sure the job_id should be part of the result as it is already an attribute of the Task

Also instead of returning ad-hoc tuples (which might differ across different types of tasks), I would define a simple generic `TaskResult` class that has two attributes:  `success` (boolean) and `value` (anything)

openeo/extra/job_management/__init__.py
```diff
@@ -668,25 +708,72 @@ def on_job_cancel(self, job: BatchJob, row):
...
+        :param df:
+            The dataFrame that specifies the jobs and tracks their statuses. The status of jobs in the DataFrame is updated based on the task results.
+
+        :param stats:
+            The dictionary for tracking statistics related to job processing. The statistics are updated based on the task results.
+
+        """
+
+        for task, result in worker_pool_results:
+            job_id, success, data = result
```
soxofaan, line 736: as mentioned I would decouple this ad-hoc tuple unpacking by using a `_TaskResult` instead

openeo/extra/job_management/__init__.py
```diff
@@ -668,25 +708,72 @@ def on_job_cancel(self, job: BatchJob, row):
...
+
+            # Find rows with matching job id and a status of 'queued_for_start'
+            idx = df.index[(df["id"] == job_id) & (df["status"] == "queued_for_start")]
```
soxofaan, line 739: in this generic `_postprocess_futures` we should not look at "queued_for_start", that should only be a concern when we are clearly in a `_JobStartTask` context, which is not explicitly the case .
I know it's the only possibility at the moment, I just mean that this current implementation is not future-proof

openeo/extra/job_management/__init__.py
```diff
@@ -668,25 +708,72 @@ def on_job_cancel(self, job: BatchJob, row):
...
+            
+            if not idx.empty:
```
soxofaan, line 741: I guess this `if` should also get an  `else`?

openeo/extra/job_management/__init__.py
```diff
@@ -668,25 +708,72 @@ def on_job_cancel(self, job: BatchJob, row):
...
+                # Set new status based on task success
+                new_status = "queued" if success else "start_failed"
+                df.loc[idx, "status"] = new_status
+                _log.info(f"Updated job {job_id} status to {new_status} in dataframe.")
+            else:
+                _log.warning(f"No entry for job {job_id} with status 'queued_for_start' found in dataframe, passing on.")
+            
+            # Log details and update statistics
+            if isinstance(task, _JobStartTask) and success:
+                _log.info(f"Job {job_id} started sucessfully with status: {data}")
```
soxofaan, line 751: It feels like these logs have a lot of duplication with the logs a couple of lines higher

soxofaan (commented):
some quick notes

openeo/extra/job_management/_thread_worker.py
```diff
@@ -0,0 +1,108 @@
...
+    def process_futures(self) -> List[_TaskResult]:
+
+        updates = []
+        done, _ = concurrent.futures.wait(
+            [f for f, _ in self._future_task_pairs], timeout=0,
+            return_when=concurrent.futures.FIRST_COMPLETED
+        )
+
+        # Process completed futures and their tasks
+        for future, task in self._future_task_pairs[:]:
```
soxofaan, line 94: While thinking about whether it's safe to copy the whole list of futures (by the `[:]` trick), I started wondering if this whole method couldn't be reshaped to something like

```python

results = []
to_keep = []

for future, task in self._future_task_pairs:
    if future.done():
        try:
            results.append(future.result())
        except Exception as e:
            ...
    else:
        to_keep.append((future, task))

self._future_task_pairs = to_keep
return results 

```

soxofaan (commented):
some more notes

openeo/extra/job_management/_thread_worker.py
```diff
@@ -0,0 +1,108 @@
...
+            if future in done:
+                try:
+                    updates.append(future.result())
+                    
+                except Exception as e:
+                    _log.exception(f"Error processing task: {e}")
```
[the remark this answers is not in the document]
HansVRP: This would indeed occur incase we cannot get the result from a completed future. 

Uncertain how we then would want to update the dataframe, given that we do not know if the actual job was successful or failed. 

My idea here was to simply add some logging and let the back-end handle the status update on its own....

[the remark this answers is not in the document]
soxofaan: if no `_TaskResult` is "returned" for that job, it will stay in "queued_for_start" state forever, and the job running loop will never terminate I guess.
So I think we should put a dummy `_TaskResult` here to update the job status in some way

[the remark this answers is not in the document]
soxofaan: I don't think it makes sense to resubmit the task, because this is an internal logic failure, it's not something like a failed request that could be retried

HansVRP:
@soxofaan ready for review

HansVRP:
image

Also checking a real tests. Jobs are getting created and running in parallel fast. Now the main limitation is the amouint of parallel jobs set in the job manager itself

HansVRP:
image

scl_native.tif
soxofaan, line 1: hmm these random tiff files are probably not intended to be committed

soxofaan: I've pushed a fixup commit

openeo/extra/job_management/_thread_worker.py
```diff
@@ -0,0 +1,195 @@
...
+        # Process completed futures and their tasks
+        for future, task in self._future_task_pairs:
+            if future in done:
+                try:
+                    result = future.result()
+
+                except Exception as e:
+                    _log.exception(f"Error processing task: {e}")
+                    result = _TaskResult(
+                        job_id=task.job_id, db_update={"status": "start_failed"}, stats_update={"start_job error": 1}
```
soxofaan, line 182: `task.job_id` is not guaranteed to exist
this also introduces _JobStartTask specifics into the generic _JobManagerWorkerThreadPool

tests/extra/job_management/test_thread_worker.py
```diff
@@ -0,0 +1,166 @@
...
+            {"id": "job-123", "status": "queued_for_start", "other_field": "foo"},
+            {"id": "job-456", "status": "queued_for_start", "other_field": "bar"},
+            {"id": "job-789", "status": "other", "other_field": "baz"},
+        ]
+    )
+    return df
+
+
+@pytest.fixture
+def initial_stats():
```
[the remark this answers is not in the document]
soxofaan: removed

tests/extra/job_management/test_thread_worker.py
```diff
@@ -0,0 +1,166 @@
...
+import requests
+
+# Import the refactored classes and helper functions from your codebase.
+# Adjust the import paths as needed.
+from openeo.extra.job_management._thread_worker import (
+    _JobManagerWorkerThreadPool,
+    _JobStartTask,
+)
+
+# --- Fixtures and Helpers ---
```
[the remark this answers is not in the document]
soxofaan: removed

tests/extra/job_management/test_thread_worker.py
```diff
@@ -0,0 +1,166 @@
...
+        results = worker_pool.process_futures()
+        if results:
+            return results
+        time.sleep(interval)
+    raise TimeoutError(f"Timed out after {timeout}s waiting for worker pool results.")
+
+
+# --- Tests for the Worker Thread Pool and Futures Postprocessing ---
+
+
```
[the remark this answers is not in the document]
soxofaan: added some tests

openeo/extra/job_management/_thread_worker.py
soxofaan: moved job_id to base class
and also created intermediate ConnectedTask class

tests/extra/job_management/test_thread_worker.py
```diff
@@ -0,0 +1,166 @@
...
+    return _setup
+
+
+@pytest.fixture
+def valid_task():
+    """Fixture to create a valid _JobStartTask instance."""
+    return _JobStartTask(root_url="https://foo.test", bearer_token="test-token", job_id="test-job-123")
+
+
+import time
```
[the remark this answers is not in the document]
soxofaan: fixed

tests/extra/job_management/test_thread_worker.py
```diff
@@ -0,0 +1,196 @@
...
+
+
+
+@pytest.fixture
+def successful_backend_mock(requests_mock):
+    """
+    Returns a helper to set up a successful backend.
+    Mocks a version check, job start, and job status check.
+    """
+    # TODO: use DummyBackend here instead?
```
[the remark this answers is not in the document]
soxofaan: done

tests/extra/job_management/test_thread_worker.py
```diff
@@ -0,0 +1,196 @@
...
+
+
+class TestJobManagerWorkerThreadPool:
+    def test_worker_thread_lifecycle(self, worker_pool):
+        """Test that the worker thread pool starts and shuts down as expected."""
+
+        # Before shutdown, the executor should be active
+        assert not worker_pool._executor._shutdown
+        worker_pool.shutdown()
+        assert worker_pool._executor._shutdown
```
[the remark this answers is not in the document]
soxofaan: done

tests/extra/job_management/test_thread_worker.py
```diff
@@ -0,0 +1,196 @@
...
+
+    def test_worker_pool_bookkeeping(self, worker_pool, valid_task, successful_backend_mock, requests_mock):
+        """Ensure that processed futures are removed from the pool's internal tracking."""
+        successful_backend_mock(valid_task.root_url, valid_task.job_id)
+        worker_pool.submit_task(valid_task)
+        results = wait_for_results(worker_pool)
+
+        # Assuming your refactoring clears out futures after processing,
+        # the internal list (or maps) should be empty.
+        assert len(worker_pool._future_task_pairs) == 0
```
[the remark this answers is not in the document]
soxofaan: done

openeo/extra/job_management/__init__.py
```diff
@@ -598,26 +627,88 @@ def _launch_job(self, start_job, df, i, backend_name, stats: Optional[dict] = No
...
+        """
+        results, _ = worker_pool.process_futures()
+        stats_updates = collections.defaultdict(int)
+
+        for result in results:
+            try:
+                # Handle job database updates
+                if result.db_update:
+                    _log.debug(f"Processing update for job {result.job_id}")
+                    job_db._update_row(job_id=result.job_id, updates=result.db_update)
```
soxofaan, line 698: `_update_row` is not defined on general `JobDatabaseInterface`, only on `FullDataFrameJobDatabase`, so this is going to break `STACAPIJobDatabase` workflows

openeo/extra/job_management/__init__.py
```diff
@@ -877,6 +967,52 @@ def _merge_into_df(self, df: pd.DataFrame):
         else:
             self._df = df
 
+    def _update_row(self, job_id: str, updates: dict):
```
soxofaan, line 970: It's not ideal to introduce a new required API `_update_row` here, while there is already `persist`, which is very similar, e.g. see it's docs:
>  Store job data to the database.
>  The provided dataframe may contain partial information, which is merged into the larger database.

openeo/extra/job_management/__init__.py
```diff
@@ -598,26 +627,88 @@ def _launch_job(self, start_job, df, i, backend_name, stats: Optional[dict] = No
...
+                            _log.info(f"Submitting task {task} to thread pool")
+                            self._worker_pool.submit_task(task)
+
+                            stats["job_queued_for_start"] += 1
+                            df.loc[i, "status"] = "queued_for_start"
                         except OpenEoApiError as e:
-                            _log.error(e)
-                            df.loc[i, "status"] = "start_failed"
-                            stats["job start error"] += 1
+                            _log.info(f"Failed submitting task {task} to thread pool with error: {e}")
```
soxofaan, line 650: Why changing log level to info? 
In current implementation this situation is not recovered from, so that should be flagged properly I think

tests/extra/job_management/test_job_management.py
```diff
@@ -720,6 +748,47 @@ def get_status(job_id, current_status):
         filled_running_start_time = final_df.iloc[0]["running_start_time"]
         assert isinstance(rfc3339.parse_datetime(filled_running_start_time), datetime.datetime)
 
+    def test_process_threadworker_updates(self, job_manager, tmp_path):
```
soxofaan, line 751: this test seems to be interacting too much with internal implementation details of _JobManagerWorkerThreadPool to be a test about the job manager:
- `DummyTask` copies `_JobStartTask` behavior, why not use `_JobStartTask` directly?
- actively waiting for the futures to be done: that should be work for the job manager
- calling `_process_threadworker_updates`: that should be work for job manager too

soxofaan:
FYI: I merged master in your feature branch `hv_issue719-job-manager-threaded-job-start`  to trigger a new build

openeo/extra/job_management/__init__.py
```diff
@@ -525,7 +553,7 @@ def _job_update_loop(
         not_started = job_db.get_by_status(statuses=["not_started"], max=200).copy()
         if len(not_started) > 0:
             # Check number of jobs running at each backend
-            running = job_db.get_by_status(statuses=["created", "queued", "running"])
+            running = job_db.get_by_status(statuses=["created", "queued", "queued_for_start", "running"])
```
soxofaan, line 570: "running" is misleading name here (it also includes jobs that are only created). More appropriate would be something like "assigned" .
I'm actually wondering if "created" and "queued_for_start" should be included here: a job that is only created does not cause a "load" to account for.

openeo/extra/job_management/__init__.py
```diff
@@ -542,7 +570,9 @@ def _job_update_loop(
                         stats["job_db persist"] += 1
                         total_added += 1
 
-        # Act on jobs
+        self._process_threadworker_updates(self._worker_pool, job_db, stats)
```
soxofaan: use keyword args for clarity/safety

openeo/extra/job_management/__init__.py
```diff
@@ -599,26 +628,101 @@ def _launch_job(self, start_job, df, i, backend_name, stats: Optional[dict] = No
...
+                            stats["job queued for start failed"] += 1
             else:
                 # TODO: what is this "skipping" about actually?
                 df.loc[i, "status"] = "skipped"
                 stats["start_job skipped"] += 1
 
+    def _process_threadworker_updates(
+        self,
+        worker_pool: '_JobManagerWorkerThreadPool',
+        job_db: 'JobDatabaseInterface',
```
soxofaan: I think these type annotations don't need quoting

openeo/extra/job_management/__init__.py
```diff
@@ -599,26 +628,101 @@ def _launch_job(self, start_job, df, i, backend_name, stats: Optional[dict] = No
...
-                            df.loc[i, "status"] = job.status()
-                            stats["job get status"] += 1
+                            job_con = job.connection
+                            task = _JobStartTask(
+                                root_url=job_con.root_url,
+                                bearer_token=job_con.auth.bearer if isinstance(job_con.auth, BearerAuth) else None,
+                                job_id=job.job_id,
+                                df_idx = i
+                            )
+                            _log.info(f"Submitting task {task} to thread pool")
```
soxofaan, line 660: logging the task like this exposes the bearer token

openeo/extra/job_management/__init__.py
```diff
@@ -599,26 +628,101 @@ def _launch_job(self, start_job, df, i, backend_name, stats: Optional[dict] = No
...
+                    _log.error(
+                        f"Skipping invalid stats_update {res.stats_update} for job '{res.job_id}': {e}"
+                    )
+
+        # No valid updates: nothing to persist
+        if not updates:
+            return
+
+        # Build DataFrame of updates indexed by df_idx
+        df_updates = pd.DataFrame(updates).set_index('df_idx', drop=True)
```
soxofaan: `set_index` can be avoided I think with
```suggestion
        df_updates = pd.DataFrame(updates, index="df_idx")
```

openeo/extra/job_management/__init__.py
```diff
@@ -599,26 +628,101 @@ def _launch_job(self, start_job, df, i, backend_name, stats: Optional[dict] = No
...
+
+        # Determine which rows to upsert
+        existing_indices = set(df_updates.index).intersection(job_db.read().index)
```
soxofaan: any reason to do `job_db.read()`  here instead of using `job_db` as is?
problem is that `read()` method is not generally defined for `JobDatabaseInterface` ,  it's only a defined on `FullDataFrameJobDatabase`,  so this will break `STACAPIJobDatabase` usage, which does not have `read()` method.

to do, to check:
- possible to drop `read()`?
- make the `read()` conditional on `isinstance(..., FullDataFrameJobDatabase)`

openeo/extra/job_management/__init__.py
```diff
@@ -874,11 +977,12 @@ def get_by_status(self, statuses, max=None) -> pd.DataFrame:
 
     def _merge_into_df(self, df: pd.DataFrame):
         if self._df is not None:
-            self._df.update(df, overwrite=True)
+            self._df.update(df, overwrite=True) 
         else:
             self._df = df
 
 
+
```
soxofaan, line 999: drop these noisy whitespace edits

soxofaan (commented):
some initial notes

openeo/extra/job_management/__init__.py
soxofaan: nope, nevermind

tests/extra/job_management/test_job_management.py
```diff
@@ -81,6 +88,25 @@ def sleep_mock():
         yield sleep
 
 
+class DummyTask(Task):
+    """
+    A Task that simply sleeps and then returns a predetermined _TaskResult.
```
soxofaan: The "sleep" feature of this dummy task was lost here. Is that intentional?

openeo/extra/job_management/__init__.py
soxofaan: after looking a bit more into this: 
so the `job_db.read()` is to get a pandas dataframe from the db, so the read can not simply be dropped. Still this approach will break `STACAPIJobDatabase` usage. So I wonder, if it isn't just possible to `persist` all updates without looking to existing indices.

openeo/extra/job_management/__init__.py
```diff
@@ -599,26 +628,99 @@ def _launch_job(self, start_job, df, i, backend_name, stats: Optional[dict] = No
...
+            return
+
+        # Build DataFrame of updates indexed by df_idx
+        df_updates = pd.DataFrame(updates).set_index("df_idx", drop=True)
+
+        # Determine which rows to upsert
+        existing_indices = set(df_updates.index).intersection(job_db.read().index)
+        if existing_indices:
+            df_upsert = df_updates.loc[sorted(existing_indices)]
+            job_db.persist(df_upsert)
```
soxofaan: Hmm, I've been looking into compatibility of this approach with `STACAPIJobDatabase`, and it seems that its `persist` does not support partial updates (at the moment): the whole item/row must be persisted, not a subset of "columns" (or properties)

openeo/extra/job_management/__init__.py
soxofaan: FYI: I added a test on master with 4dfdf77ca8bb1158b00562d67591752ef9845054  to have some minimal coverage of `MultiBackendJobManager.run_jobs` with `STACAPIJobDatabase`

soxofaan: which indeed exposes this problem now
```
AttributeError: 'STACAPIJobDatabase' object has no attribute 'read'
```

soxofaan: removed the unnecessary split in  `existing_indices` and `missing` with cd26f5a2bbc25a6bc72114e3d04a11b54d9f969d

openeo/extra/job_management/__init__.py
soxofaan: `test_run_jobs_basic` now indeed fails with:

```
openeo/extra/job_management/__init__.py:712: in _process_threadworker_updates
    job_db.persist(df_updates)
openeo/extra/job_management/stac_job_db.py:209: in persist
    df.apply(handle_row, axis=1)
/opt/hostedtoolcache/Python/3.11.13/x64/lib/python3.11/site-packages/pandas/core/frame.py:10381: in apply
    return op.apply().__finalize__(self, method="apply")
           ^^^^^^^^^^
/opt/hostedtoolcache/Python/3.11.13/x64/lib/python3.11/site-packages/pandas/core/apply.py:916: in apply
    return self.apply_standard()
           ^^^^^^^^^^^^^^^^^^^^^
/opt/hostedtoolcache/Python/3.11.13/x64/lib/python3.11/site-packages/pandas/core/apply.py:1063: in apply_standard
    results, res_index = self.apply_series_generator()
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
/opt/hostedtoolcache/Python/3.11.13/x64/lib/python3.11/site-packages/pandas/core/apply.py:1081: in apply_series_generator
    results[i] = self.func(v, *self.args, **self.kwargs)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
openeo/extra/job_management/stac_job_db.py:206: in handle_row
    item = self.item_from(series)
           ^^^^^^^^^^^^^^^^^^^^^^
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <openeo.extra.job_management.stac_job_db.STACAPIJobDatabase object at 0x7ff01ca751d0>
series = id        job-001
status     queued
Name: 1, dtype: object

    def item_from(self, series: pd.Series) -> pystac.Item:
        """
        Convert a pandas.Series to a STAC Item.
    
        :param series: pandas.Series to be converted.
        :param geometry_name: Name of the geometry column in the series.
        :return: pystac.Item
        """
        series_dict = series.to_dict()
>       item_id = series_dict.pop("item_id")
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^
E       KeyError: 'item_id'

openeo/extra/job_management/stac_job_db.py:131: KeyError
```

soxofaan: In this (separate) PR I try to make `STACAPIJobDatabase.persist` support partial updates:
- https://github.com/Open-EO/openeo-python-client/pull/794

soxofaan:
ok tests pass again :partying_face: 

some remaining todos
- changelog entry covering both the threaded aspect, as well as the new API `JobDatabaseInterface.get_by_indices` for job db implementers to implement
- double check the lost sleep in the dummy task (see higher)

openeo/extra/job_management/__init__.py
soxofaan: I'll add it as todo note

soxofaan:
FYI: as this is a very long running PR with messy history, I tried to rebase the feature branch of this PR to clean it up a bit (e.g. commit squashing) to improve the signal-noise ratio and created PR #806 . The diff is practically identical to #736 (except for one empty line).

The discussion can stay here (#736),  but #806 will be merged in the end

soxofaan:
ok, I decided to merge this (through PR #806) in 41710047e0f02ba546de14390f1d390b0e2f6a5b :partying_face:

PR #739, opened 2025-02-24
Leverage pystac for deeper spatial dimension detection

### LINKED ISSUE #737: load_stac + resample_spatial incompatibility
minimal reproduction example:

```
import openeo
connection = openeo.connect(url="openeo.dataspace.copernicus.eu")

cube = connection.load_stac("https://stac.openeo.vito.be/collections/wenr_features")
cube.resample_spatial(projection=3035, resolution=10.0)
```
gives
> `MetadataException: Expected two spatial dimensions but found spatial_indices=[]`

Current load_stac implementation only detects:
```
>>> cube.metadata.dimension_names()
['bands', 't']
```

and `resample_spatial` expects spatial dimensions to work on

### LINKED ISSUE #743: Simpler or more predictable dimension detection with `load_stac`?
We recently had various discussion that working with `load_stac` based cubes can hard because it is not always clear if certain dimensions are present. Which causes confusion down the line, e.g. "can I drop the time dimension?" or "should I re-add this dimension before `merge_cubes`"?

STAC is pretty flexible, and it is not always very clear if a `load_stac`'ed datacube is expected to have a temporal dimension for example. Moreover in openEO, it's also important to know the dimension _names_ to properly handle openEO processes like `reduce_dimension`/`apply_dimension`/..., and this has to be consistent both at client and backend side.

With the STAC [datacube](https://github.com/stac-extensions/datacube) extension it is possible to explicitly list and name dimensions, which seems to be the source of truth for this problem. Note that this extension allows to explicitly omit a dimension, and as such should not be assumed to be present in the data cube.

However, when this datacube extension is not available in the STAC metadata, it seems best, based on discussion with users, to just naively assume a typical 4D EO datacube:
- spatial dimensions "x" and "y"
- temporal dimension "t"
- band dimension "bands"

(based on the [openEO recommendation on openEO collection dimension names](https://github.com/Open-EO/openeo-api/blob/f7afd4909fb7d4bf3ad1041e4543c9b17df63ff9/openapi.yaml#L4498-L4513))

soxofaan:
for #737


soxofaan:
this relates as well to
-  https://github.com/Open-EO/openeo-python-client/issues/743

tests/test_metadata.py
```diff
@@ -937,6 +937,93 @@ def test_metadata_from_stac_temporal_dimension(tmp_path, stac_dict, expected):
...
+    not _PYSTAC_1_9_EXTENSION_INTERFACE,
+    reason="No backport of implementation/test below PySTAC 1.9 extension interface",
+)
+@pytest.mark.parametrize(
+    ["stac_dict", "expected"],
+    [
+        (
+            # Item without cube:dimensions metadata -> assume spatial dimensions x and y
+            StacDummyBuilder.item(),
+            {"x": [None, None], "y": [None, None]},
```
jdries, line 950: if no dimension metadata is present, I would assume a default cube with also temporal and bands dimension

soxofaan: yes indeed, that is the topic of https://github.com/Open-EO/openeo-python-client/issues/743

this PR is just about improving the spatial extent detection if there is STAC metadata about that

soxofaan:
overruled by https://github.com/Open-EO/openeo-python-client/pull/867

PR #750, opened 2025-03-13
feature: artifact-helper

pvbouwel:
Allow uploading artifacts using:
```
from openeo.extra.artifacts import ArtifactHelper

artifact_helper = ArtifactHelper.from_openeo_connection(connection)
storage_uri = artifact_helper.upload_file(object_name, src_file_path)
presigned_uri = artifact_helper.get_presigned_url(storage_uri)
```

Where object_name is a logical string name that will be used to give a unique name in the backend storage.
And src_file_path points to a local file that needs to be stored remotely.

This MVP version still has the sts and s3 URIs hard coded.  This is temporary. It is possible to expand the links that are exposed at the root_url (https://openeo.dataspace.copernicus.eu/openeo/1.2/)  to include rels like 'artifacts-s3-endpoint' and 'artifacts-sts-endpoint'. That way different artifact backends can be chosen based on the OpenEO environment.

### DIFF
--- a/openeo/extra/artifacts/__init__.py
+++ b/openeo/extra/artifacts/__init__.py
@@ -1,0 +1,9 @@
+# Hard coded at this time but this could be a builder that depending on connection builds a client for the Storage
+from openeo.extra.artifacts.internal_s3.artifact_helper import S3ArtifactHelper as ArtifactHelper
+"""
+from openeo.extra.artifacts import ArtifactHelper
+
+artifact_helper = ArtifactHelper.from_openeo_connection(connection)
+storage_uri = artifact_helper.upload_file(object_name, src_file_path)
+presigned_uri = artifact_helper.get_presigned_url(storage_uri)
+"""
--- a/openeo/extra/artifacts/artifact_helper.py
+++ b/openeo/extra/artifacts/artifact_helper.py
@@ -1,0 +1,54 @@
+from __future__ import annotations
+from abc import ABC, abstractmethod
+from typing import Optional
+
+from openeo.extra.artifacts.config import StorageConfig
+from openeo.extra.artifacts.uri import StorageURI
+from openeo.rest.connection import Connection
+from pathlib import Path
+
+
+class ArtifactHelper(ABC):
+    @classmethod
+    def from_openeo_connection(cls, conn: Connection, config: Optional[StorageConfig] = None) -> ArtifactHelper:
+        """
+        Create a new Artifact helper from the OpenEO connection. This is the starting point to upload artifacts
+        """
+        if config is None:
+            config = cls._get_default_storage_config()
+        config.load_openeo_connection_metadata(conn)
+        return cls._from_openeo_connection(conn, config)
+
+    @abstractmethod
+    def upload_file(self, object_name: str, src_file_path: str | Path) -> StorageURI:
+        """
+        A method to store an artifact remotely and get a URI understandable by the OpenEO processor
+        """
+
+    @abstractmethod
+    def get_presigned_url(self, storage_uri: StorageURI, expires_in_seconds: int) -> str:
+        """
+        A method to convert a StorageURI to a signed https URL which can be accessed via normal http libraries.
+
+        These URIs should be kept secret as they provide access to the data.
+        """
+
+    def __init__(self, config: StorageConfig):
+        if not config.is_openeo_connection_metadata_loaded():
+            raise RuntimeError("config should have openeo connection metadata loaded prior to initialization.")
+        self._config = config
+
+    @classmethod
+    @abstractmethod
+    def _get_default_storage_config(cls) -> StorageConfig:
+        """
+        A method that provides a default storage config for the Artifact Helper
+        """
+
+    @classmethod
+    @abstractmethod
+    def _from_openeo_connection(cls, conn: Connection, config: StorageConfig) -> ArtifactHelper:
+        """
+        The implementation that creates an artifact helper. This method takes a config which has already been
+        initialized from the metadata of the OpenEO connection.
+        """
--- a/openeo/extra/artifacts/config.py
+++ b/openeo/extra/artifacts/config.py
@@ -1,0 +1,36 @@
+from __future__ import annotations
+
+from abc import ABC, abstractmethod
+
+from openeo import Connection
+
+_METADATA_LOADED = "_sc_metadata_loaded"
+
+
+class StorageConfig(ABC):
+    """
+    Storage config allows overriding configuration for the interaction with the backend storage.
+    It greatly depends on the type of storage so the enforced API is limited to load metadata using the connection.
+    """
+    @abstractmethod
+    def _load_openeo_connection_metadata(self, conn: Connection) -> None:
+        """
+        Implementations implement their logic of adapting config based on metadata from the current OpenEO connection
+        in this method.
+        """
+
+    def load_openeo_connection_metadata(self, conn: Connection) -> None:
+        """
+        This is the method that is actually used to load metadata. Metadata is only loaded once.
+        """
+        if not self.is_openeo_connection_metadata_loaded():
+            self._load_openeo_connection_metadata(conn)
+            setattr(self, _METADATA_LOADED, True)
+
+    def is_openeo_connection_metadata_loaded(self) -> bool:
+        """
+        This is a helper to check whether metadata is loaded.
+        """
+        if not hasattr(self, _METADATA_LOADED):
+            return False
+        return getattr(self, _METADATA_LOADED)
--- a/openeo/extra/artifacts/internal_s3/artifact_helper.py
+++ b/openeo/extra/artifacts/internal_s3/artifact_helper.py
@@ -1,0 +1,68 @@
+from __future__ import annotations
+
+import datetime
+from typing import TYPE_CHECKING
+from boto3.s3.transfer import TransferConfig
+
+from openeo.extra.artifacts.artifact_helper import ArtifactHelper
+from openeo.extra.artifacts.internal_s3.sts import OpenEOSTSClient
+from openeo.extra.artifacts.internal_s3.config import S3Config
+from openeo.extra.artifacts.internal_s3.model import S3URI
+
+if TYPE_CHECKING:
+    from pathlib import Path
+    from openeo.rest.connection import Connection
+    from openeo.extra.artifacts.internal_s3.model import AWSSTSCredentials
+
+
+class S3ArtifactHelper(ArtifactHelper):
+    BUCKET_NAME = "openeo-artifacts"
+    # From what size will we switch to multi-part-upload
+    MULTIPART_THRESHOLD_IN_MB = 50
+
+    def __init__(self, creds: AWSSTSCredentials, config: S3Config):
+        super().__init__(config)
+        self._creds = creds
+        self.s3 = config.build_client("s3", session_kwargs=creds.as_kwargs())
+    
+    @classmethod
+    def _from_openeo_connection(cls, conn: Connection, config: S3Config) -> S3ArtifactHelper:
+        sts = OpenEOSTSClient(config=config)
+        creds = sts.assume_from_openeo_connection(conn)
+        return S3ArtifactHelper(creds, config=config)
+
+    def _user_prefix(self) -> str:
+        """Each user has its own prefix retrieve it"""
+        return self._creds.get_user_hash()
+    
+    def _get_upload_prefix(self) -> str:
+        return f"{self._user_prefix()}/{datetime.datetime.now(datetime.UTC).strftime('%Y/%m/%d')}/"
+    
+    def _get_upload_key(self, object_name: str) -> str:
+        return f"{self._get_upload_prefix()}{object_name}"
+    
+    def upload_file(self, object_name: str, src_file_path: str | Path) -> S3URI:
+        mb = 1024 ** 2
+        config = TransferConfig(multipart_threshold=self.MULTIPART_THRESHOLD_IN_MB * mb)
+        bucket = self.BUCKET_NAME
+        key = self._get_upload_key(object_name)
+        self.s3.upload_file(
+            str(src_file_path),
+            bucket,
+            key,
+            Config=config
+        )
+        return S3URI(bucket, key)
+    
+    def get_presigned_url(self, storage_uri: S3URI, expires_in_seconds: int = 7 * 3600 * 24) -> str:
+        url = self.s3.generate_presigned_url(
+            'get_object',
+            Params={'Bucket': storage_uri.bucket, 'Key': storage_uri.key},
+            ExpiresIn=expires_in_seconds
+        )
+        assert isinstance(self._config, S3Config)
+        return self._config.add_trace_id_qp_if_needed(url)
+
+    @classmethod
+    def _get_default_storage_config(cls) -> S3Config:
+        return S3Config()
--- a/openeo/extra/artifacts/internal_s3/config.py
+++ b/openeo/extra/artifacts/internal_s3/config.py
@@ -1,0 +1,112 @@
+from dataclasses import dataclass
+import boto3
+import botocore
+
+from openeo import Connection
+from openeo.extra.artifacts.internal_s3.tracer import add_trace_id, add_trace_id_as_query_parameter
+from openeo.extra.artifacts.config import StorageConfig
+from botocore.config import Config
+from typing import Optional
+from packaging.version import Version
+
+
+if Version(botocore.__version__) < Version("1.36.0"):
+    # Before 1.36 checksuming was not done by default anyway and therefore
+    # there was no opt-out.
+    no_default_checksum_cfg = Config()
+else:
+    no_default_checksum_cfg = Config(
+        request_checksum_calculation='when_required',
+    )
+
+
+@dataclass
+class S3Config(StorageConfig):
+    """The s3 endpoint url protocol:://fqdn[:portnumber]"""
+    s3_endpoint_url: Optional[str] = None
+    """The sts endpoint url protocol:://fqdn[:portnumber]"""
+    sts_endpoint_url: Optional[str] = None
+    """The trace_id is if you want to send a uuid4 identifier to the backend"""
+    trace_id: str = ""
+    """You can change the botocore_config used but this is an expert option"""
+    botocore_config: Optional[Config] = None
+    """The role ARN to be assumed"""
+    sts_role_arn: Optional[str] = None
+
+    def _load_openeo_connection_metadata(self, conn: Connection) -> None:
+        """
+        Hard coding since connection does not allow automatic determining config yet.
+        """
+        if self.s3_endpoint_url is None:
+            self.s3_endpoint_url = "https://s3.waw3-1.openeo.v1.dataspace.copernicus.eu"
+
+        if self.sts_endpoint_url is None:
+            self.sts_endpoint_url = "https://sts.waw3-1.openeo.v1.dataspace.copernicus.eu"
+
+        if self.sts_role_arn is None:
+            self.sts_role_arn = "arn:aws:iam::000000000000:role/S3Access"
+
+    def __post_init__(self):
+        self.botocore_config = self.botocore_config or no_default_checksum_cfg
+
+    def build_client(self, service_name: str, session_kwargs: Optional[dict] = None):
+        """
+        Build a boto3 client for an OpenEO service provider.
+
+        service_name is the service you want to consume: s3|sts
+        session_kwargs: a dictionary with keyword arguments that will be passed when creating the boto session
+        """
+        session_kwargs = session_kwargs or {}
+        session = boto3.Session(region_name=self._get_storage_region(), **session_kwargs)
+        client = session.client(
+            service_name, 
+            endpoint_url=self._get_endpoint_url(service_name),
+            config=self.botocore_config,   
+        )
+        if self.trace_id != "":
+            add_trace_id(client, self.trace_id)
+        return client
+
+    @staticmethod
+    def _remove_protocol_from_uri(uri: str):
+        uri_separator = "://"
+        idx = uri.find(uri_separator)
+        if idx < 0:
+            raise ValueError("_remove_protocol_from_uri must be of form protocol://...")
+        return uri[idx+len(uri_separator):]
+    
+    def _get_storage_region(self) -> str:
+        """
+        S3 URIs follow the convention detailed on https://docs.aws.amazon.com/general/latest/gr/s3.html
+        """
+        s3_names = ["s3", "s3-fips"]
+        reserved_words = ["dualstack", "prod", "stag", "dev"]
+        s3_endpoint_parts = self._remove_protocol_from_uri(self.s3_endpoint_url).split(".")
+        for s3_name in s3_names:
+            try:
+                old_idx = s3_endpoint_parts.index(s3_name)
+                idx = old_idx + 1
+                while idx != old_idx:
+                    old_idx = idx
+                    for reserved_word in reserved_words:
+                        if s3_endpoint_parts[idx] in reserved_word:
+                            idx += 1
+                return s3_endpoint_parts[idx]
+            except ValueError:
+                continue
+        raise ValueError(f"Cannot determine region from {self.s3_endpoint_url}")
+
+    def _get_endpoint_url(self, service_name: str) -> str:
+        if service_name == "s3":
+            return self.s3_endpoint_url
+        elif service_name == "sts":
+            return self.sts_endpoint_url
+        raise ValueError(f"Unsupported service {service_name}")
+
+    def add_trace_id_qp_if_needed(self, url: str) -> str:
+        if self.trace_id == "":
+            return url
+        return add_trace_id_as_query_parameter(url, self.trace_id)
+
+    def get_sts_role_arn(self) -> str:
+        return self.sts_role_arn
--- a/openeo/extra/artifacts/internal_s3/model.py
+++ b/openeo/extra/artifacts/internal_s3/model.py
@@ -1,0 +1,56 @@
+from __future__ import annotations
+import hashlib
+from dataclasses import dataclass
+from openeo.extra.artifacts.uri import StorageURI
+
+
+@dataclass(frozen=True)
+class AWSSTSCredentials:
+    AWS_ACCESS_KEY_ID: str
+    AWS_SECRET_ACCESS_KEY: str
+    AWS_SESSION_TOKEN: str
+    subject_from_web_identity_token: str
+
+    @classmethod
+    def from_assume_role_response(cls, resp: dict) -> AWSSTSCredentials:
+        d = resp["Credentials"]
+        return AWSSTSCredentials(
+            AWS_ACCESS_KEY_ID=d["AccessKeyId"],
+            AWS_SECRET_ACCESS_KEY=d["SecretAccessKey"],
+            AWS_SESSION_TOKEN=d["SessionToken"],
+            subject_from_web_identity_token=resp["SubjectFromWebIdentityToken"]
+        )
+
+    def as_kwargs(self) -> dict:
+        return {
+            "aws_access_key_id": self.AWS_ACCESS_KEY_ID,
+            "aws_secret_access_key": self.AWS_SECRET_ACCESS_KEY,
+            "aws_session_token": self.AWS_SESSION_TOKEN
+        }
+
+    def get_user_hash(self) -> str:
+        hash_object = hashlib.sha1(self.subject_from_web_identity_token.encode())
+        return hash_object.hexdigest()
+
+
+@dataclass(frozen=True)
+class S3URI(StorageURI):
+    bucket: str
+    key: str
+
+    @classmethod
+    def from_str(cls, uri: str) -> S3URI:
+        s3_prefix = "s3://"
+        if uri.startswith(s3_prefix):
+            without_prefix = uri[len(s3_prefix):]
+            without_prefix_parts = without_prefix.split("/")
+            bucket = without_prefix_parts[0]
+            if len(without_prefix_parts) == 1:
+                return S3URI(bucket, "")
+            else:
+                return S3URI(bucket, "/".join(without_prefix_parts[1:]))
+        else:
+            raise ValueError(f"Input {uri} is not a valid S3 URI should be of form s3://<bucket>/<key>")
+
+    def __str__(self):
+        return f"s3://{self.bucket}/{self.key}"
--- a/openeo/extra/artifacts/internal_s3/sts.py
+++ b/openeo/extra/artifacts/internal_s3/sts.py
@@ -1,0 +1,37 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+from openeo.extra.artifacts.internal_s3.model import AWSSTSCredentials
+from openeo.extra.artifacts.internal_s3.config import S3Config
+from openeo.rest.auth.auth import BearerAuth
+
+if TYPE_CHECKING:
+    from openeo.rest.connection import Connection
+
+
+class OpenEOSTSClient:
+    def __init__(self, config: S3Config):
+        self.config = config
+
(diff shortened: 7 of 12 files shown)


soxofaan (commented):
this is quite a large diff :smile:  and I haven't gone deep yet, but here are some initial notes

openeo/extra/artifacts/internal_s3/artifact_helper.py
```diff
@@ -0,0 +1,68 @@
...
+        """Each user has its own prefix retrieve it"""
+        return self._creds.get_user_hash()
+    
+    def _get_upload_prefix(self) -> str:
+        return f"{self._user_prefix()}/{datetime.datetime.now(datetime.UTC).strftime('%Y/%m/%d')}/"
+    
+    def _get_upload_key(self, object_name: str) -> str:
+        return f"{self._get_upload_prefix()}{object_name}"
+    
+    def upload_file(self, object_name: str, src_file_path: str | Path) -> S3URI:
```
soxofaan, line 44: Isn't it more logical to put `src_file_path` argument before `object_name` (like bash `cp src dst`).
It would also allow to make `object_name` optional  provide a autogenerated name.

minor note: `src_file_path` looks a bit long as name, could be just `path`, `source` or `src` I think

openeo/extra/artifacts/internal_s3/artifact_helper.py
```diff
@@ -0,0 +1,68 @@
...
+import datetime
+from typing import TYPE_CHECKING
+from boto3.s3.transfer import TransferConfig
+
+from openeo.extra.artifacts.artifact_helper import ArtifactHelper
+from openeo.extra.artifacts.internal_s3.sts import OpenEOSTSClient
+from openeo.extra.artifacts.internal_s3.config import S3Config
+from openeo.extra.artifacts.internal_s3.model import S3URI
+
+if TYPE_CHECKING:
```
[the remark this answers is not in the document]
pvbouwel: Will fix cleanup because this guard indeed does not solve any problem.

openeo/extra/artifacts/internal_s3/config.py
```diff
@@ -0,0 +1,112 @@
...
+
+from openeo import Connection
+from openeo.extra.artifacts.internal_s3.tracer import add_trace_id, add_trace_id_as_query_parameter
+from openeo.extra.artifacts.config import StorageConfig
+from botocore.config import Config
+from typing import Optional
+from packaging.version import Version
+
+
+if Version(botocore.__version__) < Version("1.36.0"):
```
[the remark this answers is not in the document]
pvbouwel: Since 1.36.0 is only from 2025 I prefer to not force it onto users but it makes sense to not take another dependency so will refactor to use ComparableVersion

openeo/extra/artifacts/internal_s3/config.py
```diff
@@ -0,0 +1,112 @@
...
+    botocore_config: Optional[Config] = None
+    """The role ARN to be assumed"""
+    sts_role_arn: Optional[str] = None
+
+    def _load_openeo_connection_metadata(self, conn: Connection) -> None:
+        """
+        Hard coding since connection does not allow automatic determining config yet.
+        """
+        if self.s3_endpoint_url is None:
+            self.s3_endpoint_url = "https://s3.waw3-1.openeo.v1.dataspace.copernicus.eu"
```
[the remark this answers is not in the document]
pvbouwel: I agree the hardcoding is not preferred and it is meant as a temporary stop gap.
I don't think there is a proper default value hence the None of the dataclass seemed sensible.

The reason for this boiler plate style code is because it can then be easily replaced with a request to the openeo_URL to check metadata of the backend and to configure based on that.

openeo/extra/artifacts/internal_s3/model.py
```diff
@@ -0,0 +1,56 @@
...
+import hashlib
+from dataclasses import dataclass
+from openeo.extra.artifacts.uri import StorageURI
+
+
+@dataclass(frozen=True)
+class AWSSTSCredentials:
+    AWS_ACCESS_KEY_ID: str
+    AWS_SECRET_ACCESS_KEY: str
+    AWS_SESSION_TOKEN: str
```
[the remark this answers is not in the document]
pvbouwel: It is left over but from functionality that is not there so there is no reason to not follow Python styling. I'll refactor.

openeo/extra/artifacts/internal_s3/model.py
```diff
@@ -0,0 +1,56 @@
...
+@dataclass(frozen=True)
+class S3URI(StorageURI):
+    bucket: str
+    key: str
+
+    @classmethod
+    def from_str(cls, uri: str) -> S3URI:
+        s3_prefix = "s3://"
+        if uri.startswith(s3_prefix):
+            without_prefix = uri[len(s3_prefix):]
```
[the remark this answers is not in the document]
pvbouwel: refactored to use urlparse

openeo/extra/artifacts/__init__.py
```diff
@@ -0,0 +1,9 @@
+# Hard coded at this time but this could be a builder that depending on connection builds a client for the Storage
+from openeo.extra.artifacts.internal_s3.artifact_helper import S3ArtifactHelper as ArtifactHelper
```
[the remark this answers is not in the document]
pvbouwel: Well I wanted users to use ArtifactHelper rather than S3ArtifactHelper as it is meant as a generic builder that can be extended when needed. I was not aware of how it would get processed by the sphinx doc system.

Would it be better to already have the builder eventhough it has a dummy implementation for now? 
https://github.com/Open-EO/openeo-python-client/pull/750/commits/a3eea6214d998dff461f39e6ec12c1fe0fb79969

openeo/extra/artifacts/__init__.py
```diff
@@ -0,0 +1,9 @@
...
+"""
+from openeo.extra.artifacts import ArtifactHelper
+
+artifact_helper = ArtifactHelper.from_openeo_connection(connection)
+storage_uri = artifact_helper.upload_file(object_name, src_file_path)
+presigned_uri = artifact_helper.get_presigned_url(storage_uri)
+"""
```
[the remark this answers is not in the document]
pvbouwel: It was documentation but moved it now with addressing other comment

openeo/extra/artifacts/config.py
```diff
@@ -0,0 +1,36 @@
...
+            self._load_openeo_connection_metadata(conn)
+            setattr(self, _METADATA_LOADED, True)
+
+    def is_openeo_connection_metadata_loaded(self) -> bool:
+        """
+        This is a helper to check whether metadata is loaded.
+        """
+        if not hasattr(self, _METADATA_LOADED):
+            return False
+        return getattr(self, _METADATA_LOADED)
```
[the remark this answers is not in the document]
pvbouwel: This avoids the need of initialization which makes it nicely extendable by dataclasses.

openeo/extra/artifacts/artifact_helper.py
```diff
@@ -0,0 +1,54 @@
...
+
+    @abstractmethod
+    def get_presigned_url(self, storage_uri: StorageURI, expires_in_seconds: int) -> str:
+        """
+        A method to convert a StorageURI to a signed https URL which can be accessed via normal http libraries.
+
+        These URIs should be kept secret as they provide access to the data.
+        """
+
+    def __init__(self, config: StorageConfig):
```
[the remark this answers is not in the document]
pvbouwel: The idea was to have important methods that users are aware of first. Users should never call this init so hiding it was a concious decission. If you still prefer it to be on top I am open to move it but personally I like it being hidden.

openeo/extra/artifacts/internal_s3/__init__.py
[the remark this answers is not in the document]
pvbouwel: I wasn't aware same convention was used for modules

openeo/extra/artifacts/internal_s3/config.py
```diff
@@ -0,0 +1,112 @@
...
+            endpoint_url=self._get_endpoint_url(service_name),
+            config=self.botocore_config,   
+        )
+        if self.trace_id != "":
+            add_trace_id(client, self.trace_id)
+        return client
+
+    @staticmethod
+    def _remove_protocol_from_uri(uri: str):
+        uri_separator = "://"
```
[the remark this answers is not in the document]
pvbouwel: For the other comment I moved to urlparse as I indeed needed real manipulation of the URI. For this one however I just need to strip the protocol. I don't see how full parsing of the URL will help to simplify. The problem with parsing is that it is easy to miss edge cases and have query parameters dropped or other special behavior due to certain characters. By just keeping the string opaque no special handling is needed.

openeo/extra/artifacts/internal_s3/sts.py
```diff
@@ -0,0 +1,37 @@
...
+        self.config = config
+
+    def assume_from_openeo_connection(self, conn: Connection) -> AWSSTSCredentials:
+        """
+        Takes an OpenEO connection object and returns temporary credentials to interact with S3
+        """
+        auth = conn.auth
+        assert auth is not None
+        if not isinstance(auth, BearerAuth):
+            raise ValueError("Only connections that use federation are allowed.")
```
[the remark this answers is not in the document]
pvbouwel: I'll rephrase it.

openeo/extra/artifacts/internal_s3/sts.py
```diff
@@ -0,0 +1,37 @@
...
+        auth_token = auth.bearer.split('/')
+        sts = self.config.build_client("sts")
+
+        return AWSSTSCredentials.from_assume_role_response(
+            sts.assume_role_with_web_identity(
+                RoleArn=self._get_aws_access_role(),
+                RoleSessionName=auth_token[1],
+                WebIdentityToken=auth_token[2],
```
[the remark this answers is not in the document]
pvbouwel: Support for others can be added when there is a need currently JWT is the only one we support but in the end the only important part is that a token is verifiable .

The RoleSessionName is not really that important to the authentication process but it would make sense to use a more useful value. I will adapt to  f"artifact-helper-{datetime.datetime.now(datetime.UTC).strftime('%Y%m%d%H%M%S')}"

openeo/extra/artifacts/uri.py
```diff
@@ -0,0 +1,15 @@
...
+
+class StorageURI(ABC):
+    """A URI that is specific to a storage backend. The protocol determines what this URL looks like"""
+    @classmethod
+    @abstractmethod
+    def from_str(cls, uri: str) -> StorageURI:
+        """factory method to create a typed object from its string representation"""
+
+    @abstractmethod
+    def __str__(self):
```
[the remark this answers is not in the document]
pvbouwel: Make sense went for `to_string`

openeo/extra/artifacts/internal_s3/artifact_helper.py
pvbouwel: Yes makes a lot of sense as object name naturally can be the filename by default and only special cases would want to chose the target object name.

Incorporated feedback including minor note.

openeo/extra/artifacts/internal_s3/config.py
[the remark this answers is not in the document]
pvbouwel: Config is now fetched using the connection url.

pvbouwel:
@soxofaan provided docs as requested. I did move around code but mostly renames of files to start with _ to discourage users from entering.

Other changes:

- introduced feature_flags for s3sts such that backend can advertise what is supports
- now decided on advertised feature flag whether to enable/disable checksumming
- more tests
- Added JSON lexer for sphinx as JSON would not render properly otherwise

docs/conf.py
```diff
@@ -36,15 +38,16 @@
...
-    'sphinx.ext.viewcode',
-    'sphinx.ext.doctest',
-    'myst_parser',
-    "sphinx.ext.intersphinx",
+    "sphinx.ext.autodoc",
+    "sphinx_autodoc_typehints",
+    "sphinx.ext.viewcode",
+    "sphinx.ext.doctest",
+    "myst_parser",
+    "jsonlexer",
```
soxofaan: is it intentional to remove "sphinx.ext.intersphinx"  here?

docs/conf.py
```diff
@@ -195,13 +198,3 @@
...
-
-
-# Mapping for external documentation
-intersphinx_mapping = {
-    "python": ("https://docs.python.org/3", None),
-    "numpy": ("https://numpy.org/doc/stable/", None),
-    "xarray": ("https://docs.xarray.dev/en/stable/", None),
-    "pandas": ("https://pandas.pydata.org/pandas-docs/stable/", None),
-    "urllib3": ("https://urllib3.readthedocs.io/en/stable/", None),
-}
```
soxofaan: is it intentional to remove this intersphinx_mapping ?

docs/api-artifacts.rst
```diff
@@ -0,0 +1,186 @@
...
+    This is a new experimental API, subject to change.
+
+    The artifacts functionality relies on extra Python packages. They can be installed using:
+
+    .. code-block:: shell
+
+     pip install "openeo[artifacts]" --upgrade
+
+
+When running OpenEO jobs it is not uncommon to require artifacts that should be accessible during job execution. This
```
soxofaan: FYI: while I'm not a big fan of it myself, the official name of "openEO" is with lower case  "o" in beginning.  The docs should stick to that

openeo/extra/artifacts/_s3sts/artifact_helper.py
```diff
@@ -0,0 +1,105 @@
...
+
+    def __init__(self, conn: Connection, config: S3STSConfig):
+        super().__init__(config)
+        self.conn = conn
+        self.config = config
+        self._creds = self.get_new_creds()
+        self._s3: Optional[S3Client] = None
+
+    @classmethod
+    def _from_openeo_connection(cls, conn: Connection, config: S3STSConfig) -> S3STSArtifactHelper:
```
soxofaan: FYI: we typically use "connection" (instead of abbreviation) as argument name for where we have to pass connection objects
```suggestion
    def _from_openeo_connection(cls, connection: Connection, config: S3STSConfig) -> S3STSArtifactHelper:
```

docs/api-artifacts.rst
```diff
@@ -0,0 +1,186 @@
...
+
+.. code-block:: python
+
+    import openeo
+    from openeo.extra.artifacts import ArtifactHelper
+
+    connection = openeo.connect("my-openeo.prod.example").authenticate_oidc()
+
+    artifact_helper = ArtifactHelper.from_openeo_connection(connection)
+    storage_uri = artifact_helper.upload_file(object_name, src_file_path)
```
soxofaan: This was changed, right?
```suggestion
    storage_uri = artifact_helper.upload_file(src_path, object_name)
```

docs/cookbook/sampling.md
```diff
@@ -6,10 +6,10 @@ but rather want to extract a result at specific locations.
 Examples include extracting training data for model calibration, or computing the result for
 areas where validation data is available.
 
-An important constraint is that most implementations assume that sampling is an operation 
-on relatively small areas, of for instance up to 512x512 pixels (but often much smaller). 
+An important constraint is that most implementations assume that sampling is an operation
```
soxofaan: I'm a bit confused about why changes to this file are included in this PR. Because this is not related to this feature request right? Is this intended?

docs/_ext/jsonlexer.py
```diff
@@ -0,0 +1,5 @@
+from pygments.lexers import JsonLexer
+
+
+def setup(app):
+    app.add_lexer("json", JsonLexer)
```
soxofaan: I'm a bit curious about the need to do this from this PR, is the support for JSON in the docs that broken/poor at the moment? I'll see if I can fix this up directly in master

openeo/extra/artifacts/artifact_helper.py
```diff
@@ -0,0 +1,48 @@
...
+from openeo.extra.artifacts._s3sts.config import S3STSConfig
+from openeo.extra.artifacts.exceptions import UnsupportedArtifactsType
+
+config_to_helper: Dict[Type[ArtifactsStorageConfigABC], Type[ArtifactHelperABC]] = {S3STSConfig: S3STSArtifactHelper}
+config_type_to_helper: Dict[str, Type[ArtifactHelperABC]] = {
+    ArtifactsStorageConfigABC.get_type_from(cfg): helper for cfg, helper in config_to_helper.items()
+}
+
+
+class ArtifactHelper(ArtifactHelperBuilderABC):
```
soxofaan: To be honest I'm getting a bit confused by these helper and helperbuilder class hierarchies. 

 there is ArtifactHelperBuilderABC and ArtifactHelperABC in `_artifact_helper_abc.py`

And here you have ArtifactHelper  which is subclass of ArtifactHelper**Builder**ABC  (instead of ArtifactHelperABC). Or should this call be called ArtifactHelper**Builder**?

soxofaan: While I understand the need for a  hierarchy of helper classes, 
I'm not sure I understand the need for a hierarchy for helperbuilder classes.
Wouldn't you just need a single helper builder endpoint (e.g. a single function) instead of a whole hierarchy?

openeo/extra/artifacts/_s3sts/artifact_helper.py
soxofaan: this applies to other places in the PR too I think

docs/_ext/jsonlexer.py
soxofaan: Inspired by this bit of the PR I just tried with 98fae3d62f4c8fa8e51eb3ffe17ebf47990f9e87  to use `.. code-block:: json`  in docs/udp.rst  and that seems to work just fine as far as I see on https://open-eo.github.io/openeo-python-client/udp.html

setup.py
```diff
@@ -72,26 +78,27 @@
...
         "requests>=2.26.0",
-        "urllib3>=1.9.0",
         "shapely>=1.6.4",
         "numpy>=1.17.0",
         "xarray>=0.12.3,<2025.01.2",  # TODO #721 xarray non-nanosecond support
         "pandas>0.20.0",
         # TODO #578: pystac 1.5.0 is highest version available for lowest Python version we still support (3.7).
-        "pystac>=1.5.0",
+        # TODO #715 compatibility with pystac 1.12
+        "pystac>=1.5.0,<1.12",
```
soxofaan: This revert of 4311cad8b84a4f10631d038a35c1f201f7e48b4b is not intended I guess?

setup.py
```diff
@@ -72,26 +78,27 @@
     test_suite="tests",
     install_requires=[
         "requests>=2.26.0",
-        "urllib3>=1.9.0",
```
soxofaan: this removal of urllib dependency is not intended I guess?

setup.py
```diff
@@ -34,13 +34,15 @@
...
     "pystac-client>=0.7.5",
+    "moto>=5.0.0",
 ]
 
 docs_require = [
     "sphinx",
     "sphinx-autodoc-annotation",
     "sphinx-autodoc-typehints>=2.2.3",
     "myst-parser",
+    "pygments",  # Indirect dependency but we call it explicitly for JSON lexer
```
soxofaan: as mentioned: if this extra JSON lexer stuff is not necessary, this explicit dependency can be removed again

docs/_ext/jsonlexer.py
pvbouwel: I am no longer sure. I recall it wasn´t rendering the JSON at all but now that I clean it up again it keeps on working so will remove the manually added lexer.

docs/api-artifacts.rst
pvbouwel: changed occurrences.

docs/api-artifacts.rst
pvbouwel: correct, altered.

docs/conf.py
pvbouwel: No, undone as I reverted jsonlexer now again.

docs/conf.py
pvbouwel: No, not really sure how it got removed, undoing it.

docs/cookbook/sampling.md
pvbouwel: Was not meant to be staged. It is a fix for a broken link and formatter then took care of trailing spaces but I'll get that change out of this PR.

openeo/extra/artifacts/_s3sts/artifact_helper.py
pvbouwel: Went through all the conns and changed to connection. For class attributes went for internal attribute

setup.py
pvbouwel: removed again

setup.py
pvbouwel: Indeed not.

setup.py
pvbouwel: nu re-reverted

openeo/extra/artifacts/artifact_helper.py
pvbouwel: I think you are right. I believe this stems from when it was unclear to use the capabilities and to allow multiple builder implementations but this is adding unnecessary complexity early on. As you pointed out in he first comment the name is also not the greatest adding confusion into the mix so a 1-function builder endpoint sounds much more appropriate. I replaced the logic with a `build_artifact_helper` as `build_` is used already in other places of the code base.

docs/api-artifacts.rst
```diff
@@ -0,0 +1,185 @@
+.. _api-openeo-extra-artifacts:
+
+====================================
+API: openeo.extra.artifacts
+====================================
+
+.. warning::
+    This is a new experimental API, subject to change.
+
+    The artifacts functionality relies on extra Python packages. They can be installed using:
```
soxofaan, line 13: > relies on extra Python packages

this should not be part of the warning block, I think this can be a separate `.. important::` block

docs/api-artifacts.rst
```diff
@@ -0,0 +1,185 @@
...
+
+
+When running openEO jobs it is not uncommon to require artifacts that should be accessible during job execution. This
+requires the artifacts to be accessible from within the openEO processing environment. :py:mod:`openeo.extra.artifacts` tries
+to perform the heavy lifting for this use case by allowing staging artifacts to a secure but temporary location using 3
+simple steps:
+
+1. Connect to your openEO backend
+2. Create an artifact helper from your openEO connection
+3. Upload your file using the artifact helper and optionally get a presigned URI
```
soxofaan, line 27: there should a note here as well that the involved openEO backend must expose some additional metadata in it's capabilities doc to make that possible.
Ideally we point to an official [API extension](https://github.com/Open-EO/openeo-api/tree/master/extensions) (which can be work in progress)

docs/api-artifacts.rst
```diff
@@ -0,0 +1,185 @@
...
+    import openeo
+    from openeo.extra.artifacts import build_artifact_helper
+
+    connection = openeo.connect("my-openeo.prod.example").authenticate_oidc()
+
+    artifact_helper = build_artifact_helper(connection)
+    storage_uri = artifact_helper.upload_file(path, object_name)
+    presigned_uri = artifact_helper.get_presigned_url(storage_uri)
+
+Note that the storage_uri can be used from regular execution steps of your openEO job. The presigned uri could be used
```
soxofaan: > regular execution steps

maybe give examples of this: would that be `load_stac` or `load_geojson`?

docs/api-artifacts.rst
```diff
@@ -0,0 +1,185 @@
...
+
+.. autoclass:: openeo.extra.artifacts._artifact_helper_abc.ArtifactHelperABC
+    :members: upload_file, get_presigned_url
+    :no-index:
+
+
+How does it work ?
+==================
+
+1) :py:meth:`openeo.extra.artifacts.artifact_helper.ArtifactHelper.from_openeo_connection` is a factory method that
```
soxofaan: this has become `build_artifact_helper` right?

docs/api-artifacts.rst
```diff
@@ -0,0 +1,185 @@
...
+   an Object storage that has an S3 API it will:
+
+   1. Get temporary S3 credentials based on config advertised by the backend and the session from your connection
+   2. Upload the file into object storage and return an S3 URI which the backend can resolve
+   3. Optional the :py:meth:`openeo.extra.artifacts._artifact_helper_abc.ArtifactHelperABC.get_presigned_url` makes a
+      URI signed with the temporary credentials such that it works standalone (Some tools and execution steps do not
+      support handling of internal references. presigned URLs should work in any tool).
+
+
+.. _artifacts-exceptions:
```
soxofaan: this label is out of place I guess

docs/api-artifacts.rst
```diff
@@ -0,0 +1,185 @@
...
+
+.. code-block:: json
+
+    {
+       // ...
+       "artifacts": {
+         "providers": [
+           {
+             "config": {  // The config block its keys can differ for other config types
+               "bucket": "openeo-artifacts",                      // The bucket where the artifacts will be stored
```
soxofaan: these comments that far on the right will display poorly in the docs, there is typically not much horizontal space.
I would try to stay below, say, 100 chars per line

docs/api-artifacts.rst
```diff
@@ -0,0 +1,185 @@
...
+               "role": "arn:aws:iam::000000000000:role/S3Access", // The role that will be assumed via STS
+               "s3_endpoint": "https://my.s3.test",               // Where S3 API calls are sent
+               "sts_endpoint": "https://my.sts.test"              // Where STS API calls are sent
+             },
+             "id": "s3",                                          // This id is a logical name
+             "type": "S3STSConfig"                                // The config type of the ArtifactHelper
```
soxofaan: put id and type above config?

openeo/extra/artifacts/_backend.py
```diff
@@ -0,0 +1,114 @@
...
+        except KeyError as e:
+            raise NoAdvertisedProviders() from e
+
+    def get_preferred_artifacts_provider(self) -> ProviderConfig:
+        try:
+            return ProviderConfig.from_typed_dict(self._get_artifacts_providers()[0])
+        except IndexError:
+            return ProviderConfig("id", "type", {}, exc=NoAdvertisedProviders())
+        except ArtifactsException as e:
+            return ProviderConfig("id", "type", {}, exc=e)
```
soxofaan: is it intentional to use id "id" and type "type" here? Or is that still placeholder code?

if there is no known id/type, I'd find it more natural to use `None` or a string like "n/a" or "unknown"

openeo/extra/artifacts/_artifact_helper_abc.py
```diff
@@ -0,0 +1,95 @@
...
+from openeo.extra.artifacts._uri import StorageURI
+from openeo.rest.connection import Connection
+
+
+class ArtifactHelperABC(ABC):
+    """
+    This class defines the *interface* that an artifact helper should implement and support. This is used by OpenEO users
+    willing to manage artifacts.
+
+    Instances that implement it get created by the `openeo.extra.artifacts.artifact_helper.ArtifactHelper` factory
```
soxofaan: `ArtifactHelper` is not a thing anymore, right?

openeo/extra/artifacts/exceptions.py
```diff
@@ -0,0 +1,61 @@
...
+    is not supported by the backend.
+
+    :parameter type_id: The type_id that is not supported but which was attempted to be used
+    """
+
+    def __init__(self, type_id: str):
+        self.type_id = type_id
+
+    def __str__(self):
+        return f"The OpenEO backend does not support {self.type_id}"
```
soxofaan: I'm not sure about the pattern to override `__str__` in exception subclasses. Often, exceptions are `repr`ed instead of `str`ed, and then you won't get that error message.

I think it's more robust and simpler, to do everything in `__init__`, e.g.

```python
class UnsupportedArtifactsType(ArtifactsException):
    def __init__(self, type_id: str):
        super().__init__(f"The OpenEO backend does not support {type_id}")
```

(if you would still want to be able to recover the `type_id` from the exception, you can still attach it as attribute in `__init__` of course, but I left that out for clarity)

docs/api-artifacts.rst
pvbouwel: Added a link to the documentation on the additional metadata. I'd rather keep the work of defining the API extension (even a work-in-progress) outside of this PR as to not keep on delaying this.

docs/api-artifacts.rst
pvbouwel: Yes, missed during refactoring I was a bit careless since Sphinx did no longer give red warnings but it seems it does not for broken references to python methods. Changed it now

docs/api-artifacts.rst
pvbouwel: And no longer needed anyway so cleaned it up.

docs/api-artifacts.rst
pvbouwel: Make sense, moved the comments above.

docs/api-artifacts.rst
pvbouwel: Yes, that is clearer

openeo/extra/artifacts/_artifact_helper_abc.py
pvbouwel: correct, replaced it with build function.

openeo/extra/artifacts/_backend.py
pvbouwel: I like the "n/a" suggestion so I went for that.

openeo/extra/artifacts/exceptions.py
pvbouwel: Makes sense

docs/api-artifacts.rst
pvbouwel: It does seem that at time of writing storage_uri's cannot be used by themselves yet. I have updated the doc to explain the presigned_uri should be used.

docs/api-artifacts.rst
soxofaan: Sure, this PR should not be blocked by the state of the specification of the extension, but we should bootstrap that process to avoid hardcoding proprietary stuff in a generic client

docs/api-artifacts.rst
soxofaan: yeah when doing incremental builds during development, sphinx sometimes misses dangling pointers

docs/api-artifacts.rst
soxofaan: bootstrapped it here:
- https://github.com/Open-EO/openeo-api/issues/566

soxofaan: it might be important to settle some basics of the spec, e.g. if "artifacts" as toplevel key in the capabilities doc isn't too generic as a name

soxofaan:
I would also add a changelog entry along the lines of

```
- Initial/experimental implementation of artifact upload helper ([Open-EO/openeo-api#566](https://github.com/Open-EO/openeo-api/issues/566))
```

docs/api-artifacts.rst
```diff
@@ -0,0 +1,200 @@
+.. _api-openeo-extra-artifacts:
+
+====================================
+API: openeo.extra.artifacts
+====================================
+
+.. warning::
+    This is a new experimental API, subject to change.
+
```
soxofaan, line 10: also add this note:

```

.. versionadded:: 0.44.0

```

(assuming that will be the release it will be added)

pvbouwel: Squashed current state and added the version + changelog as a separate commit

docs/api-artifacts.rst
```diff
@@ -0,0 +1,116 @@
+.. _api-openeo-extra-artifacts:
+
+====================================
+API: openeo.extra.artifacts
+====================================
+.. versionadded:: 0.44.0
```
soxofaan:
```suggestion
.. versionadded:: 0.45.0
```

(0.44 was already released in mean time)

soxofaan (commented):
I'd merge latest master in this feature branch

PR #754, opened 2025-03-24
#259 normalize "EPSG:123"-style CRS to integer

soxofaan:
in bbox `spatial_extent` of `load_collection` and related

leftover for #259


openeo/rest/datacube.py
```diff
@@ -3043,6 +3043,8 @@ def _get_geometry_argument(
         and isinstance(argument, dict)
         and all(k in argument for k in ["west", "south", "east", "north"])
     ):
+        if "crs" in argument:
```
soxofaan, line 3046: just use `to_bbox_dict` (or related) here instead of reinventing that wheel?

PR #755, opened 2025-03-31
made load_stac nicer to mismatch in band names

VictorVerhaert:
Note: warning message could be made clearer but out of inspiration

### DIFF
--- a/openeo/metadata.py
+++ b/openeo/metadata.py
@@ -375,22 +375,20 @@ class CubeMetadata:
         :return: Updated metadata
         """
         self.assert_valid_dimension(dimension)
-        loc = self.dimension_names().index(dimension)
-        new_dimensions = self._dimensions.copy()
-        new_dimensions[loc] = new_dimensions[loc].rename_labels(target, source)
-
-        return self._clone_and_update(dimensions=new_dimensions)
+        return self._clone_and_update(
+            dimensions=[
+                d.rename_labels(target=target, source=source) if d.name == dimension else d for d in self._dimensions
+            ]
+        )
 
     def rename_dimension(self, source: str, target: str) -> CubeMetadata:
         """
         Rename source dimension into target, preserving other properties
         """
         self.assert_valid_dimension(source)
-        loc = self.dimension_names().index(source)
-        new_dimensions = self._dimensions.copy()
-        new_dimensions[loc] = new_dimensions[loc].rename(target)
-
-        return self._clone_and_update(dimensions=new_dimensions)
+        return self._clone_and_update(
+            dimensions=[d.rename(name=target) if d.name == source else d for d in self._dimensions]
+        )
 
     def reduce_dimension(self, dimension_name: str) -> CubeMetadata:
         """Create new CubeMetadata object by collapsing/reducing a dimension."""
--- a/tests/test_metadata.py
+++ b/tests/test_metadata.py
@@ -599,6 +599,55 @@ def test_metadata_bands_dimension(spec):
     assert metadata.band_common_names == ["F00", None]
 
 
+def test_cubemetadata_rename_labels():
+    metadata = CubeMetadata(
+        dimensions=[
+            SpatialDimension(name="x", extent=None),
+            BandDimension(name="b", bands=[Band("red"), Band("green")]),
+        ]
+    )
+    renamed = metadata.rename_labels(dimension="b", target=["raspberry", "lime"])
+    assert isinstance(renamed, CubeMetadata)
+    assert renamed.dimension_names() == ["x", "b"]
+    assert renamed.band_names == ["raspberry", "lime"]
+    assert renamed.bands == [Band("raspberry"), Band("lime")]
+
+
+def test_cubemetadata_rename_labels_invalid_dimension():
+    metadata = CubeMetadata(
+        dimensions=[
+            SpatialDimension(name="x", extent=None),
+            BandDimension(name="b", bands=[Band("red"), Band("green")]),
+        ]
+    )
+    with pytest.raises(ValueError, match=re.escape("Invalid dimension 'tempus'. Should be one of ['x', 'b']")):
+        _ = metadata.rename_labels(dimension="tempus", target=["1999"])
+
+
+def test_cubemetadata_rename_dimension():
+    metadata = CubeMetadata(
+        dimensions=[
+            SpatialDimension(name="x", extent=None),
+            BandDimension(name="b", bands=[Band("red"), Band("green")]),
+        ]
+    )
+    renamed = metadata.rename_dimension("b", "bands")
+    assert renamed.dimension_names() == ["x", "bands"]
+    assert renamed.band_dimension.name == "bands"
+    assert renamed.band_names == ["red", "green"]
+
+
+def test_cubemetadata_rename_dimension_invalid_dimension():
+    metadata = CubeMetadata(
+        dimensions=[
+            SpatialDimension(name="x", extent=None),
+            BandDimension(name="b", bands=[Band("red"), Band("green")]),
+        ]
+    )
+    with pytest.raises(ValueError, match=re.escape("Invalid dimension 'bandzz'. Should be one of ['x', 'b']")):
+        _ = metadata.rename_dimension("bandzz", "bands")
+
+
 def test_collectionmetadata_reduce_dimension():
     metadata = CollectionMetadata(
         {"cube:dimensions": {"x": {"type": "spatial"}, "b": {"type": "bands", "values": ["red", "green"]}}}


VictorVerhaert:
there is one test `test_run_local_udf_from_file_netcdf_with_context` that should not have changed in behaviour that fails.

openeo/rest/datacube.py
```diff
@@ -441,10 +441,23 @@ def load_stac(
...
+                    metadata = metadata.add_dimension(
+                        name="bands",
+                        type="bands",
+                        label=None,
+                    )
+                if all(metadata.band_dimension.contains_band(b) for b in bands):
+                    metadata = metadata.filter_bands(band_names=bands)
+                else:
+                    logging.warning(
+                        "Some bands are not available in the collection metadata. Using requested bands as is."
```
soxofaan, line 457: The warning should list which provided bands bands are not consistent with bands listed in metadata, to make this warning more user friendly/actionable

openeo/metadata.py
```diff
@@ -218,6 +218,16 @@ def rename_labels(self, target, source) -> Dimension:
     def rename(self, name) -> Dimension:
         return BandDimension(name=name, bands=self.bands)
 
+    def contains_band(self, band: Union[int, str]) -> bool:
```
soxofaan, line 221: I think this one deserves some direct unit tests

openeo/rest/datacube.py
```diff
@@ -441,10 +441,24 @@ def load_stac(
...
+                        name="bands",
+                        type="bands",
+                        label=None,
+                    )
+                missing_bands = [b for b in bands if not metadata.band_dimension.contains_band(b)]
+                if len(missing_bands) == 0:
+                    metadata = metadata.filter_bands(band_names=bands)
+                else:
+                    logging.warning(
+                        f"Bands {missing_bands} are not available in the collection metadata. Using requested bands as is."
```
soxofaan, line 458: "collection metadata" might be misleading.
STAC metadata might be better here

openeo/rest/datacube.py
```diff
@@ -441,10 +441,24 @@ def load_stac(
...
-                # TODO: also apply spatial/temporal filters to metadata?
-                metadata = metadata.filter_bands(band_names=bands)
-        except Exception:
+                if not metadata.has_band_dimension():
+                    metadata = metadata.add_dimension(
+                        name="bands",
+                        type="bands",
+                        label=None,
+                    )
+                missing_bands = [b for b in bands if not metadata.band_dimension.contains_band(b)]
```
soxofaan, line 453: "missing" is a bit confusing here. Maybe "unknown" fits the situation a bit better

CHANGELOG.md
```diff
@@ -15,6 +15,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- `load_stac` will now take over the requested band names if there is a mismatch/problem with the extracted collection metadata ([#752](https://github.com/Open-EO/openeo-python-client/issues/752))
```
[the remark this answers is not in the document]
soxofaan: Another attempt:
```suggestion
- When the bands provided to `Connection.load_stac(..., bands=[...])` do not fully match the bands the client extracted from the STAC metadata, a warning will be triggered, but the provided band names will still be used during the client-side preparation of the process graph. This is a pragmatic approach to bridge the gap between differing interpretations of band detection in STAC. Note that this might produce process graphs that are technically invalid and might not work on other backends or future versions of the backend you currently use. It is recommended to consult with the provider of the STAC metadata and openEO backend on the correct and future-proof band names. ([#752](https://github.com/Open-EO/openeo-python-client/issues/752))
```

openeo/rest/datacube.py
soxofaan: yes indeed, I mean subset instead of intersection

tests/rest/test_connection.py
```diff
@@ -2957,7 +2957,40 @@ def test_load_stac_no_cube_extension_temporal_dimension(self, con120, tmp_path,
...
+
+        caplog.set_level(logging.WARNING)
+        # Test with non-existing bands in the collection metadata
+        cube = con120.load_stac(str(stac_path), bands=bands)
+        assert cube.metadata.band_names == bands
+        if expected_warning is None:
+            assert caplog.text == ""
+        else:
+            assert expected_warning in caplog.text
+        caplog.clear()
```
soxofaan, line 2991: you don't have to do `caplog.clear()` at the end of a test, that's automatic

tests/rest/test_connection.py
```diff
@@ -2968,8 +3001,24 @@ def test_load_stac_band_filtering(self, con120, tmp_path):
...
+        stac_path.write_text(json.dumps(stac_data))
+
+        cube = con120.load_stac(str(stac_path))
+        assert cube.metadata.band_names == []
+
+        caplog.set_level(logging.WARNING)
+        cube = con120.load_stac(str(stac_path), bands=["B01", "B02"])
+        assert cube.metadata.band_names == ["B01", "B02"]
+        assert (
+            "The specified bands ['B01', 'B02'] are not a subset of the bands [] found in the STAC metadata (unknown bands: ['B01', 'B02']). Using specified bands as is."
```
soxofaan, line 3017: > `... not a subset of the bands [] ...`

hmm I'm a bit confused here: if there is no band dimension, doesn't the `add_dimension()` above at least define a band with name `None` ? How can the list of bands be empty?

VictorVerhaert: https://github.com/Open-EO/openeo-python-client/blob/4d350c4383757deed006d8d2d49cc4a49c438cbe/openeo/metadata.py#L619-L624 receives an empty dict and returns an empty lists which causes bands to be empty in https://github.com/Open-EO/openeo-python-client/blob/4d350c4383757deed006d8d2d49cc4a49c438cbe/openeo/metadata.py#L676

So I guess there isn't a case where the metadata has no bands dimension. I'll remove https://github.com/Open-EO/openeo-python-client/blob/4d350c4383757deed006d8d2d49cc4a49c438cbe/openeo/rest/datacube.py#L447-L452

soxofaan: > So I guess there isn't a case where the metadata has no bands dimension 

not yet, but that will be possible in the future with #743

openeo/metadata.py
```diff
@@ -423,6 +423,25 @@ def add_dimension(self, name: str, label: Union[str, float], type: Optional[str]
             dim = Dimension(type=type or "other", name=name)
         return self._clone_and_update(dimensions=self._dimensions + [dim])
 
+    def ensure_band_dimension(
+        self, *, name: Optional[str] = None, bands: List[Union[Band, str]], warning: str
```
VictorVerhaert, line 427: I'm not really a fan of having a function in metadata that takes a warning as an argument just to log it. Logging could happen outside of this function imo

tests/rest/test_connection.py
```diff
@@ -2957,66 +2977,174 @@ def test_load_stac_no_cube_extension_temporal_dimension(self, con120, tmp_path,
...
             ),
             (["B03", "B02"], None),
             (["B01", "B02", "B03"], None),
         ],
     )
-    def test_load_stac_band_filtering(self, con120, tmp_path, caplog, bands, expected_warning):
-        stac_path = tmp_path / "stac.json"
-        stac_data = StacDummyBuilder.collection(
-            summaries={"eo:bands": [{"name": "B01"}, {"name": "B02"}, {"name": "B03"}]}
+    def test_load_stac_band_filtering(self, dummy_backend, build_stac_ref, caplog, bands, expected_warning):
```
VictorVerhaert, line 3011: for my own reference, what is the difference between con120 and dummy backend, and when to use which?

tests/rest/test_connection.py
```diff
@@ -2957,66 +2977,174 @@ def test_load_stac_no_cube_extension_temporal_dimension(self, con120, tmp_path,
...
+        cube.execute()
+        assert dummy_backend.get_pg("load_stac")["arguments"] == {
+            **expected_pg_args,
+            "url": stac_ref,
+        }
+
+        if expected_warning:
+            assert expected_warning in caplog.text
+        else:
+            assert not caplog.text
```
VictorVerhaert, line 3146: what is the purpose of this unit test?

openeo/metadata.py
soxofaan: yeah I was hesitant about that, 
but forcing the caller to think of a warning  message is to make clear that this function should only be used exceptionally

tests/rest/test_connection.py
soxofaan: to test how a band dimension with a non-traditional name like "bandzz" is handled.
In your original implementation, you had the assumption that the band dimension is always named "bands", which is probably correctin in 99.9% of the current cases, but I wanted to eliminate that hardcoded assumption

tests/rest/test_connection.py
VictorVerhaert: thanks!

tests/rest/test_connection.py
```diff
@@ -2957,66 +2977,174 @@ def test_load_stac_no_cube_extension_temporal_dimension(self, con120, tmp_path,
...
+        # This is a temporary mock.patch hack to make metadata_from_stac return metadata without a band dimension
+        # TODO #743: Do this properly through appropriate STAC metadata
+        from openeo.metadata import metadata_from_stac as orig_metadata_from_stac
+
+        def metadata_from_stac(url: str):
+            metadata = orig_metadata_from_stac(url=url)
+            assert metadata.has_band_dimension()
+            metadata = metadata.drop_dimension("bands")
+            assert not metadata.has_band_dimension()
+            return metadata
```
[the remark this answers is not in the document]
VictorVerhaert: this comment was backed by another comment that got left out somehow, but I overlooked

> not yet, but that will be possible in the future with https://github.com/Open-EO/openeo-python-client/issues/743

so nvm this comment

VictorVerhaert:
expanded the `ensure_bands_dimension` to preserve existing bands. Back to you for the reviews

soxofaan:
> expanded the ensure_bands_dimension to preserve existing bands

I'm curious why that is necessary? When you get into the `ensure_bands_dimension` code path, it implies you don't trust the original band metadata (or there is none), so why  doing all the effort the still copying stuff?

Is this to support a particular use case?

openeo/metadata.py
jdries: Note that for cases like this where there is doubt about the API that is added, simply making it somehow 'private' or 'internal' API can help to be future proof.

VictorVerhaert:
reverted my last commit as indeed it complicates things for a cause we will not encounter 
Made `nesure_bands_dimension` "private"

soxofaan:
merged in 5e77f00a70eb2f8475c7fdab89b292be2bde1da8

(FYI: I did a rebase first to eliminate the commit-revert pair and to get a cleaner history)

PR #756, opened 2025-04-02
First version of tile based job splitter

### LINKED ISSUE #745: 20km UTM tile splitter
Example grid: https://artifactory.vgt.vito.be/ui/repos/tree/General/auxdata-public/grids/utm-tiling-grid-20km.gpkg

Define a function like this:

`def split_area(aoi: Union[bounds, shapely.Geometry], projection="AUTO:42001", tile_size:float=20.0)`

That simply returns a list of geometries that define grid cells, coordinates should be in projection.

Then we want to make it trivial to construct a job_db based on this, so wrap this code in a function:

```
df = gpd.GeoDataFrame({
    "geometry": split_area(aoi),
})
job_db = create_job_db("jobs.csv").initialize_from_df(df)
```

VincentVerelst:
A first version of a `split_area` function based on a `SizeBasedTileGrid` class. 

- Accepts `dict` and `shapely.geometry.Polygon` objects
- Resulting split tiles never exceed the bounds of the original AOI
- Tile size can be specified in km or degree
- For a more detailed explanation of how it behaves: see unit tests

closes #745 

@soxofaan, @jdries, besides the general review, I could use your help to advice on how to expose this to users and leave flexibility of adding more `TileGridInterface` classes in the future.


openeo/extra/job_management/job_splitting.py
```diff
@@ -0,0 +1,136 @@
...
+from shapely.geometry import shape
+from shapely.ops import transform
+
+
+class JobSplittingFailure(Exception):
+    pass
+
+
+# TODO: This function is also defined in openeo-python-driver. But maybe we want to avoid a dependency on openeo-python-driver?
+def reproject_bounding_box(bbox: Dict, from_crs: Optional[str], to_crs: str) -> Dict:
```
jdries, line 16: Can we somehow mark this and actually all classes as private API?
By exposing minimal new API, we have to worry less when making changes, and can minimize the new API to review carefully.

openeo/extra/job_management/job_splitting.py
```diff
@@ -0,0 +1,136 @@
...
+                    tiles.append(reproject_bounding_box(tile.as_dict(), from_crs=self.epsg, to_crs=bbox_crs))
+                else:
+                    tiles.append(tile.as_polygon())
+
+        return tiles
+
+
+def split_area(
+    aoi: Union[Dict, shapely.geometry.Polygon], projection="EPSG:3857", tile_size: float = 20.0
+) -> List[Union[Dict, shapely.geometry.Polygon]]:
```
jdries, line 127: 1. Should we return only one type? Input can indeed be polygon, but output is always rectangular tiles? I would however return them as polygon as well.
2. Many tiling systems also can assign a 'tile id', would be good if that is also returned

openeo/extra/job_management/job_splitting.py
```diff
@@ -0,0 +1,136 @@
...
+    """
+    Split area of interest into tiles of given size and projection.
+    :param aoi: area of interest (bounding box or shapely polygon)
+    :param projection: projection to use for splitting. Default is web mercator (EPSG:3857)
+    :param tile_size: size of tiles in km for UTM projections or degrees for WGS84
```
jdries, line 132:
```suggestion
    :param tile_size: size of tiles in unit of measure of the projection
```

jdries (changes requested):
logged some initial remarks on the API

openeo/extra/job_management/job_splitting.py
VincentVerelst: Ok, so `split_area` should then return a `geopandas.GeoDataFrame` with columns `tile_id` and `geometry`?

Do you have any suggestions for the `tile_id` convention for a `SizeBasedTileGrid`?

jdries: Here's an example precomputed grid, which also has tile id's:
https://artifactory.vgt.vito.be/artifactory/auxdata-public/grids/LAEA-20km-EU27.geojson
I would propose that this splitter should be able to replicate the same grid with the same tile id's.

openeo/extra/job_management/job_splitting.py
```diff
@@ -0,0 +1,136 @@
+import abc
+import math
+from typing import Dict, List, NamedTuple, Optional, Union
+
+import pyproj
```
soxofaan, line 5: FYI: pyproj is not a core dependency, only an optional one under the `localprocessing` extra at the moment.

Not sure if this dependency can be avoided, or is it a hard dependency for this whole module?

openeo/extra/job_management/job_splitting.py
```diff
@@ -0,0 +1,136 @@
...
+
+# TODO: This class is also defined in openeo-aggregator. But maybe we want to avoid a dependency on openeo-aggregator?
+class BoundingBox(NamedTuple):
+    """Simple NamedTuple container for a bounding box"""
+
+    west: float
+    south: float
+    east: float
+    north: float
+    crs: str = "EPSG:4326"
```
soxofaan, line 41: Note: in strict openEO, EPSG CRS descriptors are integer codes, instead of the string format "EPSG:XXX". String CRS descriptors are assumed to be WKT2.

e.g. see https://github.com/Open-EO/openeo-processes/blob/6141771129d58d2a292db3b91d80c812c49f4e52/load_collection.json#L67
> Coordinate reference system of the extent, specified as as [EPSG code](http://www.epsg-registry.org/) or [WKT2 CRS string](http://docs.opengeospatial.org/is/18-010r7/18-010r7.html).


```json
 "anyOf": [
                                {
                                    "title": "EPSG Code",
                                    "type": "integer",
                                    "subtype": "epsg-code",
                                    "minimum": 1000,
                                    "examples": [
                                        3857
                                    ]
                                },
                                {
                                    "title": "WKT2",
                                    "type": "string",
                                    "subtype": "wkt2-definition"
                                }
                            ],
```


If we introduce something new in the client I think we should follow that convention from the start.
I'm not sure actually why I did that wrong in the aggregator

openeo/extra/job_management/job_splitting.py
```diff
@@ -0,0 +1,136 @@
...
+    """
+    box = shapely.geometry.box(bbox["west"], bbox["south"], bbox["east"], bbox["north"])
+    if from_crs is None:
+        from_crs = bbox["crs"]
+    tranformer = pyproj.Transformer.from_crs(crs_from=from_crs, crs_to=to_crs, always_xy=True)
+    reprojected = transform(tranformer.transform, box)
+    return dict(zip(["west", "south", "east", "north"], reprojected.bounds), crs=to_crs)
+
+
+# TODO: This class is also defined in openeo-aggregator. But maybe we want to avoid a dependency on openeo-aggregator?
```
soxofaan, line 33: yes, we can not depend on aggregator from the client (the other way around is already happening, so that would actually be a circular dependency). 
just remove this TODO

openeo/extra/job_management/job_splitting.py
```diff
@@ -0,0 +1,136 @@
...
+import shapely
+from shapely.geometry import shape
+from shapely.ops import transform
+
+
+class JobSplittingFailure(Exception):
+    pass
+
+
+# TODO: This function is also defined in openeo-python-driver. But maybe we want to avoid a dependency on openeo-python-driver?
```
soxofaan, line 15: yes, we can not depend on openeo-python-driver from the client (the other way around is already happening, so that would actually be a circular dependency). 
just remove this TODO

openeo/extra/job_management/job_splitting.py
soxofaan: > return a geopandas.GeoDataFrame with ...

note that geopandas is also not a core dependency of the client. It's only a test dependency

openeo/extra/job_management/job_splitting.py
```diff
@@ -0,0 +1,136 @@
...
+
+def split_area(
+    aoi: Union[Dict, shapely.geometry.Polygon], projection="EPSG:3857", tile_size: float = 20.0
+) -> List[Union[Dict, shapely.geometry.Polygon]]:
+    """
+    Split area of interest into tiles of given size and projection.
+    :param aoi: area of interest (bounding box or shapely polygon)
+    :param projection: projection to use for splitting. Default is web mercator (EPSG:3857)
+    :param tile_size: size of tiles in km for UTM projections or degrees for WGS84
+    :return: list of tiles (dicts with keys "west", "south", "east", "north", "crs" or shapely polygons). For dicts the original crs is preserved. For polygons the projection is set to the given projection.
```
soxofaan, line 133: > For polygons the projection is set to the given projection.

I'm confused here: shapely polygons don't have a CRS attribute, so you can't *set* the projection, you can at most assume something

openeo/extra/job_management/job_splitting.py
```diff
@@ -0,0 +1,136 @@
...
+    east: float
+    north: float
+    crs: str = "EPSG:4326"
+
+    @classmethod
+    def from_dict(cls, d: Dict) -> "BoundingBox":
+        return cls(**{k: d[k] for k in cls._fields if k not in cls._field_defaults or k in d})
+
+    @classmethod
+    def from_polygon(cls, polygon: shapely.geometry.Polygon, projection: Optional[str] = None) -> "BoundingBox":
```
soxofaan, line 48: why only `Polygon` support and no `MultiPolygon`? I think we can support that here without any effort

openeo/extra/job_management/job_splitting.py
```diff
@@ -0,0 +1,136 @@
...
+        """Calculate tiles to cover given bounding box"""
+        ...
+
+
+class SizeBasedTileGrid(TileGridInterface):
+    """
+    Specification of a tile grid, parsed from a size and a projection.
+    """
+
+    def __init__(self, epsg: str, size: float):
```
soxofaan, line 74: as mentioned above: epsg codes should be annotated as integers (that's just what epsg codes are). 
at best I would normalize string  `"EPSG:123"` to int `123` .

e.g. check out `openeo.util.normalize_crs`

openeo/extra/job_management/job_splitting.py
```diff
@@ -0,0 +1,136 @@
...
+        self.epsg = epsg
+        self.size = size
+
+    @classmethod
+    def from_size_projection(cls, size: float, projection: str) -> "SizeBasedTileGrid":
+        """Create a tile grid from size and projection"""
+        return cls(projection.lower(), size)
+
+    def get_tiles(self, geometry: Union[Dict, shapely.geometry.Polygon]) -> List[Union[Dict, shapely.geometry.Polygon]]:
```
soxofaan, line 83: as mentioned above: I guess it's trivial to extend `geometry` support to MultiPolygons too

openeo/extra/job_management/job_splitting.py
```diff
@@ -0,0 +1,136 @@
...
+                    south=max(y * tile_size, to_cover.south),
+                    east=min((x + 1) * tile_size + x_offset, to_cover.east),
+                    north=min((y + 1) * tile_size, to_cover.north),
+                    crs=self.epsg,
+                )
+
+                if isinstance(geometry, dict):
+                    tiles.append(reproject_bounding_box(tile.as_dict(), from_crs=self.epsg, to_crs=bbox_crs))
+                else:
+                    tiles.append(tile.as_polygon())
```
soxofaan, line 120: This looks like risky behavior if I understand correctly: in one case you reproject and in another you don't. I'm not sure users will anticipate that when naively using this function.
I understand you can document that, but these kind of important details are easily glossed over.

tests/extra/job_management/test_job_splitting.py
```diff
@@ -0,0 +1,242 @@
...
+
+@pytest.fixture
+def mock_dict_with_crs_utm():
+    return {
+        "west": 0.0,
+        "south": 0.0,
+        "east": 100_000.0,
+        "north": 100_000.0,
+        "crs": "EPSG:3857",
+    }
```
soxofaan, line 41: I guess it's a matter of taste, but these four fixtures `mock_polygon_wgs`, ... `mock_dict_with_crs_utm` feel like overkill to me.
I guess it is to be DRY, but in testing I think this extra level of indirection for these simple constructs is worse for readability than violating DRY.
Also, if you still to be DRY, I think you should just use constants instead of pytest fixtures (you are not really using them as fixtures, in terms of setup/teardown)

soxofaan (commented):
some quick notes

openeo/extra/job_management/job_splitting.py
VincentVerelst: For me personally, this is confusing: the unit of measure of UTM projection is meter, but we expect `tile_size` to be in kilometer.

openeo/extra/job_management/job_splitting.py
```diff
@@ -0,0 +1,136 @@
...
+            bbox_crs = self.epsg
+        else:
+            raise JobSplittingFailure("geometry must be a dict or a shapely.geometry.Polygon")
+
+        if self.epsg == "epsg:4326":
+            tile_size = self.size
+            x_offset = 0
+        else:
+            tile_size = self.size * 1000
+            x_offset = 500_000
```
soxofaan, line 98: this if-else looks too simple
- epsg could be in upper case "EPSG:4326"
- not being "epsg:4326" is not enough to be sure it's UTM

also: the multiplication with 1000 looks quite ad-hoc and based on undocumented assumptions. In UTM the base unit is meters, but here you seem to assume I guess that the tile size is given in km, but that is currently not clear

openeo/extra/job_management/job_splitting.py
VincentVerelst: Proposed wording change: `For dicts the original crs is preserved. For polygons the projection is assumed to be the given tilegrid projection.`

openeo/extra/job_management/job_splitting.py
VincentVerelst: We indeed need to be able to reproject polygons in this module. Not sure if that's possible without pyproj? Possibly with geopandas, but that's also not a core dependency, and it built on pyproj I think

openeo/extra/job_management/job_splitting.py
soxofaan: at first sight I think this method tries to do too much (bboxes with crs, polygons without crs ..., reprojections, ...) 

I'd extract the core functionality (double for-loop) as a helper and make its logic "CRS-free". And reuse that in more specialized versions (for CRS-aware bboxes, for naive polygons, ...)

openeo/extra/job_management/job_splitting.py
soxofaan: you can still do basic splitting without need for reprojection if you stay within the same CRS, right?
So I think it should be possible to keep `pyproj` an optional dependency

openeo/extra/job_management/job_splitting.py
VincentVerelst: - Do you have any recommended utils on how to detect if a certain CRS is in UTM?

- The expected units are indeed km for UTM (taken from the aggregator, can also just change it). It's documented in the `split_area` annotation, but I'll also add it to this class.

soxofaan: epsg codes in ranges 32601-32660 and 32701-32760 are UTM crses.
e.g. also see https://github.com/Open-EO/openeo-python-driver/blob/af50be3f59f2152ff36c15512735343fe7e463c5/openeo_driver/util/utm.py#L29-L33

The base unit of UTM is meter, so using something else (even if documented) is asking for trouble somewhere down the line. If you want to support working with kilometer to make it user friendly, you should make that explicit in some way (e.g. in the aggregator, the tiling specification is expected to contain an explicit unit like "km" : https://github.com/Open-EO/openeo-aggregator/blob/a0d40582c813a4d1ed62221389fd5a776106c809/src/openeo_aggregator/partitionedjobs/splitting.py#L80-L103 )

VincentVerelst: Ok, then I propose we just assume tile_size to be in meter for UTM crs

openeo/extra/job_management/job_splitting.py
soxofaan: yes I would make sure to have a core function that does just tile splitting in a naive way (in the sense that it does not considers any CRS aspects). This function would then not have to depend on pyproj. 
Then, reuse this function for particular uses cases that possibly require reprojects, which is fine as long as it does not leak into the core splitting function.
Not sure if it's clear what I mean :)

openeo/extra/job_management/job_splitting.py
soxofaan: I made the whole module private for now with 20899c7127ce804a83f8dcfc2397b58662537ca1

soxofaan:
as discussed offline: I made job_splitting a private submodule, moved a couple of review notes as TODO comments in the code and merged in ee18290f

PR #759, opened 2025-04-07
747 robustranged download support

dsamaey:
Manually tested with https://openeo.dataspace.copernicus.eu/openeo/1.2/jobs/j-2504020555234b3287f2fe696bea5ded/results/assets/MzJjYzdkZGItZjdlMS00YjFjLTk3OTYtZjlmZTM5Y2I4ZmVi/efc6009b1fdb1d283b91e65538400aac/feature-datacubes_v1_feature-cube_year2024_E412N322.tif?expires=1744207076

Chunks are directly written to the file since all chunks are downloaded (and retried) sequentially.


openeo/rest/_connection.py
```diff
@@ -194,6 +194,24 @@ def get(
...
+    ) -> Response:
+        """
+        Do HEAD request to REST API.
+
+        :param path: API path (without root url)
+        :param params: Additional query parameters
+        :param auth: optional custom authentication to use instead of the default one
+        :return: response: Response
+        """
+        return self.request("get", path=path, params=params, auth=auth, **kwargs)
```
EmileSonneveld, line 213: Is "get" intentional here?

dsamaey: Nope

openeo/rest/job.py
```diff
@@ -424,6 +418,43 @@ def load_bytes(self) -> bytes:
     # TODO: more `load` methods e.g.: load GTiff asset directly as numpy array
 
 
+    def _download_to_file(self, url: str, target: Path, chunk_size: int):
+        head = self.job.connection.head(url, stream=True)
+        if head.ok and 'Accept-Ranges' in head.headers and 'bytes' in head.headers['Accept-Ranges']:
```
soxofaan, line 423: you can simplify this to
```suggestion
        if head.ok and head.headers.get("Accept-Ranges") == "bytes":
```

openeo/rest/job.py
```diff
@@ -424,6 +418,43 @@ def load_bytes(self) -> bytes:
...
+            file_size = int(head.headers['Content-Length'])
```
soxofaan, line 424: I guess you should also check that there is a "Content-Length" header, otherwise this will fail

openeo/rest/job.py
```diff
@@ -424,6 +418,43 @@ def load_bytes(self) -> bytes:
...
+            self._download_chunked(url=url, target=target, file_size=file_size, chunk_size=chunk_size)
+        else:
+            self._download_unchunked(url=url, target=target)
+
+
+    def _download_chunked(self, url: str, target: Path, file_size: int, chunk_size: int):
```
soxofaan, line 430: I think some things are mixed up here. Originally we already had chunked downloading (`for block in response.iter_content`), but what is added in this feature is "range request" based downloading, so the naming of these private methods is a bit confusing.

openeo/rest/job.py
```diff
@@ -424,6 +418,43 @@ def load_bytes(self) -> bytes:
...
+                    except OpenEoApiPlainError as error:
+                        tries_left -= 1
+                        if tries_left > 0 and error.http_status_code in RETRIABLE_STATUSCODES:
+                            logger.warning(f"Failed to retrieve chunk {from_byte_index}-{to_byte_index} from {url} (status {error.http_status_code}) - retrying")
+                            continue
+                        else:
+                            raise error
+
+
+    def _download_unchunked(self, url: str, target: Path):
```
soxofaan, line 451: As mentioned, if ranged downloading is not possible, the fallback should still be original chunked downloading with `for block in response.iter_content`

openeo/rest/job.py
```diff
@@ -501,7 +532,7 @@ def get_asset(self, name: str = None) -> ResultAsset:
                     "No asset {n!r} in: {a}".format(n=name, a=[a.name for a in assets])
                 )
 
-    def download_file(self, target: Union[Path, str] = None, name: str = None) -> Path:
+    def download_file(self, target: Union[Path, str] = None, name: str = None, chunk_size=DEFAULT_DOWNLOAD_CHUNK_SIZE) -> Path:
```
soxofaan, line 535: make this additional argument a keyword-only argument:
```suggestion
    def download_file(self, target: Union[Path, str] = None, name: str = None, *, chunk_size=DEFAULT_DOWNLOAD_CHUNK_SIZE) -> Path:
```

openeo/rest/job.py
```diff
@@ -513,12 +544,12 @@ def download_file(self, target: Union[Path, str] = None, name: str = None) -> Pa
...
         """
         try:
-            return self.get_asset(name=name).download(target=target)
+            return self.get_asset(name=name).download(target=target, chunk_size=chunk_size)
         except MultipleAssetException:
             raise OpenEoClientException(
                 "Can not use `download_file` with multiple assets. Use `download_files` instead.")
 
-    def download_files(self, target: Union[Path, str] = None, include_stac_metadata: bool = True) -> List[Path]:
+    def download_files(self, target: Union[Path, str] = None, include_stac_metadata: bool = True, chunk_size=DEFAULT_DOWNLOAD_CHUNK_SIZE) -> List[Path]:
```
soxofaan, line 552: same (and include `include_stac_metadata` as well):
```suggestion
    def download_files(self, target: Union[Path, str] = None, *, include_stac_metadata: bool = True, chunk_size=DEFAULT_DOWNLOAD_CHUNK_SIZE) -> List[Path]:
```

tests/rest/test_job.py
```diff
@@ -536,10 +541,28 @@ def job_with_1_asset(con100, requests_mock, tmp_path) -> BatchJob:
...
+    }})
+    requests_mock.head(API_URL + "/dl/jjr1.tiff", headers={"Content-Length": f"{len(TIFF_CONTENT)}", "Accept-Ranges": "bytes"})
+
+    chunk_size = 1000
+    for r in range(0, len(TIFF_CONTENT), chunk_size):
+        from_bytes = r
+        to_bytes = min(r + chunk_size, len(TIFF_CONTENT)) - 1
+        requests_mock.get(API_URL + "/dl/jjr1.tiff", request_headers={"Range": f"bytes={from_bytes}-{to_bytes}"},
+                     response_list = [{"status_code": 500, "text": "Server error"},
+                                      {"status_code": 206, "content": TIFF_CONTENT[from_bytes:to_bytes+1]}])
```
soxofaan, line 563: This for-loop of mocked request is mimics the actual implementation very closely. 
I think it's more robust to avoid this tight coupling by creating a single request mock, but with a dynamic handler where you inspect the received request headers as documented in https://requests-mock.readthedocs.io/en/latest/response.html#dynamic-response

In this test file we have already an example of that:
https://github.com/Open-EO/openeo-python-client/blob/4aba08a6a758385a01e7d4b00ba5f9a00daa2655/tests/rest/test_job.py#L794-L800

jdries:
@soxofaan  ok to merge? Broken downloads is a big issue for many users, and this might really help quite a bit.

CHANGELOG.md
```diff
@@ -29,6 +29,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `sar_backscatter`: try to retrieve coefficient options from backend ([#693](https://github.com/Open-EO/openeo-python-client/issues/693))
 - Improve error message when OIDC provider is unavailable ([#751](https://github.com/Open-EO/openeo-python-client/issues/751))
 - Added `on_response_headers` argument to `DataCube.download()` and related to handle (e.g. `print`) the response headers ([#560](https://github.com/Open-EO/openeo-python-client/issues/560))
+- Added more robust download for large job result files (if supported by the server)
```
soxofaan, line 32: can you link to the github ticket here as well?

openeo/rest/job.py
```diff
@@ -424,6 +419,43 @@ def load_bytes(self) -> bytes:
...
+                            continue
+                        else:
+                            raise error
+
+
+    def _download_all_at_once(self, url: str, target: Path, *, chunk_size: int=DEFAULT_DOWNLOAD_CHUNK_SIZE):
+        with self.job.connection.get(path=url, stream=True) as r:
+            r.raise_for_status()
+            with target.open("wb") as f:
+                shutil.copyfileobj(fsrc=r.raw, fdst=f, length=chunk_size)
```
soxofaan, line 456: This should fall back to the original chunked implementation
```
        response = self._get_response(stream=True)
        with target.open("wb") as f:
            for block in response.iter_content(chunk_size=chunk_size):
                f.write(block)
```

because this change loads in all data in memory, which is a regression

tests/rest/test_job.py
```diff
@@ -8,19 +9,23 @@
 
 import pytest
 import requests
+from charset_normalizer import from_bytes
```
soxofaan, line 12: where does this come from? That looks like a new dependency

as far as I understand you are not using `from_bytes`, except overwriting it lower

soxofaan (commented):
I think there is still a regression here:
non ranged downloads will be done as a whole in memory instead of being chunked

openeo/rest/job.py
dsamaey: shutil.copyfileobj(fsrc=r.raw, fdst=f, length=chunk_size) performs a chunked copy (https://docs.python.org/3/library/shutil.html#:~:text=shutil.copyfileobj,will%20be%20copied.)

jdries:
@dsamaey  indeed, for the 'download all at once' code path, we can just make sure to use what was already there, and then this should be fine:
```
response = self._get_response(stream=True)
with target.open("wb") as f:
    for block in response.iter_content(chunk_size=chunk_size):
        f.write(block)
```
Can you merge that in?

Issue #761, opened 2025-04-15
APEx reference check needs better representation (ascii art diff, diff image, statistics)


dsamaey:
**Statistics are most important**
#percentage of pixels that are different (tolerance)
median, mean, min, max
Perhaps write diff to a png -- Issue with dimensionality (per timestamp, per band) max 10 files.

**Location of differences**
Which timestamp(s) failed
Which band(s) failed
bbox of differences as geojson (so you can crop)
* it happens that changes are very concentrated

**Extras**
Numpy has something similar. We have unit tests that already print more informative messages.

Pipeline for cdse staging might not be configured yet for writing to s3 bucket. Ask @JeroenVerstraelen.

JeroenVerstraelen:
Definition of done:
Integrate into error logs of automatically generated github issues
e.g. https://github.com/ESA-APEx/apex_algorithms/issues/164

soxofaan:
FYI: I missed part of the discussion this morning, but to test  this feature on APEx benchmarks, it's not necessary to go through the cumbersome process of merging an WIP PR, doing a release, waiting for the benchmarks to run automatically, etc.

It's not that hard to just test this locally.

Roughly (mutatis mutandis):

```bash
# Set up local context for benchmarks
git clone https://github.com/ESA-APEx/apex_algorithms.git
cd apex_algorithms
# Create isolated venv for benchmarks project
python -m venv --prompt .  venv
. venv/bin/activate
python -m pip install qa/tools
python -m pip install -r qa/benchmarks/requirements.txt

# Override openeo package with local checkout, having necessary changes in feature branch
python -m pip install ../path/to/local/openeo-python-client

# Optionally: go back to commit with known benchmark errors here

# Run a benchmark
cd qa/benchmarks
pytest -vv -k 'test_run_benchmark[max_ndvi]'
```

Note that the first time you do this, you have to watch out for authentication instructions

soxofaan:
> Definition of done:
> Integrate into error logs of automatically generated github issues
> e.g. https://github.com/ESA-APEx/apex_algorithms/issues/164

FYI: There should be no effort to integrate that, it should work out of the box I think

soxofaan:
manually triggered a run (ran the "bap_composite" benchmark): 
https://github.com/ESA-APEx/apex_algorithms/actions/runs/14906019298/job/41868461282

which shows the ASCII art:

[...]

PR #765, opened 2025-04-22
Issue #761 better diff for apex reference check

dsamaey:


setup.py
```diff
@@ -22,7 +21,8 @@
     "mock",
     "requests-mock>=1.8.0",
     "httpretty>=1.1.4",
-    "urllib3<2.3.0",  # httpretty doesn't work properly with urllib3>=2.3.0. See #700 and https://github.com/gabrielfalcao/HTTPretty/issues/484
+    "urllib3<2.3.0",
+    # httpretty doesn't work properly with urllib3>=2.3.0. See #700 and https://github.com/gabrielfalcao/HTTPretty/issues/484
```
soxofaan, line 25: any reason to change this ?

note that this comment is about urllib dependency, now it seems to be about netcdf.

openeo/testing/results.py
```diff
@@ -163,6 +293,31 @@ def assert_xarray_dataarray_allclose(
...
+    """
+    Assert that two Xarray ``DataArray`` instances are equal (with tolerance).
+
+    :param actual: actual data, provided as Xarray DataArray object or path to NetCDF/GeoTIFF file.
+    :param expected: expected or reference data, provided as Xarray DataArray object or path to NetCDF/GeoTIFF file.
+    :param rtol: relative tolerance
+    :param atol: absolute tolerance
+    :raises AssertionError: if not equal within the given tolerance
+
+    .. versionadded:: 0.31.0
```
soxofaan, line 312: next version will probably be 0.41.0

openeo/testing/results.py
```diff
@@ -88,6 +91,129 @@ def _as_xarray_dataarray(data: Union[str, Path, xarray.DataArray]) -> xarray.Dat
     return data
 
 
+def ascii_art(diff_data: DataArray) -> str:
+    scale: int = max(1, int(diff_data.sizes["x"] / 100))
+    data_max = diff_data.max().item()
+    if data_max == 0:
+        data_max = 1
+    grayscale_characters = "$@B%8&WM#*oahkbdpqwmZO0QLCJUYXzcvunxrjft/\|()1{}[]?-_+~<>i!lI;:,\"^`'. "
```
soxofaan, line 99: just curious: where does this "gradient" come from?

also note that with the (github.com) font I'm currently looking looking at this, it doesn't seem to be monotonically increasing in "lightness":  e.g. `+` comes after `_`  but it looks "darker"

openeo/testing/results.py
```diff
@@ -88,6 +91,129 @@ def _as_xarray_dataarray(data: Union[str, Path, xarray.DataArray]) -> xarray.Dat
...
+    coarsened = diff_data.coarsen(dim={"x": scale, "y": scale}, boundary="pad").mean()
+    coarsened = coarsened.transpose("y", "x", ...)
+    top = "┌" + "─" * coarsened.sizes["x"] + "┐\n"
+    bottom = "\n└" + "─" * coarsened.sizes["x"] + "┘"
+
+    def pixelChar(v) -> str:
+        if np.isnan(v):
+            return " "
+        i = int(v * 69 / data_max)
```
soxofaan, line 108: this probably only works for positive values.

note that it's not uncommon to have EO data that is mixed negative-positive (e.g NDVI) or striclty negative (e.g. decibel scaled radar data)

openeo/testing/results.py
```diff
@@ -88,6 +91,129 @@ def _as_xarray_dataarray(data: Union[str, Path, xarray.DataArray]) -> xarray.Dat
...
+    Compare two xarray DataArrays with tolerance and report mismatch issues (as strings)
+
+    Checks that are done (with tolerance):
+    - (optional) Check fraction of mismatching pixels (difference exceeding some tolerance).
+      If fraction is below a given threshold, ignore these mismatches in subsequent comparisons.
+      If fraction is above the threshold, report this issue.
+    - Compare actual and expected data with `xarray.testing.assert_allclose` and specified tolerances.
+
+    :return: list of issues (empty if no issues)
+    """
```
soxofaan, line 135: can you document what the difference is with the existing `_compare_xarray_dataarray` ?

openeo/testing/results.py
```diff
@@ -250,7 +405,10 @@ def assert_xarray_allclose(
     if isinstance(actual, xarray.Dataset) and isinstance(expected, xarray.Dataset):
         assert_xarray_dataset_allclose(actual, expected, rtol=rtol, atol=atol)
     elif isinstance(actual, xarray.DataArray) and isinstance(expected, xarray.DataArray):
-        assert_xarray_dataarray_allclose(actual, expected, rtol=rtol, atol=atol)
+        if (["x", "y", "band"]).elements_in(expected.dims):
```
soxofaan, line 408: This doesn't look like valid python

you probably mean something like 
```suggestion
        if {"x", "y", "band"}.issubset(expected.dims):
```

I'm also not sure why "bands" is included in this check

tests/testing/test_results.py
```diff
@@ -45,15 +43,13 @@ def test_simple_defaults(self):
...
                     "Shape mismatch: (2, 3) != (3,)",
-                    dirty_equals.IsStr(regex="Left and right DataArray objects are not close.*", regex_flags=re.DOTALL),
                 ],
             ),
             (
                 xarray.DataArray([[1], [2], [3]]),
                 [
                     "Dimension mismatch: ('dim_0', 'dim_1') != ('dim_0',)",
                     "Shape mismatch: (3, 1) != (3,)",
-                    dirty_equals.IsStr(regex="Left and right DataArray objects are not close.*", regex_flags=re.DOTALL),
```
soxofaan, line 56: I don't completely understand at the moment why these issues are not triggered anymore with the new implementation

tests/testing/test_results.py
```diff
@@ -75,20 +71,12 @@ def test_simple_shape_mismatch(self, actual, expected_issues):
                     "Dimension mismatch: ('y', 'x') != ('x', 'y')",
                     "Coordinates mismatch for dimension 'x': [0 1 2] != [0 1]",
                     "Coordinates mismatch for dimension 'y': [0 1] != [0 1 2]",
-                    dirty_equals.IsStr(
-                        regex=r"Left and right DataArray objects are not close.*Differing dimensions:.*\(y: 2, x: 3\) != \(x: 2, y: 3\)",
-                        regex_flags=re.DOTALL,
-                    ),
```
soxofaan, line 81: This looks like a regression: there is interesting feedback being removed here

tests/testing/test_results.py
```diff
@@ -351,6 +335,108 @@ def test_allclose_minimal_success(self, tmp_path, actual_dir, expected_dir):
...
+        )
+        actual_ds.to_netcdf(actual_dir / "data.nc")
+        with raises_assertion_error_or_not(
+            r"Issues for file 'data.nc'.*"
+            r"Issues for variable 'b2'.*"
+            r"t 2: value difference exceeds tolerance \(rtol 1e-06, atol 1e-06\), min:33.0, max: 42.0, mean: 37.5, var: 11.2.*"
+            r"t 2: differing pixels: 4/20 \(20.0%\), spread over 8.3% of the area"
+        ):
+            assert_job_results_allclose(actual=actual_dir, expected=expected_dir, tmp_path=tmp_path)
+
```
soxofaan, line 439: Can you also add a simple test for `ascii_art` (e.g. with a simple 10 by 5 use case or something like that).
It's currently a public function, so people/projects might start depending on it

openeo/testing/results.py
```diff
@@ -88,6 +91,129 @@ def _as_xarray_dataarray(data: Union[str, Path, xarray.DataArray]) -> xarray.Dat
...
+                mask = diff_data.notnull()
+                if mask.dims[0] != "y":
+                    mask = mask.transpose()
+                c1 = coord_grid[0][mask]
+                c2 = coord_grid[1][mask]
+                coordinates = np.dstack((c1, c2)).reshape(-1, 2)
+
+                art = ascii_art(diff_data)
+                print(f"Difference ascii art for {key}")
+                print(art)
```
soxofaan, line 194: I also have my doubts about using a raw `print` here, because that might not play well with pytest based reporting (e.g. we also try to produce a HTML report in APEx).

I guess it's a matter of experimenting  on what currently happens here, or what alternatives are possible

openeo/testing/results.py
```diff
@@ -8,8 +8,11 @@
 from pathlib import Path
 from typing import List, Optional, Union
 
+import numpy as np
 import xarray
 import xarray.testing
+from scipy.spatial import ConvexHull
```
soxofaan, line 14: This makes scipy a runtime dependency (now it's just a test dependency). Scipy is quite a heavy dependency (in terms of compiled extensions and additional transitive dependencies), which we we generally want to avoid.

If we keep it as optional dependency, this should be documented, e.g. here:
https://github.com/Open-EO/openeo-python-client/blob/c6bc1c91ff7ba542f044c5ffea6d4ac3379f242e/docs/installation.rst?plain=1#L82-L95

Or, is it actually necessary to have scipy as hard dependency here? Is that convex hull feature vital for the reporting quality? I think there could be a simpler fallback fraction-of-area calculation if ConvexHull is not available

soxofaan (commented):
I started experimenting with this, and got in trouble with this new scipy dependency

openeo/testing/results.py
soxofaan: for example: you use the convex hull to estimate the impacted area of coordinates with differences, but a much simpler/coarser approximation is just taking the min/max values of these coordinates to get the bounding box. Which does not require scipy

openeo/testing/results.py
soxofaan: I also noticed this warning when running tests:
```
  .../openeo-python-client/openeo/testing/results.py:99: SyntaxWarning: invalid escape sequence '\|'
    grayscale_characters = "$@B%8&WM#*oahkbdpqwmZO0QLCJUYXzcvunxrjft/\|()1{}[]?-_+~<>i!lI;:,\"^`'. "
```

openeo/testing/results.py
```diff
@@ -88,6 +91,129 @@ def _as_xarray_dataarray(data: Union[str, Path, xarray.DataArray]) -> xarray.Dat
     return data
 
 
+def ascii_art(diff_data: DataArray) -> str:
+    scale: int = max(1, int(diff_data.sizes["x"] / 100))
```
soxofaan, line 95: as noted in https://github.com/Open-EO/openeo-python-client/issues/761#issuecomment-2858602486  I think a size of 100 is too much

I would limit it to e.g. 40x40 or something like that

and while at it: just make this max size/width  a function argument (with a reasonable default)

openeo/testing/results.py
soxofaan: Ok it does seem to work in the HTML report:
image

but as mentioned elsewhere, the ascii art is a bit too large for good overview in that context

openeo/testing/results.py
dsamaey: it's an absolute diff

PR #768, opened 2025-05-05
Add provenance output support to execute() response

hapix:
### Summary

This PR adds support for returning workflow provenance data as part of the `execute()` result in the OpenEO Python client. It complements the provenance generation added in `openeo-pg-parser-networkx` via the `yProv4WFS` library.

### Key Changes

- Added optional `provenance` output to `execute()` result structure

### Dependencies

This PR depends on the provenance functionality in [`openeo-pg-parser-networkx`](https://github.com/Open-EO/openeo-pg-parser-networkx).

### DIFF
--- a/openeo/local/connection.py
+++ b/openeo/local/connection.py
@@ -2,6 +2,8 @@ import datetime
 import logging
 from pathlib import Path
 from typing import Callable, Dict, List, Optional, Union
+import os
+
 
 import numpy as np
 import xarray as xr
@@ -270,6 +272,7 @@ class LocalConnection():
         *,
         validate: Optional[bool] = None,
         auto_decode: bool = True,
+        return_provenance: bool = False,
     ) -> xr.DataArray:
         """
         Execute locally the process graph and return the result as an xarray.DataArray.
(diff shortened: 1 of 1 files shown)


openeo/local/connection.py
```diff
@@ -282,4 +285,16 @@ def execute(
...
+
+        # Return the result and get the workflow provinance (yprov4wfs)
+        result = pg.to_callable(PROCESS_REGISTRY)()
+        workflow = pg.workflow
+
+        # To save the provenance file in the specific path use:
+        # workflow.prov_to_json(directory_path=save_path)
+
+        if return_provenance:
+            return result, workflow
```
soxofaan, line 298: This custom return should be documented in the docblock and return annotation

openeo/local/connection.py
```diff
@@ -270,6 +272,7 @@ def execute(
         *,
         validate: Optional[bool] = None,
         auto_decode: bool = True,
+        return_provenance: bool = False,
```
soxofaan, line 275: You are adding a custom argument here to a more public API of `Connection.execute()`, which is fine as long as you call this `LocalConnection.execute()` yourself. But in general this method will be called automatically without this argument (because it's not in the official API), e.g.:

```python
cube = local_connection.load_collection(...)
res = cube.execute()  
```
the latter `execute()` is a method defined on `DataCube` and does not support `return_provenance`, let alone it will pass it properly to `LocalConnection.execute`

openeo/local/connection.py
soxofaan: that being said, I'm not a big fan of the pattern of returning different data structures (tuple of two things instead of a single `DataArray`) depending on input arguments.
Especially, because in normal usage of the openEO python client, the `execute` method of connection objects (`LocallConnection` here) is usually not used directly by users, but indirectly through `DataCube.execute()` or something equivalent. So changing the input and output API of `Connection.execute` is going to create problems

Issue #764, opened 2025-04-22
Support Retry-After on `429 Too Many Requests`


soxofaan:
a `429 Too Many Requests` response can have a [`Retry-After`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Retry-After) header to indicate how long the user agent should wait before making a follow up request.

The openeo client could support this situation better, for example or especially in the job manager.

soxofaan:
Another place to support this is in the `job.start_and_wait()`  poll loop

soxofaan:
related to / duplicate of  
- https://github.com/Open-EO/openeo-python-client/issues/441
- with related PR https://github.com/Open-EO/openeo-python-client/pull/547

Issue #441, opened 2023-06-15
Automatic backoff/retry on `HTTP 429 Too Many Requests`


soxofaan:
Some backends have middleware to throttle the number of concurrent requests, returning a `429 Too Many Requests` if some limit is exceeded. The python client could provide automatic retry with (e.g. exponential) backoff functionality.

related to #440

jdries:
The python requests library even supports this quite easily
```
from requests.adapters import HTTPAdapter, Retry
retries = Retry(total=5,read=10,other=10,status=10,
                backoff_factor=0.1,
                status_forcelist=[ 502, 503, 504,404,429],
                method_whitelist=["HEAD", "GET", "OPTIONS","POST"])
session.mount('https://', HTTPAdapter(max_retries=retries))
session.mount('http://', HTTPAdapter(max_retries=retries))
```

soxofaan:
indeed, we might consider moving this helper from python driver to python client: https://github.com/Open-EO/openeo-python-driver/blob/b9f465dc276a3d0b0217d7f76a416b7cb943cd38/openeo_driver/util/http.py#L7-L31

and we apparently also have it in MultiBackendJobManager:
https://github.com/Open-EO/openeo-python-client/blob/95b89d5a8ae55d24ed3ba42b70a8abb3d6fa56b7/openeo/extra/job_management.py#L148-L167

soxofaan:
As the original PR #547 was out of sync too much with master, I restarted with new PR #770

PR #770, opened 2025-05-09
Automatically retry on 429 Too Many Requests

soxofaan:
refs #441, #764

### DIFF
--- a/openeo/rest/_connection.py
+++ b/openeo/rest/_connection.py
@@ -5,7 +5,7 @@ import sys
 from typing import Iterable, Optional, Union
 
 import requests
-import requests.adapters
+import urllib3.util
 from requests import Response
 from requests.auth import AuthBase
 
@@ -33,7 +33,7 @@ class RestApiConnection:
         session: Optional[requests.Session] = None,
         default_timeout: Optional[int] = None,
         slow_response_threshold: Optional[float] = None,
-        retry: Union[requests.adapters.Retry, dict, bool, None] = None,
+        retry: Union[urllib3.util.Retry, dict, bool, None] = None,
     ):
         self._root_url = root_url
         self._auth = None
--- a/openeo/rest/_connection.py
+++ b/openeo/rest/_connection.py
@@ -13,7 +13,7 @@ import openeo
 from openeo.rest import OpenEoApiError, OpenEoApiPlainError, OpenEoRestError
 from openeo.rest.auth.auth import NullAuth
 from openeo.util import ContextTimer, ensure_list, str_truncate, url_join
-from openeo.utils.http import session_with_retries
+from openeo.utils.http import HTTP_502_BAD_GATEWAY, session_with_retries
 
 _log = logging.getLogger(__name__)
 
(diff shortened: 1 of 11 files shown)


soxofaan:
TODO:
- [ ] changelog entry
- [ ] allow user to easily configure retry settings (e.g. number of retries, backoff_factor)
- [ ] tests about `job.start_and_wait` loop?

openeo/utils/http.py
```diff
@@ -0,0 +1,96 @@
...
+        429,  # Too Many Requests
+        500,  # Internal Server Error
+        502,  # Bad Gateway
+        503,  # Service Unavailable
+        504,  # Gateway Timeout
+    ]
+)
+
+
+def retry_adapter(
```
soxofaan, line 31: this shouldn't be called `retry_adapter` anymore now

openeo/rest/_connection.py
```diff
@@ -26,15 +28,22 @@ class RestApiConnection:
     def __init__(
         self,
         root_url: str,
+        *,
         auth: Optional[AuthBase] = None,
         session: Optional[requests.Session] = None,
         default_timeout: Optional[int] = None,
         slow_response_threshold: Optional[float] = None,
+        retry: Union[requests.adapters.Retry, dict, bool, None] = None,
```
soxofaan, line 36: `Retry` is actually defined in urllib3, not requests

openeo/utils/http.py
```diff
@@ -0,0 +1,93 @@
...
+# The sleep before the first retry will be skipped however.
+# For example with backoff_factor=2.5, the sleeps between consecutive attempts would be:
+#     0, 5, 10, 20, 40, ...
+DEFAULT_BACKOFF_FACTOR = 2.5
+
+
+DEFAULT_RETRY_FORCELIST = frozenset(
+    [
+        429,  # Too Many Requests
+        500,  # Internal Server Error
```
soxofaan, line 24: Maybe it's best to not include `Internal Server Error` by default for now.
Unlike the others (429/502/503/504)   this is a very general error situation, and not something where we can assume that just waiting a bit and retrying will resolve the problem.

soxofaan:
While working on the tests, it also occurred to me that this feature introduces a bit of conflict with poll-loop situations like `job.start_and_wait()` and the job manager  where there is already skipping of "soft errors" or temporary glitches in some way.
By enabling this `Retry` adapter by default, the user experience worse (e.g. in terms of responsiveness or throughput of the poll-loops) will get worse because of unnecessary but forced sleeps.

So this needs some more investigation (e.g can the retry config be disabled in particular places or contexts?)

openeo/rest/connection.py
```diff
@@ -114,6 +120,16 @@ class Connection(RestApiConnection):
         optional :class:`OidcAuthenticator` object to use for renewing OIDC tokens.
     :param auth: Optional ``requests.auth.AuthBase`` object to use for requests.
         Usage of this parameter is deprecated, use the specific authentication methods instead.
+    :param retry: general request retry settings, can be specified as:
+
+        - :py:class:`urllib3.util.Retry` object
```
soxofaan, line 125: make sure this external urllib3 reference turns in a real link?

Issue #776, opened 2025-05-19
openeo.testing.results: option to ignore large differences in small fraction of pixels


soxofaan:
feature request from @manugv

manugv:
**Issue:** There will always be difference in two simulations due to differences in overlapping pixels in the collection data highlighted in this issue https://github.com/VITO-RS-Vegetation/lcfm-production/issues/191. However, 
1. these differences are only for a small amount of pixels like 0.25% of the total pixels.
2. The difference can be as high as 40% of the mean of the tile. This number is too high to specify `atol/rtol` in openeo-apex-tests. 
  - Setting this high tolerances can be misleading as this tolerance is applicable to all the pixels leading to masking of actual differences.

**Feature request**
A way to specify that, it is acceptable to have 0.25-0.3% of pixels to be different when comparing in openeo-apex-tests.

PR #777, opened 2025-05-20
Added acceptable pixel tolerances feature for testing.

manugv:
Added a feature mentioned in this issue https://github.com/Open-EO/openeo-python-client/issues/776

**Feature behavior** 
- The test should either assert pixels tolerance or (atol, rtol).
- For now it cannot assert both of them at the same time.


openeo/testing/results.py
```diff
@@ -230,7 +233,12 @@ def _compare_xarray_dataarray(
         issues.append(f"Shape mismatch: {actual.shape} != {expected.shape}")
     compatible = len(issues) == 0
     try:
-        xarray.testing.assert_allclose(a=actual, b=expected, rtol=rtol, atol=atol)
+        if pxtol > _DEFAULT_PXTOL:
+            assert (
+                actual != expected
+            ).mean().item() <= pxtol, "Percentage number of pixels that are different is above the threshold"
+        else:
+            xarray.testing.assert_allclose(a=actual, b=expected, rtol=rtol, atol=atol)
```
soxofaan, line 241: I'm not sure this can be a simple if-else construct. It depends on what makes most sense to combine the different tolerances:

e.g. if the pixel tolerance is 5% and you have 6% differing pixels: do you fail the check the hard way, or do you still consider the relative/absolute tolerance?

openeo/testing/results.py
```diff
@@ -194,10 +196,11 @@ def _compare_xarray_dataarray(
     *,
     rtol: float = _DEFAULT_RTOL,
     atol: float = _DEFAULT_ATOL,
+    pxtol: Optional[float] = _DEFAULT_PXTOL,
```
soxofaan, line 199: why `Optional[float]` and not just `float`.  `Optional` means that value `None` is allowed, but that will fail below in the `if pxtol > _DEFAULT_PXTOL:`  check

soxofaan (commented):
some quick notes

I'm also not sure we have to use such a cryptic parameter "pxtol". Note that "rtol" and "atol" are just directly adaptations from numpy/array.
This pixel fraction tolerance is however a completely new parameter, so we can be less cryptic, e.g. "pixel_tolerance"

openeo/testing/results.py
manugv: It was my preference to fail it the hard way. Two reasons 
1. pixel tolerances can be computed before uploading the UDPs so it is known beforehand 
2. If I consider absolute/relative tolerance, it will fail as the differences are too high like 20-30% of max value. So I will get assertion error.

Also, this is mainly happening due to source data is not being right where overlapping pixels are not the same. In my opinion this should not happen but it is good to have a back-up mechanism to stop these failures.

Finally, if you say it would be better to run absolute/relative tolerances then I need to run this first and when I get assertion error,  I can check for pixel tolerance inside "expect AssertionError".

openeo/testing/results.py
manugv: thanks. I will fix this.

openeo/testing/results.py
soxofaan: to put my question differently:  if you set pxtol in the current state of this PR, rtol and atol are completely ignored. If this is the intention, this should be clearly documented.

I however expected this feature a bit different: you ignore x% of the "worst" pixels, but keep comparing the remaining pixels with the existing rtol/atol  tolerances.

> pixel tolerances can be computed before uploading the UDPs so it is known beforehand

I'm not sure what you mean with that and or how atol/rtol could not be computed beforehand?

soxofaan: For example, in openeo-geopyspark-integrationtests we have this somewhat comparable utility:
https://github.com/Open-EO/openeo-geopyspark-integrationtests/blob/2730962957d217bbcd8811308293c41bd8e55652/tests/test_integration.py#L1836-L1853

manugv:
> I think it's also important that this feature is covered with unit tests. Are you familiar with pytest and writing tests?

Yes. I have done it a few times. I will try finding a way to add one. Or will ping you if I need help.

manugv:
> I think it's also important that this feature is covered with unit tests. Are you familiar with pytest and writing tests?

I have added a simple test to capture all possibilities. Please let me know if I have to add more tests.

openeo/testing/results.py
```diff
@@ -230,9 +233,17 @@ def _compare_xarray_dataarray(
         issues.append(f"Shape mismatch: {actual.shape} != {expected.shape}")
     compatible = len(issues) == 0
     try:
-        xarray.testing.assert_allclose(a=actual, b=expected, rtol=rtol, atol=atol)
+        if (pixel_tolerance > _DEFAULT_PIXELTOL) and compatible:
```
soxofaan, line 236: why comparing with `_DEFAULT_PIXELTOL`? It's just about being nonzero, right?
```suggestion
        if pixel_tolerance and compatible:
```

openeo/testing/results.py
```diff
@@ -230,9 +233,17 @@ def _compare_xarray_dataarray(
...
+            _bad_pixels = (actual * 1.0 - expected * 1.0) > atol
```
soxofaan, line 237: You should also take absolute value before comparing I think
and `rtol` should be involved too

Also: no need for the underscore in front of `_bad_pixels`  local variables are "private" per definition

openeo/testing/results.py
```diff
@@ -230,9 +233,17 @@ def _compare_xarray_dataarray(
...
+            assert (
+                _bad_pixels.mean().item() * 100 <= pixel_tolerance
+            ), "Percentage number of pixels that are different is above the threshold"
```
soxofaan, line 240: put the actual percentage and threshold percentage in the error message to make this a helpful message

openeo/testing/results.py
```diff
@@ -474,6 +495,7 @@ def assert_job_results_allclose(
         :py:meth:`~openeo.rest.job.JobResults` object or path to directory with downloaded assets.
     :param rtol: relative tolerance
     :param atol: absolute tolerance
+    :param pixel_tolerance: allowable tolerance for fraction of pixels in percentage
```
soxofaan, line 498:
```suggestion
    :param pixel_tolerance: maximum fraction of pixels (in percent) that is allowed to be significantly different (in terms of ``rtol`` and ``atol``)
```

tests/testing/test_results.py
```diff
@@ -151,6 +151,39 @@ def test_simple_atol(self, actual, atol, expected_issues):
         expected = xarray.DataArray([1, 2, 3])
         assert _compare_xarray_dataarray(actual=actual, expected=expected, rtol=0, atol=atol) == expected_issues
 
+    @pytest.mark.parametrize(
+        ["actual", "atol", "pixel_tolerance", "expected_issues"],
+        [
+            (xarray.DataArray([1, 2, 3.001, 4, 5]), 1e-4, 21.0, []),
```
soxofaan, line 157: make it more pronounced (here and in the other cases) that we are talking about significant differences:
```suggestion
            (xarray.DataArray([1, 2, 300, 4, 5]), 1e-4, 21.0, []),
```

openeo/testing/results.py
manugv: ahh yes.

soxofaan:
FYI: I merged master in your feature branch, which hopefully fixes the unit test failure

openeo/testing/results.py
```diff
@@ -194,10 +196,11 @@ def _compare_xarray_dataarray(
     *,
     rtol: float = _DEFAULT_RTOL,
     atol: float = _DEFAULT_ATOL,
+    pixel_tolerance: float = _DEFAULT_PIXELTOL,
     name: str = None,
 ) -> List[str]:
     """
-    Compare two xarray DataArrays with tolerance and report mismatch issues (as strings)
+    Compare two xarray DataArrays either with tolerance or with allowable pixels and report mismatch issues (as strings)
```
soxofaan, line 203: This doc change is now not necessary anymore I think, as we can keep it general in terms of "tolerance"
```suggestion
    Compare two xarray DataArrays with tolerance and report mismatch issues (as strings)
```

openeo/testing/results.py
```diff
@@ -474,6 +497,7 @@ def assert_job_results_allclose(
         :py:meth:`~openeo.rest.job.JobResults` object or path to directory with downloaded assets.
     :param rtol: relative tolerance
     :param atol: absolute tolerance
+    :param pixel_tolerance: maximum fracton of pixels (in percent) that is allowed to be different (considering ``atol`` and ``rtol``)
```
soxofaan, line 500:
```suggestion
    :param pixel_tolerance: maximum fraction of pixels (in percent) that is allowed to be different (considering ``atol`` and ``rtol``)
```

soxofaan (commented):

PR #778, opened 2025-05-21
Changed the broken href for epsg codes.

manugv:
I changed the broken href http://www.epsg-registry.org/ to https://spatialreference.org/ref/epsg/ (mentioned in the issue https://github.com/Open-EO/openeo-processes/issues/537)
- This also led to a few formatting changes.


examples/archive/notebooks/Sentinel2_NDVI_GEE.ipynb
soxofaan, line 1: I would not touch this file. It's in archive mode (and defunct) and making a change would wrongly suggest it was recently updated to work again

Issue #788, opened 2025-07-16
Unit tests broken since c59fe34d442f919aef18313d1752ea404ead27da


pvbouwel:
Github actions shows unit tests started failing on https://github.com/Open-EO/openeo-python-client/commit/c59fe34d442f919aef18313d1752ea404ead27da

PR #789, opened 2025-07-16
test: adapt unittest to assert new warnings of c59fe34d…

pvbouwel:
…442f919aef18313d1752ea404ead27da

This is for the broken unittests in https://github.com/Open-EO/openeo-python-client/issues/788


soxofaan:
thanks!

Issue #793, opened 2025-08-08
STACAPIJobDatabase persist: support partial updates


soxofaan:
PR #736 is currently blocked because `STACAPIJobDatabase.persist` does not support partial updates as prescribed:
https://github.com/Open-EO/openeo-python-client/blob/4dfdf77ca8bb1158b00562d67591752ef9845054/openeo/extra/job_management/__init__.py#L77-L80

Instead it expects all columns to be present in the provided dataframe.

PR #736 depends on the ability of just updating a subset of columns through `persist`

soxofaan:
While wrestling with this under #794 and #798  I actually start to wonder what "partial" actually means here: does it mean subset of columns in subset of rows, or does it means just a subset of rows (with all columns filled in).

If it's the latter (not unlikely), then `STACAPIJobDatabase.persist` might actually be just fine as is.
However, #736 currently builds on the former interpretation

soxofaan:
Let's go for the interpretation "just a subset of rows (with all columns filled in).".

advantage: only one place to implement updating existing data with updates (instead of doing it for each job db implementation).
disadvantage: requires introduction of new `JobDatabaseInterface.get_by_indices` API, but it's probably useful to have this anyway (and not hard to implement).

more details:  https://github.com/Open-EO/openeo-python-client/pull/736/commits/2844e403f00ff349da0a5f2de6dec65bda585457

this closes this ticket I think

PR #794, opened 2025-08-08
Support partial updates in STACAPIJobDatabase.persist

soxofaan:
- for #793: 
- related to unblocking #736

### DIFF
--- a/openeo/extra/job_management/stac_job_db.py
+++ b/openeo/extra/job_management/stac_job_db.py
@@ -61,12 +61,16 @@ class STACAPIJobDatabase(JobDatabaseInterface):
         """
         df = MultiBackendJobManager._normalize_df(df)
 
-        if isinstance(df.index, pd.RangeIndex) and "item_id" in df.columns:
-            # Support legacy usage: default (autoincrement) index and an "item_id" column -> copy over as index
+        # Support legacy usage: default (autoincrement) index and an "item_id" column -> copy over as index
+        if "item_id" not in df.columns:
+            df = df.reset_index(names=["item_id"])
             df.index = df["item_id"]
 
         # Make sure the index (of item ids) are strings, to play well with (py)STAC schemas
         df.index = df.index.astype(str)
+        
+        # Ensure the index is set as a column named "item_id" for backward compatibility
+        df["item_id"] = df.index
         return df
 
     def initialize_from_df(self, df: pd.DataFrame, *, on_exists: str = "error"):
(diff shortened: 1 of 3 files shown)


HansVRP:
main issue seems to be related to the item_id moving from a column to the dataframe index; causing a mismatch in size. 

and the item_id is no longer popped out

openeo/extra/job_management/stac_job_db.py
```diff
@@ -117,46 +129,46 @@ def series_from(self, item: pystac.Item) -> pd.Series:
...
+            "id": item_id,
+            "properties": series_dict,
+            "geometry": series[self.geometry_column] if self.has_geometry else None,
+            "links": [],
+            "assets": {}
+        }
+        
+        # Handle datetime
+        dt = series_dict.get("datetime")
+        if not dt:
```
soxofaan, line 164: I understand the need to bring back the `series_dict.pop("item_id")`, but are the other changes relevant here?

I'd like to avoid that this PR review also spirals out of scope

openeo/extra/job_management/stac_job_db.py
```diff
@@ -117,46 +129,46 @@ def series_from(self, item: pystac.Item) -> pd.Series:
...
             item_dict["properties"]["datetime"] = pystac.utils.datetime_to_str(datetime.datetime.now())
-
-        if self.has_geometry:
-            item_dict["geometry"] = series[self.geometry_column]
-        else:
-            item_dict["geometry"] = None
```
soxofaan, line 150: is removal of these lines relevant here?

as noted, I'd like to keep this PR focused to avoid it strands in eternal review

openeo/extra/job_management/stac_job_db.py
soxofaan: > I understand the need to bring back the series_dict.pop("item_id")

on further consideration: I'd like to reconsider:
"item_id" as column name has no special meaning anymore, so it should not get special treatment  (meaning it should _not_ be popped)

openeo/extra/job_management/stac_job_db.py
```diff
@@ -180,35 +199,45 @@ def get_by_status(self, statuses: Iterable[str], max: Optional[int] = None) -> p
...
+        )
+        existing_df = self._search_result_to_df(existing_items)
 
-            df.apply(handle_row, axis=1)
+        if existing_df.empty:
+            df_to_persist = df
+        else:
+            # Merge data on item_id (in the index)
+            df_to_persist = existing_df
+            df_to_persist.update(df, overwrite=True)
```
soxofaan, line 236: While working on test coverage, it turned out that this pandas `update` might cause data loss:
it only updates the intersection of both dataframes, so if there is a mismatch between items in `existing_df` and `df`, there will be less updates than expected

PR #805, opened 2025-09-08
feature: handle expired token gracefully for artifact helper

pvbouwel:
If a token is expired we can refresh it by doing another OpenEO interaction. By allowing 3 attempts for the STS service and by doing a very wide except clause we also make ourselves more robust against intermittent errors.

### DIFF
--- a/openeo/extra/artifacts/_s3sts/sts.py
+++ b/openeo/extra/artifacts/_s3sts/sts.py
@@ -1,5 +1,8 @@
 from __future__ import annotations
 
+import logging
+from random import randint
+from time import sleep
 from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
@@ -12,12 +15,16 @@ from openeo.rest.auth.auth import BearerAuth
 from openeo.rest.connection import Connection
 from openeo.util import Rfc3339
 
+_log = logging.getLogger(__name__)
+
 
 class OpenEOSTSClient:
+    _MAX_STS_ATTEMPTS = 3
+
     def __init__(self, config: S3STSConfig):
         self.config = config
 
-    def assume_from_openeo_connection(self, connection: Connection) -> AWSSTSCredentials:
+    def assume_from_openeo_connection(self, connection: Connection, attempt: int = 0) -> AWSSTSCredentials:
         """
         Takes an OpenEO connection object and returns temporary credentials to interact with S3
         """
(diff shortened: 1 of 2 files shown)


openeo/extra/artifacts/_s3sts/sts.py
```diff
@@ -27,14 +34,32 @@ def assume_from_openeo_connection(self, connection: Connection) -> AWSSTSCredent
...
+                    RoleSessionName=f"artifact-helper-{Rfc3339().now_utc()}",
+                    WebIdentityToken=auth_token[2],
+                    DurationSeconds=43200,
+                )
             )
-        )
+        except Exception as e:
+            _log.warning("Failed to get credentials for STS access")
+
+            if attempt < 3:
```
soxofaan: you probably want to use `_MAX_STS_ATTEMPTS` here?

openeo/extra/artifacts/_s3sts/sts.py
```diff
@@ -1,5 +1,8 @@
 from __future__ import annotations
 
+import logging
+from random import randint
+from time import sleep
```
soxofaan: FYI I usually prefer just to `import time` here and use `time.sleep()` at call time, with these marginal benefits
- minimize number of imported symbols at top level to keep things tidier
- simplifies mocking of `sleep` in testing as it better decouples the tests from how the utilities were imported (e.g. `mock.patch("time.sleep")` would just work, while with `from time import sleep` you would have to worry more about implementation details)

openeo/extra/artifacts/_s3sts/sts.py
```diff
@@ -27,14 +34,32 @@ def assume_from_openeo_connection(self, connection: Connection) -> AWSSTSCredent
...
+            if attempt < 3:
+                # backoff with jitter
+                max_sleep_ms = 500 * (2**attempt)
+                sleep_ms = randint(0, max_sleep_ms)
+                _log.info(f"Retrying STS access in {sleep_ms} ms")
+                sleep(sleep_ms / 1000.0)
+                attempt += 1
+                # Do an API call with OpenEO to trigger a refresh of our token.
+                connection.describe_account()
+                _log.info("Retrying to get credentials for STS access")
```
soxofaan: maybe add attempt counter to message?

openeo/extra/artifacts/_s3sts/sts.py
```diff
@@ -27,14 +34,32 @@ def assume_from_openeo_connection(self, connection: Connection) -> AWSSTSCredent
...
+                return self.assume_from_openeo_connection(connection, attempt)
+            else:
+                _log.fatal("Maximum attempts performed for STS access", exc_info=e)
+                raise RuntimeError("Could not get credentials from STS") from e
```
soxofaan, line 61: doing a "fatal" log and an exception looks a bit like duplication (the exception will likely also be logged at higher level). And "fatal" is maybe a bit much, I don't think we go higher than "error" elsewhere

soxofaan (commented):
some quick notes

openeo/extra/artifacts/_s3sts/sts.py
pvbouwel: Will, change; I appreciate the explanation

PR #815, opened 2025-10-03
Issue741 movedatabasecode

HansVRP:
first proposal to seperate the job database/dataframe logic and the processbasedjobstart


openeo/extra/job_management/__init__.py
```diff
@@ -207,21 +148,7 @@ def start_job(
...
-        "running_start_time": _ColumnProperties(dtype="str"),
-        # TODO: these columns "cpu", "memory", "duration" are not referenced explicitly from MultiBackendJobManager,
-        #       but are indirectly coupled through handling of VITO-specific "usage" metadata in `_track_statuses`.
-        #       Since bfd99e34 they are not really required to be present anymore, can we make that more explicit?
-        "cpu": _ColumnProperties(dtype="str"),
-        "memory": _ColumnProperties(dtype="str"),
-        "duration": _ColumnProperties(dtype="str"),
-        "costs": _ColumnProperties(dtype="float64"),
-    }
+
```
soxofaan: The three comments above should be moved along with `_COLUMN_REQUIREMENTS`

openeo/extra/job_management/_job_database.py
```diff
@@ -0,0 +1,366 @@
...
+        :param indices: List of indices to include.
+
+        :return: DataFrame with jobs filtered by indices.
+        """
+        ...
+
+# Expected columns in the job DB dataframes.
+# TODO: make this part of public API when settled?
+# TODO: move non official statuses to seperate column (not_started, queued_for_start)
+COLUMN_REQUIREMENTS = {
```
soxofaan: I don't think `COLUMN_REQUIREMENTS` belongs under the `_job_database` module. 
The requirements describe what the job manager expects as columns, so this should stay close to the job manager module (`__init__.py` at the moment)

soxofaan (commented):
some quick notes

openeo/extra/job_management/_job_database.py
HansVRP: I agree that sematic wise it should live closer to the job manager and the job database should be as agnostic as possible. 

It does introduce a circular dependency when keeping it in _init_.

Should we create a constants.py where we capture these parameters?

HansVRP: the discussed split will still have some issues; 

dataframe/database depend on the column properties of the job manager. 

job manager depends on database code for get_job_db. 

Will need to figure out the cleanest way to seperate

soxofaan:
to not  let this linger too long, I just did some follow-up commits, did a rebase and merged in 8bd7935776631544f1c4c3e0defd3f2394502a0f

Issue #817, opened 2025-10-13
threaded tasks in job manager: support auto-refresh of auth token


VictorVerhaert:
During a large scale production we end up with blocks where jobs don't get started due to `Failed to start job 'j-2510101449364f5e9a844d1bc59e2653': OpenEoApiError('[401] TokenInvalid: Authorization token has expired or is invalid. Please authenticate again.')`

<img width="1645" height="727" alt="Image" src="https://github.com/user-attachments/assets/33a4d42f-95bc-41cb-a407-3b20a89ed693" />
The blocks are rather short making me think it only impacts all the threads that have already been created.

I think it might have something to do with the auth bearer not being thread safe when refreshing.
Adding a token getting in the thread with a locked getter (thread safe) could solve this, but it is certainly not the only solution here

PR #818, opened 2025-10-15
proactively refresh bearer for threaded task

soxofaan:
for 
- #817


VincentVerelst:
8 hours is indeed exactly the interval at which I encountered this problem. I anyway need to reprocess some jobs for a project. I'm running those jobs now on this branch. Will keep you posted!

PR #821, opened 2025-10-22
Fix access to federation properties for services

m-mohr:
Previously, the method call would return an error because the federation details can't be read from a List. Fixed the issue by passing in the full response, not just the array of services.


soxofaan:
Indeed, I added tests
and merged in 7d11a4ed3e45e17bebf316a1e783b74a03cdfc6e

Issue #741, opened 2025-02-26
Job manager: Move jobdatabase classes out of __init__


VictorVerhaert:
the __init__ file currently contains a lot of logic making it hard for new users to figure out how it works.

One way of cleaning this file is moving job database logic to a separate file `job_database.py' (and including the stac api one)

In order to prevent existing links from breaking we can import the databases and add them to the __all__ variable as described here:
https://docs.astral.sh/ruff/rules/unused-import/
(I realize we don't use ruff as a linting tool but still believe we should add it to __all__)

soxofaan:
fully agree to move all implementation out of `job_management/__init__.py` into submodules

However, there are a couple of ongoing job manager PRs, so we have to be careful when to start and merge the move to avoid branch conflicts

HansVRP:
one issue I forsee is a potential circular import; 

FullDataFrameJobDatabase depends on the multibackend job manager's _normalize_df

soxofaan:
just merged #815  so I think this one can be closed 👍

PR #824, opened 2025-10-30
 Move ProcessBasedJobCreator to own submodule

soxofaan:
part of 
- #741

Split off ProcessBasedJobCreator to own submodule

inspired by #815


tests/extra/job_management/test_process_based.py
```diff
@@ -0,0 +1,709 @@
...
+import pandas as pd
+import pytest
+
+import openeo
+from openeo import BatchJob
+from openeo.extra.job_management import (
+    CsvJobDatabase,
+    MultiBackendJobManager,
+    ParquetJobDatabase,
+    ProcessBasedJobCreator,
```
soxofaan: import this from actual submodule

PR #826, opened 2025-11-04
Implement list_auth_providers

m-mohr:
This implements a new method list_auth_providers so that clients can get a full list of authentication methods/providers easily.

Planned to be used in e.g. the openEO QGIS plugin, which plans to mimic the authentication interface of the Web Editor:

{452A3336-0A1E-4A99-BC3D-C1A37DCAC8F8}

For this, we need a way to retrieve all auth methods. We could also have this code in the plugin itself, but it felt better suited here. Considering also, that the other clients have similar functions.


openeo/rest/connection.py
```diff
@@ -211,6 +212,37 @@ def _get_refresh_token_store(self) -> RefreshTokenStore:
...
         return self._refresh_token_store
 
+    def list_auth_providers(self) -> list[dict]:
+        providers = []
+        cap = self.capabilities()
+
+        # Add OIDC providers
+        oidc_path = "/credentials/oidc"
+        if cap.supports_endpoint(oidc_path, method="GET"):
+            try:
```
soxofaan, line 226: I would not guard this with a try-except. 
If this request goes wrong, you want to know about that in full detail as a caller (instead of getting an empty list). I would even not do the `if isinstance(data, dict)`

soxofaan: essentially, you can condense this down to two lines:
```python
data = self.get(oidc_path, expected_status=200).json()
providers.extend({**p, "type": "oidc"} for p in data["providers"])
```

openeo/rest/connection.py
```diff
@@ -211,6 +212,37 @@ def _get_refresh_token_store(self) -> RefreshTokenStore:
...
+                        providers.append(provider)
+            except OpenEoApiError:
+                pass
+
+        # Add Basic provider
+        basic_path = "/credentials/basic"
+        if cap.supports_endpoint(basic_path, method="GET"):
+            providers.append(
+                {
+                    "id": uuid.uuid4().hex,
```
soxofaan: instead of a uuid, I would just use `basic_path` here as well. Change on collision in practice is astronomically tiny 😄

openeo/rest/connection.py
```diff
@@ -211,6 +212,37 @@ def _get_refresh_token_store(self) -> RefreshTokenStore:
...
+                    "issuer": self.build_url(basic_path),
+                    "type": "basic",
+                    "title": "Internal",
```
soxofaan, line 243: I'd rather call this "Basic Auth".
"Internal" could also be useful for one of the OIDC provider options (e.g. for special use cases where you have a special, VPN-guarded OIDC provider for developers/testers/machine-to-machine-auth/...)

tests/rest/test_connection.py
```diff
@@ -767,6 +767,59 @@ def test_create_connection_lazy_refresh_token_store(requests_mock):
...
+def test_list_auth_providers(requests_mock, api_version):
+    requests_mock.get(
+        API_URL,
+        json={
+            "api_version": api_version,
+            "endpoints": [
+                {"methods": ["GET"], "path": "/credentials/basic"},
+                {"methods": ["GET"], "path": "/credentials/oidc"},
+            ],
+        },
```
soxofaan: FYI: there a helper to build this response with:
```python
build_capabilities(api_version=api_version, basic_auth=True, oidc_auth)
```

tests/rest/test_connection.py
```diff
@@ -767,6 +767,59 @@ def test_create_connection_lazy_refresh_token_store(requests_mock):
...
+    assert p2["type"] == "oidc"
+    assert p2["issuer"] == "https://other.example"
+    assert p2["title"] == "Other"
+
+    basic = next(filter(lambda x: x["type"] == "basic", providers), None)
+    assert isinstance(basic, dict)
+    assert isinstance(basic["id"], str)
+    assert len(basic["id"]) > 0
+    assert basic["issuer"] == API_URL + "credentials/basic"
+    assert basic["title"] == "Internal"
```
soxofaan, line 803: instead of using separate asserts, one for each fields, I prefer a single assert that checks the response as a whole, which is perfectly doable here, e.g.

```python
assert providers == [
    {
        "type": "oidc", 
        "issuer": ...
        "title": ...
    },
    {
         ...
]
```

The advantage of this is that you can easily see and understand the expected output.
And when there is mismatch,  pytest will visualize the difference in a very accessible way (when verbosity level is high enough, e.g. see https://docs.pytest.org/en/stable/how-to/output.html#verbosity)

openeo/rest/connection.py
m-mohr: Fine for me, was just worried about the tiny chance of a collision. I mean doesn't hurt to use uuid either, right?

openeo/rest/connection.py
m-mohr: The Web Editor calls it "Internal" so thought it might be more consistent and intuitive across clients, and also easier to document potentially. I think if backends want to use "Internal" as name for OIDC, maybe just disable Basic?

But if you prefer it, we can change it. We could then just update to "Internal" in the QGIS plugin by checking the ID, but it would lead to the same result eventually (backends would have a "conflict" with "Internal") ;-)

tests/rest/test_connection.py
m-mohr: And oidc_auth would be the dict below that contains the providers?

tests/rest/test_connection.py
m-mohr: I wasn't sure whether this might lead to problems when the (stub) API (for whatever reason) would change the order of the array?!

openeo/rest/connection.py
m-mohr: Hmm, I think I'd find it a bit more usable and user friendly if it doesn't throw errors but instead just offers me what's currently available (at least e.g. in the QGIS UI experience). I understand that in a Python coding environment an error might be preferrable. We could also send a warning instead of doing the pass?

openeo/rest/connection.py
```diff
@@ -9,6 +9,7 @@
 import os
 import shlex
 import urllib.parse
+import uuid
```
m-mohr:
```suggestion
```

openeo/rest/connection.py
m-mohr:
```suggestion
                    "id": basic_path,
```

openeo/rest/connection.py
```diff
@@ -211,6 +212,37 @@ def _get_refresh_token_store(self) -> RefreshTokenStore:
...
+        oidc_path = "/credentials/oidc"
+        if cap.supports_endpoint(oidc_path, method="GET"):
+            try:
+                data = self.get(oidc_path, expected_status=200).json()
+                if isinstance(data, dict):
+                    for provider in data.get("providers", []):
+                        provider["type"] = "oidc"
+                        providers.append(provider)
+            except OpenEoApiError:
+                pass
```
m-mohr:
```suggestion
            except OpenEoApiError as e:
                warnings.warn(f"Unable to load the OpenID Connect provider list: {e.message}")
```

openeo/rest/connection.py
m-mohr: @niebl Depending on the outcome, you may need to reflect this in the QGIS plugin code.

openeo/rest/connection.py
m-mohr: @niebl Depending on the outcome, you may need to reflect this in the QGIS plugin code.

openeo/rest/connection.py
soxofaan: with UUID you would get a different id each time without a good reason (it's still exactly the same provider), while further tooling might desire a bit of consistency here

openeo/rest/connection.py
soxofaan: to be honest, I also find "Internal" in de web editor a bit confusing, especially because the openEO API does not use this terminology at all in this context

and the tool tip also doesn't help :stuck_out_tongue: 
image


but anyway, this is probably for another discussion/thread, so I'm fine with keeping it "internal" for now

tests/rest/test_connection.py
soxofaan: no sorry, typo:
```python
build_capabilities(api_version=api_version, basic_auth=True, oidc_auth=True)
```

this just builds the `GET /` document, you still have to build the `/authentication/oidc` doc seperately

openeo/rest/connection.py
soxofaan: In practice, the `/credentials/oidc` at backend-side is a very simple endpoint to provide, so if that would fail there are probably larges problems at play (e.g. whole backend down).

Also, `/credentials/oidc` does not make any guarantees about the actual availability of the listed OIDC providers, so even if you catch errors at this level, there is still a lot more that can go wrong in the context of availability.


Anyway, if you keep the try-except, I would definitely add logging of a warning instead of silently skipping the problem

openeo/rest/connection.py
soxofaan: we try to avoid usage `warnings` (I see we still have some cases of that in connection.py), as it generally composes not that good with other tooling. Instead use `_log.warning()`.

Also when including the exception in the warning message, we typically just do `{e!r}` so that more useful info is included (error code, http code, error message, correlation id )

so:
```
_log.warning(f"Unable to load the OpenID Connect provider list: {e!r}")
```

(`!r` is a shortcut for generic `repr()`-style rendering of the exception)

openeo/rest/connection.py
m-mohr: That tooltip is a sideeffect of another change it seems. 🙈 Unintentional to have it there.

openeo/rest/connection.py
m-mohr: Added a warning for now.

Must say I don't like the {e!r} though.

Showing in a GUI (like QGIS, not Python coding)
> Unable to load the OpenID Connect provider list: OpenEoApiError('[500] Internal: Maintanence ongoing')

feels like bad UX compared to:
> Unable to load the OpenID Connect provider list: Maintanence ongoing')

But if the former is consistently done everywhere, I guess that's how it shall be for now...

openeo/rest/connection.py
m-mohr: Added a warning.

openeo/rest/connection.py
```diff
@@ -211,6 +211,37 @@ def _get_refresh_token_store(self) -> RefreshTokenStore:
             self._refresh_token_store = RefreshTokenStore()
         return self._refresh_token_store
 
+    def list_auth_providers(self) -> list[dict]:
```
soxofaan:
```suggestion
    def list_auth_providers(self) -> List[dict]:
```

`list[dict]` is only valid since python 3.9, python client still supports 3.8

not sure actually why CI doesn't catch this

PR #828, opened 2025-11-07
Issue744 download optional

HansVRP:


openeo/extra/job_management/_manager.py
```diff
@@ -157,10 +157,14 @@ def start_job(
             - get_error_log_path
             - get_job_metadata_path
 
+    :param download_results:
```
soxofaan, line 160: maybe also add a `.. versionchanged:: ` note below

and a changelog entry?

HansVRP:
unclear why the unit tests are failing

PR #831, opened 2025-11-25
allows output of run_udf to be used as temporal extent in other processes

VictorVerhaert:


VictorVerhaert:
Other PR looks a lot more general and supports test cases so closing this one

PR #834, opened 2025-12-10
Support UDF based spatial/temporal extents in load_collection/load_stac

soxofaan:
alternative PR for
- #831

### DIFF
--- a/openeo/rest/connection.py
+++ b/openeo/rest/connection.py
@@ -35,7 +35,12 @@ from requests.auth import AuthBase, HTTPBasicAuth
 import openeo
 from openeo.config import config_log, get_config_option
 from openeo.internal.documentation import openeo_process
-from openeo.internal.graph_building import FlatGraphableMixin, PGNode, as_flat_graph
+from openeo.internal.graph_building import (
+    FlatGraphableMixin,
+    PGNode,
+    _FromNodeMixin,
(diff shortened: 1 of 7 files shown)


VictorVerhaert:
created an issue to create a community notebook for this here: https://github.com/Open-EO/openeo-community-examples/issues/90

openeo/rest/connection.py
```diff
@@ -1186,8 +1191,8 @@ def load_collection(
         self,
         collection_id: Union[str, Parameter],
         spatial_extent: Union[dict, Parameter, shapely.geometry.base.BaseGeometry, str, Path, None] = None,
-        temporal_extent: Union[Sequence[InputDate], Parameter, str, None] = None,
-        bands: Union[Iterable[str], Parameter, str, None] = None,
+        temporal_extent: Union[Sequence[InputDate], Parameter, str, _FromNodeMixin, None] = None,
```
soxofaan: on second thought, it's probably more correct to use `FlatGraphableMixin` here (and other places) instead of `_FromNodeMixin`

soxofaan: nope, `_FromNodeMixin` is fine. 
Its naming is however confusing (it's more an interface than a mixin, and "from node" is also easy to misunderstand)

PR #836, opened 2025-12-11
Issue816 threadeddownload

HansVRP:


openeo/extra/job_management/_manager.py
```diff
@@ -219,6 +221,7 @@ def __init__(
         )
         self._thread = None
         self._worker_pool = None
+        self._download_pool = None
```
soxofaan: Is it necessary to have a separate pool for downloads? Isn't it possible to just use the existing worker pool?

It seems from the rest of the PR that each pool comes with a lot of (duplicate) code overhead and housekeeping

openeo/extra/job_management/_thread_worker.py
```diff
@@ -140,7 +142,46 @@ def execute(self) -> _TaskResult:
...
+class _JobDownloadTask(ConnectedTask):
+    """
+    Task for downloading job results and metadata.
+
+    :param download_dir:
+        Root directory where job results and metadata will be downloaded.
+    """
+    def __init__(self, download_dir: Path, **kwargs):
+        super().__init__(**kwargs)
+        object.__setattr__(self, 'download_dir', download_dir)
```
soxofaan: I don't think you need this complicated `__init__`. 
You can just do something like this:

```python
@dataclasses.dataclass(frozen=True)
class _JobDownloadTask(Task):
    download_dir: Path

    def execute....
```

tests/extra/job_management/test_thread_worker.py
```diff
@@ -79,6 +83,103 @@ def test_hide_token(self, serializer):
         assert "job-123" in serialized
         assert secret not in serialized
 
+class TestJobDownloadTask:
+    
+    # Use a temporary directory for safe file handling
+    @pytest.fixture
+    def temp_dir(self):
+        with tempfile.TemporaryDirectory() as temp_dir:
+            yield Path(temp_dir)
```
soxofaan: no need for this: pytest comes with a built-in `tmp_path` fixture

soxofaan (commented):
some quick notes

openeo/extra/job_management/_manager.py
HansVRP: it indeed brings quite some duplication; merging the pools however would risk that downloads block launching new jobs

HansVRP: perhaps a solution is to introduce a semaphore

openeo/extra/job_management/_thread_worker.py
```diff
@@ -150,12 +191,13 @@ class _JobManagerWorkerThreadPool:
 
     :param max_workers:
         Maximum number of concurrent threads to use for execution.
-        Defaults to 2.
+        Defaults to 1.
     """
 
-    def __init__(self, max_workers: int = 2):
+    def __init__(self, max_workers: int = 1, name: str = 'default'):
```
soxofaan: It's not necessary to give a a name here right? or how would that be used?

openeo/extra/job_management/_thread_worker.py
```diff
@@ -139,9 +141,48 @@ def execute(self) -> _TaskResult:
...
+            metadata_path = self.download_dir / f"job_{self.job_id}.json"
+            with metadata_path.open("w", encoding="utf-8") as f:
+                json.dump(job_metadata, f, ensure_ascii=False)
+            
+            _log.info(f"Job {self.job_id!r} results downloaded successfully")
+            return _TaskResult(
+                job_id=self.job_id,
+                df_idx=self.df_idx,
+                db_update={}, #TODO consider db updates?
+                stats_update={"job download": 1},
```
soxofaan: you probably also want to keep track of the number of files downloaded, not just the number of jobs that were downloaded from

openeo/extra/job_management/_thread_worker.py
```diff
@@ -207,8 +249,91 @@ def process_futures(self, timeout: Union[float, None] = 0) -> Tuple[List[_TaskRe
 
         self._future_task_pairs = to_keep
         return results, len(to_keep)
+    
+    def num_pending_tasks(self) -> int:
```
soxofaan: abbreviating "number" to "num" is a bit off brand with the rest of the openEO client API.
In this age of large monitors and AI powered autocomplete there is little reason anymore to save on keystrokens at cost of readability.
I'd go for `pending_task_count` which is is just 2 characters more 😄

openeo/extra/job_management/_thread_worker.py
```diff
@@ -207,8 +249,91 @@ def process_futures(self, timeout: Union[float, None] = 0) -> Tuple[List[_TaskRe
...
+
+    """
+    Generic wrapper that manages multiple thread pools with a dict.
+    Uses task class names as pool names automatically.
+    """
+    
+    def __init__(self, pool_configs: Optional[Dict[str, int]] = None):
+        """
+        :param pool_configs: Dict of task_class_name -> max_workers
+                            Example: {"_JobStartTask": 1, "_JobDownloadTask": 2}
```
soxofaan: I would not use the class name as key in this dict, then we are leaking ugly internals to the end user. I would just use "default", and "download"

openeo/extra/job_management/_thread_worker.py
```diff
@@ -207,8 +249,91 @@ def process_futures(self, timeout: Union[float, None] = 0) -> Tuple[List[_TaskRe
...
+        """
+        self._pools: Dict[str, _TaskThreadPool] = {}
+        self._pool_configs = pool_configs or {}
+    
+    def _get_pool_name_for_task(self, task: Task) -> str:
+        """
+        Get pool name from task class name.
+        """
+        return task.__class__.__name__
```
soxofaan: I would not make the pool name function of the task class. I would just use explicit names when calling "submit_task" from the job manager.
This also decouples the pool names from the task class names

openeo/extra/job_management/_thread_worker.py
```diff
@@ -207,8 +249,91 @@ def process_futures(self, timeout: Union[float, None] = 0) -> Tuple[List[_TaskRe
...
+        """
+        all_results = []
+        remaining_by_pool = {}
+        
+        for pool_name, pool in self._pools.items():
+            results, remaining = pool.process_futures(timeout)
+            all_results.extend(results)
+            remaining_by_pool[pool_name] = remaining
+            
+        return all_results, remaining_by_pool
```
soxofaan: does the job manager actually need the "remaining" part?
I think we can omit that to keep it simple

HansVRP:
good to merge @soxofaan ?

openeo/extra/job_management/_thread_worker.py
```diff
@@ -150,12 +194,13 @@ class _JobManagerWorkerThreadPool:
...
         Maximum number of concurrent threads to use for execution.
-        Defaults to 2.
+        Defaults to 1.
     """
 
-    def __init__(self, max_workers: int = 2):
+    def __init__(self, max_workers: int = 1, name: str = 'default'):
         self._executor = concurrent.futures.ThreadPoolExecutor(max_workers=max_workers)
         self._future_task_pairs: List[Tuple[concurrent.futures.Future, Task]] = []
+        self._name = name
```
soxofaan: What is the purpose of this name? I don't see it being used somewhere

openeo/extra/job_management/_thread_worker.py
```diff
@@ -150,12 +194,13 @@ class _JobManagerWorkerThreadPool:
 
     :param max_workers:
         Maximum number of concurrent threads to use for execution.
-        Defaults to 2.
+        Defaults to 1.
```
soxofaan: Is there a reason to change this default?

openeo/extra/job_management/_thread_worker.py
```diff
@@ -206,9 +251,90 @@ def process_futures(self, timeout: Union[float, None] = 0) -> Tuple[List[_TaskRe
         _log.info("process_futures: %d tasks done, %d tasks remaining", len(results), len(to_keep))
 
         self._future_task_pairs = to_keep
-        return results, len(to_keep)
+        return results
```
soxofaan: Is there an important reason to remove `len(to_keep)` here?

Also note that return type annotation and documentation are not updated yet here

openeo/extra/job_management/_thread_worker.py
```diff
@@ -206,9 +251,90 @@ def process_futures(self, timeout: Union[float, None] = 0) -> Tuple[List[_TaskRe
...
+    
+    def number_pending_tasks(self) -> int:
+        """Return the number of tasks that are still pending (not completed)."""
+        return len(self._future_task_pairs)
```
soxofaan: Note that separating `number_pending_tasks` from `process_futures` might be risky/misleading as `number_pending_tasks` only makes sense if `process_futures` has run recently enough, otherwise it will return possibly outdated information.

openeo/extra/job_management/_thread_worker.py
soxofaan: why is this marked resolved, I don't see a change here?

openeo/extra/job_management/_thread_worker.py
```diff
@@ -206,9 +251,90 @@ def process_futures(self, timeout: Union[float, None] = 0) -> Tuple[List[_TaskRe
...
+        Process updates from ALL pools.
+        Returns: (all_results, dict of remaining tasks per pool)
+        """
+        all_results = []
+        
+        for pool_name, pool in self._pools.items():
+            results = pool.process_futures(timeout)
+            all_results.extend(results)
+            
+        return all_results
```
soxofaan: return type does not correspond with type annotation here

openeo/extra/job_management/_thread_worker.py
HansVRP: I use this name for the initiation of multiple pools e.g: self._worker_pool.submit_task(task=task, pool_name="job_start")

openeo/extra/job_management/_thread_worker.py
HansVRP: this followed from one of our offline discussion, happy to set it back to 2

openeo/extra/job_management/_thread_worker.py
HansVRP: true, main purpose is to avoid shutting down a thread which still has work to do. Either way we wait on the futures before closing down.

openeo/extra/job_management/_thread_worker.py
HansVRP: removed it here as I mainly relied on the 'number of pending tasks' instead

HansVRP:
@soxofaan good to rereview or merge if you agree. 

I did notice in my previous commit some latency when managing the multiple pools (see failed unit test)

PR #839, opened 2025-12-17
Issue838 enforece limited queue

HansVRP:


openeo/extra/job_management/_manager.py
```diff
@@ -539,10 +539,11 @@ def _job_update_loop(
             stats["track_statuses"] += 1
 
         not_started = job_db.get_by_status(statuses=["not_started"], max=200).copy()
-        if len(not_started) > 0:
+        pending = job_db.get_by_status(statuses=["created", "queued", "queued_for_start"], max=200).copy()
+
+        if len(not_started) > 0 and len(pending) < 10:
```
soxofaan: I think there is still risk on overflow here:
e.g. pending count is 9 and you have 2 "not started" jobs -> you'll start these 2 -> possibly hitting 11 "queued"

openeo/extra/job_management/_manager.py
```diff
@@ -539,10 +539,11 @@ def _job_update_loop(
             stats["track_statuses"] += 1
 
         not_started = job_db.get_by_status(statuses=["not_started"], max=200).copy()
-        if len(not_started) > 0:
+        pending = job_db.get_by_status(statuses=["created", "queued", "queued_for_start"], max=200).copy()
```
soxofaan: the copy at the end is not necessary I think

openeo/extra/job_management/_manager.py
soxofaan: and pending also counts across multiple backends, while the limit 10 is per backend

HansVRP: ah true, standard usage is on one backend

soxofaan:
FYI: I pushed a commit to simplify the run/queuing logic a bit (and merge master)

soxofaan:
I also noticed that some tests are getting more flaky with these changes,
not sure yet what is going on

soxofaan:
> checking if it can be related to omitting this 'created' state

I was also looking at this, and it is indeed a subtle interaction between 
- not counting "created" as "running"
- mixing openeo job status (queued, running) and internal status (e.g. queued_for_start) in the same "status" column"
- unpredictability of threading

Apparently, it can happen that the "status" of a job goes from "queued_for_start" to "created".
If this happens on backend "bar", the job manager (wrongly) thinks that "bar" has capacity, and launches the last job there, while it used to launch it on "foo" (which was set up to actually finish  faster than "bar")

PR #851, opened 2026-01-13
expose the queue limit when registering a backend in the JobManager.

VictorVerhaert:


openeo/extra/job_management/_manager.py
```diff
@@ -246,6 +247,8 @@ def add_backend(
             Either a Connection to the backend, or a callable to create a backend connection.
         :param parallel_jobs:
             Maximum number of jobs to allow in parallel on a backend.
+        :param queueing_limit:
+            Maximum number of jobs to allow in queue on a backend. Setting this too high may lead to a backend rejecting jobs.
```
soxofaan, line 251:
```suggestion
            Maximum number of queued jobs (on the backend) the job manager should stay below before starting more jobs. Setting this too high may lead to the backend rejecting jobs.
```

PR #852, opened 2026-01-13
Preserve URL path components for subpath endpoints.

pantierra:
Connecting to OpenEO backends at subpaths (e.g., `https://example.com/openeo`) fails because the library strips the path component:
- `url_join()` strips root path when joining absolute paths
- `version_discovery()` doesn't preserve original URL path when backend returns URL without path

Proposing the following fixes:

- **`openeo/util.py`**: Preserve root's path when joining absolute paths
- **`openeo/rest/connection.py`**: Preserve original URL path if discovered URL lost it

### DIFF
--- a/openeo/rest/connection.py
+++ b/openeo/rest/connection.py
@@ -187,6 +187,7 @@ class Connection(RestApiConnection):
         :param url: initial backend url (not including "/.well-known/openeo")
         :return: root url of highest supported backend version
         """
+        original_url = url
         try:
             connection = RestApiConnection(url, session=session, retry=retry)
             well_known_url_response = connection.get("/.well-known/openeo", timeout=timeout)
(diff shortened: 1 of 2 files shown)


soxofaan:
Hi, thanks for contributing!

I'm however not sure  what the actual problem is.
Connecting to a url with path component is an often used feature and has always been functional as far as I know.

e.g.
```pycon
>>> import openeo
>>> connection = openeo.connect("https://openeo.dataspace.copernicus.eu/openeo")
>>> connection.version_info()
{'client': '0.47.0a3', 'api': '1.2.0', 'backend': {'root_url': 'https://openeo.datas...
```


About `url_join`: there is test coverage for all kind of permuations:
https://github.com/Open-EO/openeo-python-client/blob/a4dfa5c1418ed2fc0cf928fc8ef5116bc903b8ae/tests/test_util.py#L802-L811

so I'm not sure what problem you are actually having
 
Can you share a concrete backend URL that causes trouble?

pantierra:
Thanks for the fast reply. The error was on the server side, not the client. Sorry for the noise and thanks for all the good work. Closing here.

PR #859, opened 2026-02-06
Support new JWT token based auth (openEO API 1.3.0)

### LINKED ISSUE #850: Support new JWT token based auth from openEO API 1.3
openEO API 1.3 deprecates custom openEO bearer tokens and recommends vanilla JWT. See:
- https://github.com/Open-EO/openeo-api/issues/558
- https://github.com/Open-EO/openeo-api/pull/567

No back-end switched yet (and probably won't soon), but client has to be prepared to be ready

niebl:
#850

Includes: 
- checking conformance in `connection.authenticate_basic()` and `connection.authenticate_oidc()` and formatting the bearer tokens accordingly 
- a `/conformance` endpoint in `DummyBackend` and a `has_conformance({url})` method in capabilities. So JWT bearer token conformance can be checked.


openeo/rest/auth/auth.py
```diff
@@ -41,12 +41,21 @@ def __call__(self, req: Request) -> Request:
...
     """Bearer token for OIDC Auth (openEO API 1.0.0 style)"""
 
-    def __init__(self, provider_id: str, access_token: str):
-        super().__init__(bearer="oidc/{p}/{t}".format(p=provider_id, t=access_token))
+    def __init__(self, provider_id: str, access_token: str, jwt_conformance: bool = False):
+        if jwt_conformance:
+            bearer="{t}"
+        else:
+            bearer="oidc/{p}/{t}".format(p=provider_id, t=access_token)
+        super().__init__(bearer=bearer)
```
m-mohr:
```suggestion
    def __init__(self, provider_id: str, access_token: str, jwt_conformance: bool = False):
        if not jwt_conformance:
            access_token="oidc/{p}/{t}".format(p=provider_id, t=access_token)
        super().__init__(bearer=access_token)
```

openeo/rest/auth/auth.py
```diff
@@ -41,12 +41,21 @@ def __call__(self, req: Request) -> Request:
...
 
-    def __init__(self, access_token: str):
-        super().__init__(bearer="basic//{t}".format(t=access_token))
+    def __init__(self, access_token: str, jwt_conformance: bool = False):
+        bearer = False
+        if jwt_conformance:
+            bearer= "{t}".format(t=access_token)
+        else:
+            bearer = "basic//{t}".format(t=access_token)
+        super().__init__(bearer=bearer)
```
m-mohr:
```suggestion
    def __init__(self, access_token: str, jwt_conformance: bool = False):
        if not jwt_conformance:
            access_token = "basic//{t}".format(t=access_token)
        super().__init__(bearer=access_token)
```

openeo/rest/_testing.py
```diff
@@ -424,6 +433,20 @@ def get_status(job_id: str, current_status: str) -> str:
...
+def build_conformance(
+    *,
+    api_version: str = "1.0.0",
+    stac_version: str = "0.9.0",
+) -> list[str]:
+    conformance = [
+        "https://api.openeo.org/{api_version}",
+        "https://api.stacspec.org/v{stac_version}/core",
+        "https://api.stacspec.org/v{stac_version}/collections"
+        ]
```
m-mohr:
```suggestion
    ]
```

Make sure to run the pre-commit / lint / format (not sure what's used here). I think some of the indentations are slightly off.

openeo/rest/_testing.py
```diff
@@ -470,17 +495,23 @@ def build_capabilities(
...
                 {"path": "/process_graphs", "methods": ["GET"]},
-                {"path": "/process_graphs/{process_graph_id", "methods": ["GET", "PUT", "DELETE"]},
+                {"path": "/process_graphs/{process_graph_id}", "methods": ["GET", "PUT", "DELETE"]},
             ]
         )
 
+    conformance = build_conformance(
+        api_version=api_version, 
+        stac_version=stac_version
+        )
```
m-mohr:
```suggestion
    )
```

openeo/rest/capabilities.py
```diff
@@ -36,6 +37,15 @@ def api_version_check(self) -> ComparableVersion:
         if not api_version:
             raise ApiVersionException("No API version found")
         return ComparableVersion(api_version)
+    
+    def has_conformance(self, conformance: str) -> bool:
+        """Check if backend provides a given conformance string"""
+        if "conformsTo" in self.capabilities:
+            for url in self.capabilities["conformsTo"]:
+                if fnmatch(url, conformance):
```
m-mohr: I think I'd go with a regexp as in the JS client here to avoid potential conflicts with other special characters in fnmatch (e.g. ?)

openeo/rest/capabilities.py
```diff
@@ -36,6 +37,15 @@ def api_version_check(self) -> ComparableVersion:
         if not api_version:
             raise ApiVersionException("No API version found")
         return ComparableVersion(api_version)
+    
+    def has_conformance(self, conformance: str) -> bool:
+        """Check if backend provides a given conformance string"""
+        if "conformsTo" in self.capabilities:
+            for url in self.capabilities["conformsTo"]:
```
m-mohr: You can skip the `if "conformsTo" in self.capabilities` by just using `for url in self.capabilities.get("conformsTo", []):`

openeo/rest/connection.py
```diff
@@ -277,8 +277,12 @@ def authenticate_basic(self, username: Optional[str] = None, password: Optional[
             # /credentials/basic is the only endpoint that expects a Basic HTTP auth
             auth=HTTPBasicAuth(username, password)
         ).json()
+
+        # check for JWT bearer token conformance
+        jwt_conformance = self.capabilities().has_conformance("https://api.openeo.org/*/authentication/jwt")
```
m-mohr: Better put the "https://api.openeo.org/*/authentication/jwt" into a const either here or that you import (maybe there's a place in the Python client where such consts are defined), then we don't need to have it multiple times in the code.

openeo/rest/connection.py
```diff
@@ -416,7 +420,9 @@ def _authenticate_oidc(
             )
 
         token = tokens.access_token
-        self.auth = OidcBearerAuth(provider_id=provider_id, access_token=token)
+        # check for JWT bearer token conformance
+        jwt_conformance = self.capabilities().has_conformance("https://api.openeo.org/*/authentication/jwt")
```
m-mohr: See https://github.com/Open-EO/openeo-python-client/pull/859/changes#r2784662304

openeo/rest/connection.py
niebl: fixed in [71a4503](https://github.com/Open-EO/openeo-python-client/pull/859/commits/71a4503c7f418475fa507d2630cf98b43697a7ba)

openeo/rest/connection.py
niebl: fixed in [71a4503](https://github.com/Open-EO/openeo-python-client/pull/859/commits/71a4503c7f418475fa507d2630cf98b43697a7ba)

tests/rest/test_connection.py
```diff
@@ -59,7 +59,9 @@
 API_URL = "https://oeo.test/"
 
 # TODO: eliminate this and replace with `build_capabilities` usage
-BASIC_ENDPOINTS = [{"path": "/credentials/basic", "methods": ["GET"]}]
+BASIC_ENDPOINTS = [
+    {"path": "/credentials/basic", "methods": ["GET"]}
+    ]
```
m-mohr:
```suggestion
]
```

openeo/rest/auth/auth.py
niebl: [547cdab](https://github.com/Open-EO/openeo-python-client/pull/859/commits/547cdabe7db83bd0d51d08bc5c2c50aeabc16e84)

openeo/rest/auth/auth.py
niebl: [547cdab](https://github.com/Open-EO/openeo-python-client/pull/859/commits/547cdabe7db83bd0d51d08bc5c2c50aeabc16e84)

openeo/rest/_testing.py
```diff
@@ -441,6 +464,8 @@ def build_capabilities(
     """Build a dummy capabilities document for testing purposes."""
 
     endpoints = []
+    if basic_auth:
```
m-mohr: Shouldn't it always be added?

openeo/rest/_testing.py
```diff
@@ -441,6 +464,8 @@ def build_capabilities(
...
+        endpoints.append({"path": "/conformance", "methods": ["GET"]})
```
m-mohr:
```suggestion
```

tests/rest/test_connection.py
```diff
@@ -462,8 +464,8 @@ def test_connect_with_session():
             ],
             "https://oeo.test/openeo/1.1.0/",
             "1.1.0",
-        ),
-        (
+    ),
+    (
```
m-mohr: Why?

tests/rest/test_connection.py
```diff
@@ -885,7 +898,6 @@ def test_authenticate_oidc_authorization_code_100_single_implicit(requests_mock,
     assert conn.auth.bearer == 'oidc/fauth/' + oidc_mock.state["access_token"]
     assert "No OIDC provider given, but only one available: 'fauth'. Using that one." in caplog.text
 
-
```
m-mohr, line 888: ?

tests/rest/test_connection.py
```diff
@@ -848,7 +850,6 @@ def test_authenticate_basic(requests_mock, api_version, basic_auth):
     assert isinstance(conn.auth, BearerAuth)
     assert conn.auth.bearer == "basic//6cc3570k3n"
 
-
```
m-mohr, line 851: ?

m-mohr (approved):

openeo/rest/auth/auth.py
```diff
@@ -41,12 +41,17 @@ def __call__(self, req: Request) -> Request:
 class BasicBearerAuth(BearerAuth):
     """Bearer token for Basic Auth (openEO API 1.0.0 style)"""
 
-    def __init__(self, access_token: str):
-        super().__init__(bearer="basic//{t}".format(t=access_token))
+    def __init__(self, access_token: str, jwt_conformance: bool = False):
```
soxofaan: Isn't it easier to just use `BearerAuth` directly when "JWT conformance" mode is desired  instead of making `BasicBearerAuth` and `OidcBearerAuth`  more complex?

The point of JWT mode is to eliminate  the special bearer handling (with dedicated basic and OIDC variants), which the implementation should reflect I think

openeo/rest/_testing.py
```diff
@@ -424,6 +433,20 @@ def get_status(job_id: str, current_status: str) -> str:
 
         self.job_status_updater = get_status
 
+def build_conformance(
+    *,
+    api_version: str = "1.0.0",
+    stac_version: str = "0.9.0",
```
soxofaan: I don't think we should introduce something new defaulting to STAC 0.9.0. I would at least take STAC 1.0.0 or even 1.1.0 here

tests/rest/test_testing.py
```diff
@@ -7,54 +7,57 @@
 
 
 @pytest.fixture
-def dummy_backend(requests_mock, con120):
+def dummy_backend120(requests_mock, con120):
```
soxofaan: I would not rename  `dummy_backend` here, it ruins the signal/noise-ratio of this PR

What openEO-version or actual auth-mechanism is in place is irrelevant for most tests, so I would preserve a version-agnostic of this fixture

so I would leave the existing `dummy_backend` fixture as is  and just _add_   `dummy_backend120` and `dummy_backend130`

tests/rest/test_connection.py
```diff
@@ -1049,6 +1064,36 @@ def test_authenticate_oidc_auth_code_pkce_flow_client_from_config(requests_mock,
...
+    conn = Connection(API_URL, refresh_token_store=refresh_token_store)
+    assert isinstance(conn.auth, NullAuth)
+    conn.authenticate_oidc_authorization_code(webbrowser_open=oidc_mock.webbrowser_open)
+    capabilities = conn.capabilities()
+    assert isinstance(conn.auth, BearerAuth)
+    assert capabilities.api_version() == "1.3.0"
+    assert capabilities.has_conformance("https://api.openeo.org/*/authentication/jwt") == True
+    assert conn.auth.bearer == oidc_mock.state["access_token"]
+    # TODO: check issuer ("iss") value in parsed jwt. this will require the example jwt to be formatted accordingly 
+    assert refresh_token_store.mock_calls == []
```
soxofaan: Instead of duplicating tests for JWT conformace mode (like this single `test_authenticate_oidc_auth_code_pkce_flow`), I think we should look instead into parameterizing all existing tests here.

e.g. there are 41 occurrences of `assert.*bearer.*oidc/`  in this file, so there is a lot of test coverage that should be ported to JWT conformance mode (as it is meant to become the default/recommended approach)

soxofaan (commented):
great, thanks for this effort

some notes:

openeo/rest/__init__.py
```diff
@@ -10,6 +10,8 @@
 DEFAULT_JOB_STATUS_POLL_CONNECTION_RETRY_INTERVAL = 30
 DEFAULT_JOB_STATUS_POLL_SOFT_ERROR_MAX = 10
 
+CONFORMANCE_JWT_BEARER = "https://api.openeo.org/*/authentication/jwt"
```
soxofaan: I think it makes more sense to put this constant in rest/capabilities.py

openeo/rest/_testing.py
```diff
@@ -470,17 +493,23 @@ def build_capabilities(
...
             [
                 {"path": "/process_graphs", "methods": ["GET"]},
-                {"path": "/process_graphs/{process_graph_id", "methods": ["GET", "PUT", "DELETE"]},
+                {"path": "/process_graphs/{process_graph_id}", "methods": ["GET", "PUT", "DELETE"]},
             ]
         )
 
+    conformance = build_conformance(
+        api_version=api_version, 
+        stac_version=stac_version
```
soxofaan:
```suggestion
        stac_version=stac_version,
```

otherwise the auto-formatting tooling we use will collapse this to a single line statement

openeo/rest/capabilities.py
```diff
@@ -36,6 +37,15 @@ def api_version_check(self) -> ComparableVersion:
...
             raise ApiVersionException("No API version found")
         return ComparableVersion(api_version)
+    
+    def has_conformance(self, uri: str) -> bool:
+        """Check if backend provides a given conformance string"""
+        uri = re.escape(uri).replace('\\*', '[^/]+')
+        for conformance_uri in self.capabilities.get("conformsTo", []):
+            if re.match(uri, conformance_uri):
+                return True
+        return False
```
soxofaan: suggestion to make this a bit more pythonic:

```suggestion
        regex = re.compile(re.escape(uri).replace('\\*', '[^/]+'))
        return any(regex.fullmatch(c) for c in self.capabilities.get("conformsTo", []))
```

soxofaan: regardless, I'm not a such a fan of introducing an ad-hoc pattern matching approach in this new method `has_conformance` (where `*` is converted to `[^/\]+`). 
It's not documented and does not follow more common and standardized approaches (e.g. exact string comparison and python stdlib regexes).

I'd rather have:
- `has_conformance()` to accept simple strings (for exact comparison) and standard regexes (to be matched with `fullmatch`)
- make `CONFORMANCE_JWT_BEARER` a standard (compiled) regex pattern, so you can pass it directly to `has_conformance()`

tests/rest/test_connection.py
```diff
@@ -59,7 +59,9 @@
 API_URL = "https://oeo.test/"
 
 # TODO: eliminate this and replace with `build_capabilities` usage
-BASIC_ENDPOINTS = [{"path": "/credentials/basic", "methods": ["GET"]}]
+BASIC_ENDPOINTS = [
+    {"path": "/credentials/basic", "methods": ["GET"]}
+]
```
soxofaan: is this change intended?

note that auto-formatting tooling we use (black, darker) will probably reset this back to a one-liner

soxofaan (commented):
some more notes

openeo/rest/capabilities.py
m-mohr: I guess then we should go for option 2, otherwise we end up with lists such as

````
[
  "https://api.openeo.org/v1.2/example",
  "https://api.openeo.org/v1.3/example",
  "https://api.openeo.org/v1.4/example",
  "https://api.openeo.org/v1.5-beta.1/example",
  "https://api.openeo.org/v1.5-beta.2/example",
...
]
````

File: openeo/rest/_testing.py (hunk 16 above)
Comment (niebl) [reply openeo/rest/_testing.py]: bumped the version up to 1.1.0 in [8b72876](https://github.com/Open-EO/openeo-python-client/pull/859/commits/8b72876d5138cb6230e470da2cf5b9d1bb18825c)

File: openeo/rest/auth/auth.py (hunk 15 above)
Comment (niebl) [reply openeo/rest/auth/auth.py]: in [85ff1b0](https://github.com/Open-EO/openeo-python-client/pull/859/commits/85ff1b0cae4fbb9c8fcc8dde356a141536ad8eac) i reverted `auth.py` to the state it was in before the changes and let the authenticate-methods in `connection.py` check for conformance and select the `BearerAuth` class accordingly

File: openeo/rest/__init__.py (hunk 19 above)
Comment (niebl) [reply openeo/rest/__init__.py]: implemented this suggestion in [240c18d](https://github.com/Open-EO/openeo-python-client/pull/859/commits/240c18d565351aeaf387c3e3bc53779ee67491e1)

File: tests/rest/test_testing.py (hunk 17 above)
Comment (niebl) [reply tests/rest/test_testing.py]: revisions done in [c2c5766](https://github.com/Open-EO/openeo-python-client/pull/859/commits/c2c5766822bce0f30f23c615c605b2512fcbdd2f)

while dummy_backend120 is theoretically redundant, i decided to initiate it as a reference to dummy_backend to improve readability of the new tests

File: tests/rest/test_connection.py (hunk 18 above)
Comment (niebl) [reply tests/rest/test_connection.py]: hi @soxofaan,
that's a lot of test coverage.
[1d5c20f](https://github.com/Open-EO/openeo-python-client/pull/859/commits/1d5c20fc7e2beb3ee47aebdaae13acd679b3d9a7) now uses both `1.0.0` and `1.3.0` versions as parameters. 

There are still five failing tests under version `1.3.0`, namely the tests from `test_authenticate_oidc_auto_renew_expired_access_token_invalid_refresh_token` to `test_authenticate_oidc_auto_renew_expired_access_token_initial_device_code`

All of these have in common that they use the `get_me_handler` to extract access token and oidc_provider from the bearer token. This could be changed to extract the oicd provider from the `iss` issuer-attribute in the jwt, but afaik that is not yet included in the example bearer tokens and would need to be implemented first.

Should we change those parts of the code accordingly so the `get_me_handler` can also pick the oidc_provider from he token?

File: openeo/rest/capabilities.py (hunk 21 above)
Comment (niebl) [reply openeo/rest/capabilities.py]: [8f960e8](https://github.com/Open-EO/openeo-python-client/pull/859/commits/8f960e8b0f374c5dfada689c0105c2c6af5850ff) now uses compiled regex patterns to check for conformances.

**Comment by soxofaan** [comment]
hmm, it seems this PR got in a bad state, making the signal/noise ratio bad for review. I see various changes in places that are unrelated to the feature to implement here (conversions.py, results.py, _version.py, metadata.py, test_s3sts.py, ...)

did you by any chance cherry-picked commits from master, instead of merging master?

File: openeo/rest/auth/auth.py [hunk 23]
```diff
@@ -29,8 +29,9 @@ class BearerAuth(OpenEoApiAuthBase):
     https://open-eo.github.io/openeo-api/apireference/#section/Authentication/Bearer
     """
 
-    def __init__(self, bearer: str):
+    def __init__(self, bearer: str, origin: str):
```
soxofaan: I'm still unsure about this `origin` stuff

at least I would make it optional (e.g. default `None`)

but looking more into this: `.origin` is only used here:
```python
if isinstance(self.auth, BearerAuth) and self.auth.origin == "oidc" and self._oidc_auth_renewer:
```
as update to the original `isinstance(self.auth, OidcBearerAuth)` check

But I think that original `isinstance(self.auth, OidcBearerAuth)` check actually became unnecessary over time (the presence of `_oidc_auth_renewer`  is enough to check).

So I would just remove the `isinstance(self.auth, OidcBearerAuth)` and then you don't need the `origin` in the bearer classes

openeo/rest/auth/testing.py
```diff
@@ -281,7 +281,7 @@ def validate_access_token(self, access_token: str):
         raise LookupError("Invalid access token")
 
     def invalidate_access_token(self):
-        self.state["access_token"] = "***invalidated***"
+        self.state["access_token"] = None
```
soxofaan: is this change necessary to unbreak something?

openeo/rest/_testing.py
```diff
@@ -424,6 +433,20 @@ def get_status(job_id: str, current_status: str) -> str:
...
+    api_version: str = "1.0.0",
+    stac_version: str = "1.1.0",
+) -> list[str]:
+    conformance = [
+        "https://api.openeo.org/{api_version}",
+        "https://api.stacspec.org/v{stac_version}/core",
+        "https://api.stacspec.org/v{stac_version}/collections"
+    ]
+    if ComparableVersion(api_version) >= ComparableVersion("1.3.0"):
+        conformance.append(f"https://api.openeo.org/{api_version}/authentication/jwt")
```
soxofaan, line 440: I think it should also be possible to test use cases of an v1.3 api *without* jwt conformance class

(the current implementation assumes hard coupling of 1.3 and jwt)

openeo/rest/connection.py
```diff
@@ -736,7 +748,7 @@ def try_access_token_refresh(self, *, reason: Optional[str] = None) -> bool:
         Returns whether a new access token was obtained.
         """
         reason = f" Reason: {reason}" if reason else ""
-        if isinstance(self.auth, OidcBearerAuth) and self._oidc_auth_renewer:
+        if isinstance(self.auth, BearerAuth) and self.auth.origin == "oidc" and self._oidc_auth_renewer:
```
soxofaan: as mentioned: I think this can be simplified to just:
```suggestion
        if self._oidc_auth_renewer:
```

the original `isinstance` check got redundant here

tests/rest/conftest.py
```diff
@@ -19,6 +19,10 @@ def api_version(request):
     return request.param
 
 
+@pytest.fixture(params=["1.0.0", "1.3.0"])
+def api_version_authentication_tests(request):
```
soxofaan: I see you originally just updated the parameterization of the existing `api_version` fixture, which is more elegant I think than this more verbose and ad-hoc-looking fixture `api_version_authentication_tests`

What's the reason for introducing it?


Note: if you did this to avoid explosion of test duplication in test modules where v1.3 is not relevant (yet): pytest allows you to  override a fixture at module level and leave the one in the global conftest alone.
For example: you can define `api_version` with the additional v1.3 param just  in test_connection.py where you want to cover more API versions. All other test modules are not impacted

tests/rest/test_connection.py
```diff
@@ -839,30 +840,41 @@ def test_authenticate_basic_no_support(requests_mock, api_version):
...
     assert isinstance(conn.auth, NullAuth)
     conn.authenticate_basic(username=basic_auth.username, password=basic_auth.password)
+    capabilities = conn.capabilities()
     assert isinstance(conn.auth, BearerAuth)
-    assert conn.auth.bearer == "basic//6cc3570k3n"
+    if ComparableVersion(api_version_authentication_tests) >= ComparableVersion("1.3.0"):
+        assert capabilities.has_conformance(CONFORMANCE_JWT_BEARER) == True
+        assert conn.auth.bearer == "6cc3570k3n"
+    else:
+        assert conn.auth.bearer == "basic//6cc3570k3n"
```
soxofaan, line 875: as mentioned above: there is assumption here of hard coupling between API v1.3 and JWT support

I think that should be decoupled, because that's also how this will be in the wild

also: I would consider making a small helper for this `if-else` because I see it is duplicated  in a lot of tests below

tests/rest/test_connection.py
```diff
@@ -927,7 +945,11 @@ def test_authenticate_oidc_authorization_code_100_multiple_no_given_id(requests_
     assert isinstance(conn.auth, NullAuth)
     conn.authenticate_oidc_authorization_code(client_id=client_id, webbrowser_open=oidc_mock.webbrowser_open)
     assert isinstance(conn.auth, BearerAuth)
-    assert conn.auth.bearer == 'oidc/fauth/' + oidc_mock.state["access_token"]
+    if ComparableVersion(api_version_authentication_tests) >= ComparableVersion("1.3.0"):
+        # TODO: migth require future tests for the issuer encoded in the jwt
```
soxofaan: good point.
however, the client has zero control over the JWT in practice, so this is not a testable concern I'm afraid.

It could be a valuable test topic in  integration tests

tests/rest/conftest.py
soxofaan: (e.g. the `api_version` -> `api_version_authentication_tests` change makes this PR larger and harder to review than it could be I think)

soxofaan (commented):
sorry for delay, I finally found some time to give this another review

tests/rest/conftest.py
niebl: that's right I was attempting to prevent unneeded test duplication. 
[63d607a](https://github.com/Open-EO/openeo-python-client/pull/859/commits/63d607a17be5b0fa4e848d360c70afe5d52d4dca) now uses an override of api_version for `test_connection.py`

niebl:
Thank you very much for the review. I am sorry for the delay. I will be able to work on that feedback now

tests/rest/conftest.py
niebl: Following your suggestion to uncouple the version number and the jwt-conformance I have parametrized the test cases accordingly in [cba109f](https://github.com/Open-EO/openeo-python-client/pull/859/commits/cba109fd2ea3969d3944078e891dc0590aae2835).

But in doing so I found that the total size of the PR does not decrease in a significant way because each test still requires the Bearer-token format to be passed in order for the capabilities to be generated with the according conformance classes.
Maybe I am still overlooking some way of parametrizing this more efficiently.

tests/rest/test_testing.py
```diff
@@ -104,3 +109,17 @@ def test_setup_job_start_failure(self, dummy_backend):
...
+
+        assert capabilities120.api_version() == "1.2.0"
+        assert capabilities130.api_version() == "1.3.0"
+
+    def test_jwt_conformance(self, dummy_backend, dummy_backend130):
+        capabilities120 = dummy_backend.connection.capabilities()
+        capabilities130 = dummy_backend130.connection.capabilities()
+
+        assert capabilities120.has_conformance(CONFORMANCE_JWT_BEARER) == False
+        assert capabilities130.has_conformance(CONFORMANCE_JWT_BEARER) == True
```
niebl: `DummyBackend` does not seem to get used for testing with jwt-conformance in any of the `test_connection` cases. That may make testing for its jwt conformance redundant.

openeo/rest/_testing.py
niebl: thank you! As of [de28943](https://github.com/Open-EO/openeo-python-client/pull/859/commits/de28943f1ddb7d6e82473979f67168936a0cd5ca) the version and conformance has been decoupled

openeo/rest/auth/testing.py
niebl: as far as I remember, no functionality was broken.

openeo/rest/auth/auth.py
niebl: Judging by the code it does look like `_oidc_auth_renewer` is enough.
As origin is not checked elsewhere I am removing those changes for now until it may become necessary again to check the origin of a bearer token

tests/rest/test_testing.py
niebl: [cc4a2e5](https://github.com/Open-EO/openeo-python-client/pull/859/commits/cc4a2e5ebbff2fc3d03f1b38885dfe4b66ddd303) removed the test cases for now.

niebl:
@soxofaan thank you for the review. I was able to apply your suggestions and shrink the size of the PR a little. I am once again sorry for the delay

PR #867, opened 2026-03-03
Respect STAC cube:dimensions as source of truth in metadata_from_stac + align tests with dimension policy

suriyahgit:
## Problem

`metadata_from_stac()` previously injected openEO default dimensions (`x`, `y`, `t`, `bands`) even when a STAC object already defined dimensions via `cube:dimensions`.

This led to:

* dimension names being overridden or altered,
* inconsistencies between STAC metadata and openEO metadata,
* tests relying on implicit default behavior instead of STAC semantics.

## Fix

Introduce a clear dimension resolution policy:

* If **`cube:dimensions` is present**
  → treat it as the **source of truth** and preserve dimension names and order.

* If **`cube:dimensions` is absent**
  → fall back to openEO defaults (`x`, `y`, optional `t`).

The metadata parsing logic was refactored accordingly and aligned with PySTAC datacube capabilities.

## Testing

* Updated existing tests to no longer assume an always-present band dimension.
* Added checks verifying:

  * fallback behavior when `cube:dimensions` is missing,
  * preservation of dimension names when `cube:dimensions` exists,
  * correct temporal dimension handling across STAC object types.
* Tests now explicitly enforce the new dimension policy to prevent regressions.

This change makes metadata parsing more STAC-compliant while keeping backward-compatible fallback behavior.

Refer:
https://github.com/destine-datalake-cube/dedl-openeo-coordination/issues/3
Also solves: https://github.com/Open-EO/openeo-python-client/issues/743


openeo/metadata.py
```diff
@@ -678,31 +675,168 @@ def __str__(self) -> str:
 def metadata_from_stac(url: str) -> CubeMetadata:
     """
     Reads the band metadata a static STAC catalog or a STAC API Collection and returns it as a :py:class:`CubeMetadata`
-
     :param url: The URL to a static STAC catalog (STAC Item, STAC Collection, or STAC Catalog) or a specific STAC API Collection
+
     :return: A :py:class:`CubeMetadata` containing the DataCube band metadata from the url.
+    Philosophy:
+      - If cube:dimensions exists: treat it as source of truth (it may omit x/y/t/bands).
+      - Otherwise: apply openEO-style defaults (x, y, t) and (for Collection/Item) keep bands dimension even if empty.
```
soxofaan, line 687: Move this "Philosophy" above `:param:` and `:return:` otherwise this will break the docs I think

openeo/metadata.py
```diff
@@ -678,31 +675,168 @@ def __str__(self) -> str:
...
+    bands = parser.bands_from_stac_object(stac_object)
 
-    # At least assume there are spatial dimensions
-    # TODO #743: are there conditions in which we even should not assume the presence of spatial dimensions?
-    dimensions = [
-        SpatialDimension(name="x", extent=[None, None]),
-        SpatialDimension(name="y", extent=[None, None]),
-    ]
+    # aligning better with STAC datacube extension
+    def _cube_dimensions_dict(obj: pystac.STACObject) -> dict:
```
soxofaan:
```suggestion
    def _cube_dimensions_dict(obj: pystac.STACObject) -> Dict:
```

openeo python library still supports python 3.8, so we can't use these lower case  type annotations yet

openeo/metadata.py
```diff
@@ -678,31 +675,168 @@ def __str__(self) -> str:
...
+            return obj.extra_fields.get("cube:dimensions", {}) or {}
+        return {}
 
-    # TODO #743: conditionally include band dimension when there was actual indication of band metadata?
-    band_dimension = BandDimension(name="bands", bands=bands)
-    dimensions.append(band_dimension)
+    # Standardize fallback temporal dimension name to "t"
+    # (openEO convention). We don't try to guess backend-specific naming here; we only infer temporal extent
+    # From Collection extent or Item datetime/start/end when available.
+    def _infer_temporal_extent(obj: pystac.STACObject) -> list:
```
soxofaan: To be honest, I'm not a big fan of using nested functions extensively as used in this PR. Nested functions are impossible to reach/test in isolation, and variable scoping can get confusing or error-prone.

I'd prefer to build further on the existing `_StacMetadataParser` class instead.
E.g. this PR seems to remove `get_temporal_dimension` usage, instead of improving it

tests/test_metadata.py
```diff
@@ -1253,7 +1256,93 @@ def test_metadata_from_stac_temporal_dimension(tmp_path, stac_dict, expected):
         assert isinstance(dim, TemporalDimension)
         assert (dim.name, dim.extent) == expected
     else:
-        assert not metadata.has_temporal_dimension()
+        # With openEO defaults, a temporal dimension name ('t') can exist even when extent is unknown.
```
soxofaan: why all this code and not just updating the expected value here?

tests/test_metadata.py
```diff
@@ -1253,7 +1256,93 @@ def test_metadata_from_stac_temporal_dimension(tmp_path, stac_dict, expected):
...
+    metadata = metadata_from_stac(str(path))
+
+    got = tuple(metadata.dimension_names() or ())
+
+    # Order-insensitive check: names only
+    assert set(got) == expected_dims, (
+        f"Unexpected dimension names for STAC dict\n"
+        f"Expected set: {expected_dims}\n"
+        f"Got:          {got} (set={set(got)})"
+    )
```
soxofaan:
```suggestion
    assert set(got) == expected_dims
```

no need for custom assert messages.
pytest automatically takes care of this

tests/test_metadata.py
```diff
@@ -1253,7 +1256,93 @@ def test_metadata_from_stac_temporal_dimension(tmp_path, stac_dict, expected):
...
+            # Allow "unknown" temporal extent representation
+            if extent == [None, None]:
+                pass
+            else:
+                # Allow degenerate interval when a single datetime is available
+                assert extent[0] == extent[1]
+
+
+
+# Dimension name resolution policy (STAC cube:dimensions vs openEO defaults)
```
soxofaan, line 1292: which policy are you referring to here?

tests/test_metadata.py
```diff
@@ -1253,7 +1256,93 @@ def test_metadata_from_stac_temporal_dimension(tmp_path, stac_dict, expected):
...
+    )
+
+    # Ensure the policy logic is exercised correctly:
+    # cube:dimensions can be located at root (collection) or in properties (item)
+    cube_dims = stac_dict.get("cube:dimensions") or (stac_dict.get("properties") or {}).get("cube:dimensions")
+    if cube_dims is None:
+        assert set(got) == _DEFAULT_OPENEO_DIMS_SET, (
+            f"STAC dict has no cube:dimensions but did not fall back to openEO defaults.\n"
+            f"Got: {got}, expected defaults: {_DEFAULT_OPENEO_DIMS_SET}"
+        )
```
soxofaan:
```suggestion
        assert set(got) == _DEFAULT_OPENEO_DIMS_SET
```

no need for custom message, as pytest takes care of that automatically

this also eliminates the need to define `_DEFAULT_OPENEO_DIMS_SET` as a test constant I think

tests/test_metadata.py
```diff
@@ -1253,7 +1256,93 @@ def test_metadata_from_stac_temporal_dimension(tmp_path, stac_dict, expected):
...
+    else:
+        assert set(got) == set(cube_dims.keys()), (
+            f"STAC dict has cube:dimensions but resolved dims do not match it.\n"
+            f"cube:dimensions keys: {tuple(cube_dims.keys())}\n"
+            f"resolved dims:        {got}"
+        )
```
soxofaan:
```suggestion
        assert set(got) == set(cube_dims.keys())
```

tests/test_metadata.py
```diff
@@ -1253,7 +1256,93 @@ def test_metadata_from_stac_temporal_dimension(tmp_path, stac_dict, expected):
...
+
+    if must_have_cube_dims:
+        assert cube_dims is not None, "Test case expected cube:dimensions but it was missing."
```
soxofaan: what is this assert for? It just seems to test the test parameters but no real logic.

soxofaan (commented):
just could give this a quick scan to make some notes

tests/test_metadata.py
suriyahgit: previously i had a set of tests, so i assigned it to a variable, now that it is already lean, I replaced abstraction to replacing it with expected value

tests/test_metadata.py
suriyahgit: The mention of 'policy' here was mainly used for internal review. By dimension name resolution policy I simply mean that the dimension names defined in cube:dimensions take absolute priority when present; otherwise, the openEO default canonical dimensions are used. I'll better remove it here from the tests not to complicate things there.

tests/test_metadata.py
suriyahgit: Good catch, the boolean flag was redundant because the test input already indicates whether cube:dimensions are present, so I simplified the parametrization accordingly.

openeo/metadata.py
suriyahgit: Understood the design decision, i am working in the re-structuring of the parser functions into the `_StacMetadaParser` class.

openeo/metadata.py
suriyahgit: pushed this above Policy/Philosophy

clausmichele:
Hi @soxofaan , please let us know if there is something we can do to move on with this PR!

tests/test_metadata.py
```diff
@@ -1253,7 +1256,73 @@ def test_metadata_from_stac_temporal_dimension(tmp_path, stac_dict, expected):
...
+                pass
+            else:
+                # Allow degenerate interval when a single datetime is available
+                assert extent[0] == extent[1]
+
+
+
+# Dimension name resolution policy (STAC cube:dimensions vs openEO defaults)
+@pytest.mark.parametrize(
+    ["stac_dict", "expected_dims"],
```
jdries, line 1280: can a STAC 1.1 style test case be added which uses the new 'bands' common metadata and no datacube extension?
Collection metadata example:
https://github.com/radiantearth/stac-spec/blob/master/examples/collection-only/collection.json#L152
Bands spec:
https://github.com/radiantearth/stac-spec/blob/master/commons/common-metadata.md#band-object

jdries (changes requested):
FYI, Stefaan is on holiday
Added one comment myself.
Please also add a changelog entry.
Otherwise I'm inclined to say it looks good, assuming this change mostly preserves existing behaviour. (Biggest worry are collections with faulty datacube extension metadata that would suddenly have a change in behavior.)

openeo/metadata.py
```diff
@@ -760,39 +743,172 @@ def __init__(self, *, logger=_log, log_level=logging.DEBUG, supress_duplicate_wa
...
+                dimensions.append(BandDimension(name="bands", bands=list(bands)))
+            return dimensions
+
+        dimensions: List[Dimension] = [
+            SpatialDimension(name="x", extent=[None, None]),
+            SpatialDimension(name="y", extent=[None, None]),
+            TemporalDimension(name="t", extent=self.infer_temporal_extent(stac_object)),
+        ]
+        if isinstance(stac_object, (pystac.Collection, pystac.Item)):
+            dimensions.append(BandDimension(name="bands", bands=list(bands)))
```
soxofaan: why conditionally add the band dimension here?

openeo/metadata.py
```diff
@@ -760,39 +743,172 @@ def __init__(self, *, logger=_log, log_level=logging.DEBUG, supress_duplicate_wa
...
+                return temporal_dimensions[0]
+
+        if isinstance(stac_obj, (pystac.Collection, pystac.Item)):
+            return TemporalDimension(name="t", extent=self.infer_temporal_extent(stac_obj))
+
+    def has_cube_dimensions(self, stac_object: pystac.STACObject) -> bool:
+        cube_dimensions = self.cube_dimensions_dict(stac_object)
+        return isinstance(cube_dimensions, dict) and len(cube_dimensions) > 0
+
+    def cube_dimensions_dict(self, stac_object: pystac.STACObject) -> Dict[str, dict]:
```
soxofaan: I don't think this has to be in the public interface

openeo/metadata.py
```diff
@@ -760,39 +743,172 @@ def __init__(self, *, logger=_log, log_level=logging.DEBUG, supress_duplicate_wa
...
+        Best-effort temporal extent:
+          - Collection: extent.temporal interval
+          - Item: datetime or start/end
+        """
+        if isinstance(stac_object, pystac.Collection) and stac_object.extent and stac_object.extent.temporal:
+            interval = stac_object.extent.temporal.intervals[0]
+            return [Rfc3339(propagate_none=True).normalize(d) for d in interval]
+
+        if isinstance(stac_object, pystac.Item):
+            props = getattr(stac_object, "properties", {}) or {}
```
soxofaan: just wondering: why using `getattr` here? Isn't the presence of `properties` guaranteed?

openeo/metadata.py
```diff
@@ -760,39 +743,172 @@ def __init__(self, *, logger=_log, log_level=logging.DEBUG, supress_duplicate_wa
...
+            dt_ = props.get("datetime")
+            if dt_:
+                norm = Rfc3339(propagate_none=True).normalize(dt_)
+                return [norm, norm]
+            start = props.get("start_datetime")
+            end = props.get("end_datetime")
```
soxofaan: isn't the "datetime" field a required field, so is this code even reachable?
Or to put differently: shouldn't "start_datetime"/"end_datetime" get higher precedence over "datetime"?

openeo/metadata.py
```diff
@@ -760,39 +743,172 @@ def __init__(self, *, logger=_log, log_level=logging.DEBUG, supress_duplicate_wa
...
 
+    def metadata_from_stac_object(self, stac_object: pystac.STACObject) -> CubeMetadata:
+        """
+        Build cube metadata from a STAC object.
+        """
+        bands = self.bands_from_stac_object(stac_object)
+        dimensions = self.dimensions_from_stac_object(stac_object=stac_object, bands=bands)
+        return CubeMetadata(dimensions=dimensions)
+
+    def dimensions_from_stac_object(self, stac_object: pystac.STACObject, bands: _BandList) -> List[Dimension]:
```
soxofaan: Why having `bands` as argument here? Shouldn't it be the responsibility of this method to detect the bands? Note that this is a public method, so it's weird to have an argument for something you actually want to extract

openeo/metadata.py
```diff
@@ -760,39 +743,172 @@ def __init__(self, *, logger=_log, log_level=logging.DEBUG, supress_duplicate_wa
...
+            if not any(isinstance(d, BandDimension) for d in dimensions) and isinstance(
+                stac_object, (pystac.Collection, pystac.Item)
+            ):
+                dimensions.append(BandDimension(name="bands", bands=list(bands)))
+            return dimensions
+
+        dimensions: List[Dimension] = [
+            SpatialDimension(name="x", extent=[None, None]),
+            SpatialDimension(name="y", extent=[None, None]),
+            TemporalDimension(name="t", extent=self.infer_temporal_extent(stac_object)),
```
soxofaan: reuse `get_temporal_dimension` here instead ?

openeo/metadata.py
```diff
@@ -760,39 +743,172 @@ def __init__(self, *, logger=_log, log_level=logging.DEBUG, supress_duplicate_wa
...
+        except Exception:
+            ext = None
+
+        if not ext:
+            extra = getattr(dim, "extra_fields", {}) or {}
+            ext = extra.get("extent")
+
+        return ext or [None, None]
+
+    def parse_declared_dimensions(self, stac_object: pystac.STACObject, bands: _BandList) -> List[Dimension]:
```
soxofaan: as noted above, it looks weird to have this required `bands` argument here without explanation

soxofaan (commented):
some more notes

tests/test_metadata.py
suriyahgit: Added a STAC 1.1 `bands` common metadata test without the datacube extension in `test_metadata_from_stac_stac_1_1_common_bands_without_datacube_extension`. The test verifies that `metadata_from_stac()` still builds the fallback openEO dimensions and extracts the common `bands` metadata correctly.

openeo/metadata.py
suriyahgit: Removed this conditional injection. When `cube:dimensions` is present, the declared dimensions are now treated as the source of truth and no fallback `bands` dimension is added.

openeo/metadata.py
suriyahgit: Agreed. I made these parser helpers private: `_has_cube_dimensions()`, `_cube_dimensions_dict()`, and `_parse_declared_dimensions()`.

openeo/metadata.py
suriyahgit: Changed this to use `stac_obj.properties` directly for items.

openeo/metadata.py
suriyahgit: updated the item temporal fallback logic so `start_datetime` / `end_datetime` take precedence over `datetime` and added test coverage for that precedence.

openeo/metadata.py
suriyahgit: Agreed. `dimensions_from_stac_object()` now detects bands internally through `bands_from_stac_object()` instead of requiring a `bands` argument. I didn't notice that this `bands_from stac_object()` exists earlier, my bad.

openeo/metadata.py
suriyahgit: removed the usage of this temporal extent helper, and reusing `get_temporal_dimension` here instead.

openeo/metadata.py
suriyahgit: Removed the required `bands` argument from the public-facing parser method. `dimensions_from_stac_object()` now detects bands internally through `bands_from_stac_object()`, no longer needed to pass extracted bands into the method. This behaviour is consistent across PR now!

clausmichele:
@soxofaan @jdries this PR is ready to review when you can! This issue reminded me about this pending PR: https://forum.dataspace.copernicus.eu/t/load-stac-time-dimension-hidden/5312/3

openeo/metadata.py
```diff
@@ -766,39 +749,159 @@ def __init__(self, *, logger=_log, log_level=logging.DEBUG, supress_duplicate_wa
...
+            dt = props.get("datetime")
+            if dt:
+                norm = Rfc3339(propagate_none=True).normalize(dt)
+                return TemporalDimension(name="t", extent=[norm, norm])
+
+    def _has_cube_dimensions(self, stac_object: pystac.STACObject) -> bool:
+        cube_dimensions = self._cube_dimensions_dict(stac_object)
+        return isinstance(cube_dimensions, dict) and len(cube_dimensions) > 0
+
+    def _cube_dimensions_dict(self, stac_object: pystac.STACObject) -> Dict[str, dict]:
```
pjonsson, line 818: Can this be a static method?

You should probably check if there are other methods added by this PR that could also be made static as consequence of making this method static.

openeo/metadata.py
```diff
@@ -766,39 +749,159 @@ def __init__(self, *, logger=_log, log_level=logging.DEBUG, supress_duplicate_wa
...
+            return stac_object.extra_fields.get("cube:dimensions", {}) or {}
+        return {}
+
+    @staticmethod
+    def _safe_extent_from_pystac_cube_dim(dim) -> list:
+        """
+        PySTAC cube dimension wrapper may raise if 'extent' is missing.
+        Also, depending on serialization/version, extent might live in extra_fields.
+        """
+        try:
```
pjonsson, line 834: Can this try-block be replaced by:
```python
ext = getattr(dim, "extent", None)
```
and the code will do the same thing?

openeo/metadata.py
```diff
@@ -766,39 +749,159 @@ def __init__(self, *, logger=_log, log_level=logging.DEBUG, supress_duplicate_wa
...
+            ext = dim.extent
+        except Exception:
```
pjonsson, line 836: If the `getattr()` I suggested in the previous comment isn't right, catching `Exception` here is too broad, the code should catch the exception it's expecting and let other exceptions through so the caller can deal with unexpected errors.

openeo/metadata.py
```diff
@@ -766,39 +749,159 @@ def __init__(self, *, logger=_log, log_level=logging.DEBUG, supress_duplicate_wa
...
+
+    def _parse_declared_dimensions(self, stac_object: pystac.STACObject, bands: _BandList) -> List[Dimension]:
+        """
+        Parse dimensions declared through cube:dimensions.
+        """
+        if (
+            _PYSTAC_1_9_EXTENSION_INTERFACE
+            and getattr(stac_object, "ext", None) is not None
+            and stac_object.ext.has("cube")
+            and hasattr(stac_object.ext, "cube")
```
pjonsson, line 853: Does this line do the same as the previous line?

openeo/metadata.py
```diff
@@ -766,39 +749,159 @@ def __init__(self, *, logger=_log, log_level=logging.DEBUG, supress_duplicate_wa
...
-                name, extent = temporal_dims[0]
-                return TemporalDimension(name=name, extent=extent)
+                dimensions.append(BandDimension(name=name, bands=list(bands)))
+
+        return dimensions
+
+    def _parse_cube_dimensions_from_raw_dict(self, stac_object: pystac.STACObject, bands: _BandList) -> List[Dimension]:
+        """
+        Parse dimensions from raw cube:dimensions dict.
+        Supports 'spatial', 'temporal', and ('bands' or 'spectral' as an alias).
```
pjonsson, line 883: Parenthesis error? The sentence ends with an "and".

openeo/metadata.py
```diff
@@ -766,39 +749,159 @@ def __init__(self, *, logger=_log, log_level=logging.DEBUG, supress_duplicate_wa
...
+        Return raw cube:dimensions dict from a Collection/Item, or {}.
+        """
+        if isinstance(stac_object, pystac.Item):
+            return stac_object.properties.get("cube:dimensions", {}) or {}
+        if isinstance(stac_object, pystac.Collection):
+            return stac_object.extra_fields.get("cube:dimensions", {}) or {}
+        return {}
+
+    @staticmethod
+    def _safe_extent_from_pystac_cube_dim(dim) -> list:
```
pjonsson, line 829: Can't use `list` in Python 3.8.

And since the old code is at least partially annotated, I think it's nice if you add type annotations for the new functions/methods in this PR.

clausmichele:
@jdries @soxofaan we would really need this PR to be merged, we addressed all the reviews from VITO months ago already.

jdries (approved):

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- `metadata_from_stac()` now keeps declared STAC `cube:dimensions` as the dimension source of truth and handles STAC 1.1 common `bands` metadata without requiring the datacube extension ([#743](https://github.com/Open-EO/openeo-python-client/issues/743), [#867](https://github.com/Open-EO/openeo-python-client/pull/867)).
```
soxofaan: the addition of this changelog item looks out of sync with the current master branch where "unreleased" is targeting version 0.53.0

soxofaan (commented):
I think latest master has to be merged in this feature branch

soxofaan:
@pjonsson  also raised some review notes that need some closure

openeo/metadata.py
clausmichele: @pjonsson it is covering an edge case when one syntax doesn't find the extension.

openeo/metadata.py
clausmichele: Yes, suggestion applied

openeo/metadata.py
clausmichele: Switched to List[Any]. Added annotation for dim.

openeo/metadata.py
clausmichele: Yes. Corrected also `_band_from_eo_bands_metadata`

openeo/metadata.py
clausmichele: Fixed

CHANGELOG.md
clausmichele: Fixed

clausmichele:
@soxofaan @pjonsson your review points have been covered

soxofaan:
(I ran this PR through a copliot review and it uncovered a bug/inconsistency. Quoting the summary below)

---

Thanks for the refactor — the dimension-resolution policy (STAC `cube:dimensions` as source of truth, openEO defaults as fallback) is a nice clarification of the previous behavior, and the test coverage for the policy switch looks good. Ran the full `test_metadata.py` and the STAC-related `test_connection.py` tests locally and everything passes.

One correctness issue I'd like to flag before merge:

### `geometries` cube dimensions get mislabeled as band dimensions

In `_parse_cube_dimensions_from_pystac_extension`, the comment says:

> Important: PySTAC DimensionType only has SPATIAL + TEMPORAL. Everything else is treated as band-like.

That's not quite right — pystac's `datacube.DimensionType` also defines `GEOMETRIES` (for vector-cube dimensions, per the [datacube extension spec](https://github.com/stac-extensions/datacube)). Since the code only special-cases `SPATIAL`/`TEMPORAL` and treats everything else as a `BandDimension`, a declared `"geometries"`-typed dimension ends up being turned into a `BandDimension` when the STAC object's `stac_extensions` list includes the datacube extension URL.

Repro:
```python
stac_dict = {
    ...,
    "stac_extensions": ["https://stac-extensions.github.io/datacube/v2.2.0/schema.json"],
    "cube:dimensions": {
        "geom": {"type": "geometries"},
        "x": {"type": "spatial", "axis": "x", "extent": [0, 1]},
        "y": {"type": "spatial", "axis": "y", "extent": [0, 1]},
    },
}
# metadata_from_stac(...) -> "geom" comes back as BandDimension, not a generic/geometry Dimension
```

This is also inconsistent with the sibling method  _parse_cube_dimensions_from_raw_dict  (used when the datacube extension isn't declared in  stac_extensions ), which correctly falls back to a generic  Dimension(name, type=dim_type)  for unrecognized types — matching the existing legacy pattern already used elsewhere in this file (the older  cube:dimensions  parsing around line ~600, which does  complain("Unknown dimension type ...")  + generic  Dimension ). So the same STAC metadata can currently produce different (and in one case wrong) results depending only on whether  stac_extensions  happens to list the datacube extension URL.

Suggestion: in the pystac-extension branch, only map to  BandDimension  for the actual bands type, and fall back to a generic  Dimension(name, type=dim_type)  (optionally with a  complain /warn like the legacy code) for anything else, including  GEOMETRIES . A regression test with a  "geometries" -typed  cube:dimensions  entry (with  stac_extensions  set) would help lock this in.

Minor/non-blocking:

•  get_temporal_dimension 's internal  _has_cube_dimensions(...)  branch appears unreachable in the current call flow, since  dimensions_from_stac_object  already short-circuits to  _parse_declared_dimensions  before ever calling  get_temporal_dimension  when  cube:dimensions  is present. Worth a comment or cleanup for clarity.
• New test  test_metadata_from_stac_stac_1_1_common_bands_without_datacube_extension  is missing a blank line before the following  @pytest.mark.skipif  block.

soxofaan:
to not further stall this PR, I decided to just resolve the issue with the geometry dimension myself on a rebased version of this feature branch

merged in master with a0db65d91bee5706d420eb7cf87359c83791ae9b

thanks all!


it's unfortunate that github represents the state of this PR as "closed" instead of "merged", but I did the rebase to clean up the history from the pretty old origin of this PR

PR #885, opened 2026-04-02
Issue840 splitting custom states

HansVRP:
Draft PR; 

The main logic now looks at a seperate column which only contains the backend-status. The status column remained unchanged. 

I did reshuffle the columns to place the internal logic ones all the way to the right.


HansVRP:
Draft PR; all feedback is already welcome @soxofaan @VictorVerhaert

HansVRP:
(preliminary PR to spark discussion )

Issue #883, opened 2026-03-31
Support `PredefinedTileGrid` in `_job_splitting`


VincentVerelst:
Relevant for: https://github.com/ESA-APEx/apex_documentation/issues/182

Currently `split_area` only supports `SizeBasedTileGrid`. It would be convenient if users can also give their own grid as input via a `PredefinedTileGrid`.

Issue #717, opened 2025-01-28
Drop support for Python 3.8


soxofaan:
Python 3.8 reached end of life October last year ([2024-10-07](https://devguide.python.org/versions/#unsupported-versions)).

This ticket is to collect and cross-reference reasons for dropping support.

Previous editions:
- https://github.com/Open-EO/openeo-python-client/issues/578
- https://github.com/Open-EO/openeo-python-client/issues/460
- https://github.com/Open-EO/openeo-python-client/issues/210

soxofaan:
big reason we can not drop it yet: most deployments of openeo geopyspark driver are still bound to 3.8

soxofaan:
problem in #714: geopandas dropped python 3.8 support in 0.14.0 (Sept 2023)

soxofaan:
another possible problem as we are depending more and more on stac/pystac: pystac dropped python 3.8 support in version 1.9.0 (2023-10-23)

soxofaan:
Ding dong, the witch is dead

PR #886, opened 2026-04-02
add support for predefinedtilegrid in split_area #883

VincentVerelst:
The idea is to expose the `split_area` function to users. Users can provide an AOI, which can be split in different ways:

- By providing a `projection` and `tile_size`: splits the AOI up in square tiles
- By providing a `PredefinedTileGrid` (can also be a `geopandas.GeoDataFrame`), which splits the AOI according to that grid
- By providing a custom subclass of `TileGridInterface`: splits the AOI according to their custom class

The function returns the split AOI as a `geopandas.GeoDataFrame`


openeo/extra/job_management/_job_splitting.py
```diff
@@ -1,143 +1,260 @@
...
+        :param geometry: area of interest as a bounding-box dict
+            (with keys ``west``, ``south``, ``east``, ``north``, and optionally ``crs``),
+            a :class:`~shapely.geometry.Polygon`, or a :class:`~shapely.geometry.MultiPolygon`.
+        :return: GeoDataFrame with one row per tile and CRS set.
+        """
         ...
 
 
-class _SizeBasedTileGrid(_TileGridInterface):
+class SizeBasedTileGrid(TileGridInterface):
```
HansVRP: any reason you no longer want to use private classes?

openeo/extra/job_management/_job_splitting.py
```diff
@@ -1,143 +1,260 @@
...
+        """
+        Return the easting offset for the projection, used to align tiles to the
+        coordinate system's grid origin.
+
+        - UTM zones have a false easting of 500 000 m.
+        - EPSG:3857 and EPSG:4326 have no offset.
+
+        :raises JobSplittingFailure: for unsupported EPSG codes.
+        """
+        if self.epsg in self._UTM_NORTH or self.epsg in self._UTM_SOUTH:
```
HansVRP: how will this handle antimeredian containg shapes? Might be wise to add some niche unit tests

setup.py
```diff
@@ -90,6 +87,9 @@
         "deprecated>=1.2.12",
         'oschmod>=0.3.12; sys_platform == "win32"',
         "importlib_resources; python_version<'3.9'",
+        # TODO #717 Simplify geopandas constraints when Python 3.8 support is dropped
```
HansVRP, line 90: Do we need to rely on python 3.8 compatibility?

openeo/extra/job_management/_job_splitting.py
```diff
@@ -1,143 +1,260 @@
...
+        Supported values: ``4326`` (WGS84 degrees), ``3857`` (Web Mercator meters),
+        and UTM zones (``32601``–``32660``, ``32701``–``32760``).
+    :param size: tile size in the unit of measure of the projection.
     """
 
+    # EPSG ranges for UTM zones
+    _UTM_NORTH = range(32601, 32661)
+    _UTM_SOUTH = range(32701, 32761)
+
     def __init__(self, *, epsg: int, size: float):
```
HansVRP, line 189: what happens if someone provides a negative or 0 tilesize here?

HansVRP: Float is needed for support to lat-lon I presume?

openeo/extra/job_management/_job_splitting.py
```diff
@@ -1,143 +1,260 @@
...
+        crs: Optional[int] = None,
+    ):
+
+        if isinstance(tiles, gpd.GeoDataFrame):
+            self._gdf = tiles.copy()
+        elif isinstance(tiles, gpd.GeoSeries):
+            self._gdf = gpd.GeoDataFrame(geometry=tiles)
+        elif isinstance(tiles, list):
+            if not tiles:
+                raise JobSplittingFailure("At least one tile geometry must be provided.")
```
HansVRP, line 301: do we also check if these tile geometries are valid for openEO?

openeo/extra/job_management/_job_splitting.py
```diff
@@ -1,143 +1,260 @@
...
+            self._gdf = gpd.GeoDataFrame(geometry=tiles, crs=f"EPSG:{normalize_crs(crs)}")
         else:
-            raise JobSplittingFailure("geometry must be a dict or a shapely.geometry.Polygon or MultiPolygon")
+            raise JobSplittingFailure(
+                f"Expected a GeoDataFrame, GeoSeries, or list of geometries, got {type(tiles).__name__}."
+            )
 
-        # TODO: being a meter based EPSG does not imply that offset should be 500_000
-        x_offset = 500_000 if self._epsg_is_meters() else 0
+        if self._gdf.empty:
```
HansVRP, line 312: what if the crs of _gdf is differnt from that of the tiles?

openeo/extra/job_management/_job_splitting.py
```diff
@@ -1,143 +1,260 @@
...
+            geom = BBoxDict.from_dict(geometry).as_polygon()
+        elif isinstance(geometry, (Polygon, MultiPolygon)):
+            geom = geometry
+        else:
+            raise JobSplittingFailure(
+                f"Expected a bounding-box dict, Polygon, or MultiPolygon, got {type(geometry).__name__}."
+            )
+
+        mask = self._gdf.intersects(geom)
+        return self._gdf.loc[mask].copy().reset_index(drop=True)
```
HansVRP, line 333: what will we hand back if the intersection is empty?

openeo/extra/job_management/_job_splitting.py
```diff
@@ -1,143 +1,260 @@
...
-        return cls(*polygon.bounds, crs=4326 if crs is None else crs)
-
-    def as_dict(self) -> Dict:
-        return self._asdict()
-
-    def as_polygon(self) -> Polygon:
-        """Get bounding box as a shapely Polygon"""
-        return shapely.geometry.box(minx=self.west, miny=self.south, maxx=self.east, maxy=self.north)
-
+class TileGridInterface(metaclass=abc.ABCMeta):
```
HansVRP: do we (need to) provide guidance to users on optimal splitting? Just wondering if we will run into user who splitted into 1 km blocks and have giant dataframes as a result

openeo/extra/job_management/_job_splitting.py
```diff
@@ -1,143 +1,260 @@
...
+        elif isinstance(geometry, (Polygon, MultiPolygon)):
+            bbox = BBoxDict.from_any(geometry, crs=self.epsg)
+        else:
+            raise JobSplittingFailure(
+                f"Expected a bounding-box dict, Polygon, or MultiPolygon, got {type(geometry).__name__}."
+            )
+
+        x_offset = self._get_x_offset()
+        polygons = self._split_bounding_box(to_cover=bbox, x_offset=x_offset, tile_size=self.size)
+        return _to_geodataframe(polygons, epsg=self.epsg)
```
HansVRP: what is the performance here for huge grids?

tests/extra/job_management/test_job_splitting.py
```diff
@@ -1,168 +1,265 @@
...
+        assert isinstance(result, gpd.GeoDataFrame)
+        assert len(result) == 2
+        assert result.crs.to_epsg() == 4326
+
+    def test_split_area_3857(self):
+        aoi = {"west": 0.0, "south": 0.0, "east": 20_000.0, "north": 20_000.0, "crs": "EPSG:3857"}
+        result = split_area(aoi, projection="EPSG:3857", tile_size=20_000.0)
+        assert len(result) == 1
+        assert result.geometry[0].equals(shapely.geometry.box(0.0, 0.0, 20_000.0, 20_000.0))
+        assert result.crs.to_epsg() == 3857
```
HansVRP, line 329: DO we also have coverage for the antimeridian, custom grids with non standard crs, multipolygons; non or only partly intesecting polygons and tile grids ?

openeo/extra/job_management/_job_splitting.py
HansVRP: similarly I an wondering if we support polar regions; There ESPG:3657 becomes distorted

openeo/extra/job_management/_job_splitting.py
```diff
@@ -1,143 +1,260 @@
...
+
+        - UTM zones have a false easting of 500 000 m.
+        - EPSG:3857 and EPSG:4326 have no offset.
+
+        :raises JobSplittingFailure: for unsupported EPSG codes.
+        """
+        if self.epsg in self._UTM_NORTH or self.epsg in self._UTM_SOUTH:
+            return 500_000.0
+        elif self.epsg in (3857, 4326):
+            return 0.0
```
HansVRP: I am wondering if this anchor point should be configurable; Let's say two different users have a slightly different AOI; when they run the job manager they might get different tiling grids. 

Knowing that tile artefacts may occur; their results could become quite incompatible.

openeo/extra/job_management/_job_splitting.py
```diff
@@ -1,143 +1,260 @@
...
+        a :class:`~shapely.geometry.Polygon`, or
+        a :class:`~shapely.geometry.MultiPolygon`.
+    :param projection: EPSG string (e.g. ``"EPSG:3857"``) for the tile grid.
+        Required when *tile_grid* is not supplied and the AOI has no ``crs`` field.
+    :param tile_size: tile edge length in the unit of the projection.
+        Required when *tile_grid* is not supplied.
+    :param tile_grid: a :class:`TileGridInterface` instance or a
+        :class:`~geopandas.GeoDataFrame` (with CRS set) that defines the tiling
+        strategy.  Mutually exclusive with *projection* / *tile_size*.
+    :return: :class:`~geopandas.GeoDataFrame` with one row per tile and CRS set.
```
HansVRP: do we return in the orginal crs?

openeo/extra/job_management/_job_splitting.py
```diff
@@ -1,143 +1,260 @@
...
+        if isinstance(aoi, dict) and "crs" in aoi:
+            projection = aoi["crs"]
+        else:
+            raise JobSplittingFailure(
+                "'projection' is required when the area of interest does not contain a 'crs' field."
+            )
 
-    tile_grid = _SizeBasedTileGrid.from_size_projection(size=tile_size, projection=projection)
-    return tile_grid.get_tiles(aoi)
+    grid = SizeBasedTileGrid(epsg=normalize_crs(projection), size=tile_size)
```
HansVRP: Do we return tiles that extent beyond the AOI?

tests/extra/job_management/test_job_splitting.py
```diff
@@ -1,168 +1,265 @@
...
-
-    def test_as_polygon(self):
-        bbox = _BoundingBox(1, 2, 3, 4)
-        polygon = bbox.as_polygon()
-        assert isinstance(polygon, shapely.geometry.Polygon)
-        assert set(polygon.exterior.coords) == {(1, 2), (3, 2), (3, 4), (1, 4)}
+from openeo.util import BBoxDict
 
 
 class TestSizeBasedTileGrid:
```
HansVRP, line 15: would be interesting to add a land --> continental usecase to the test suite to check actual behavior

openeo/extra/job_management/_job_splitting.py
```diff
@@ -1,143 +1,260 @@
...
+            raise JobSplittingFailure(
+                f"Expected a bounding-box dict, Polygon, or MultiPolygon, got {type(geometry).__name__}."
+            )
+
+        x_offset = self._get_x_offset()
+        polygons = self._split_bounding_box(to_cover=bbox, x_offset=x_offset, tile_size=self.size)
+        return _to_geodataframe(polygons, epsg=self.epsg)
+
+
+class PredefinedTileGrid(TileGridInterface):
```
HansVRP: How do we handle overlapping tiles or invalid tiles? 

Would be useful to have a validate_tile_grid() functionality in the mix

tests/extra/job_management/test_job_splitting.py
```diff
@@ -1,168 +1,265 @@
+import geopandas as gpd
```
HansVRP, line 1: do we have test coverage on the behavior if the AOI is a line or a single point?

HansVRP (commented):
Some initial comments and questions

openeo/extra/job_management/_job_splitting.py
VincentVerelst: You're right. Made all classes private again to allow for more flexibility in the future. Only `split_area` is now part of the public API

openeo/extra/job_management/_job_splitting.py
VincentVerelst: This is why I rely on the `BboxDict` util class for `SizeBasedTileGrid`. 

For `PreDefinedTileGrid` it's the user's responsibility that their geometries in the grid are valid for their process graph. It's difficult to predict this from the job splitter alone. It depends on which processes are used in the process graph (`filter_bbox`, `aggregate_spatial`, `filter_spatial`, ...)

openeo/extra/job_management/_job_splitting.py
VincentVerelst: AOI's are always split and returned according to the underlying grid that is used for splitting. Whether the grid is size-based or pre-defined

openeo/extra/job_management/_job_splitting.py
VincentVerelst: No, tiles are capped at the extents of the original AOI

tests/extra/job_management/test_job_splitting.py
VincentVerelst: What do you mean by this?

openeo/extra/job_management/_job_splitting.py
VincentVerelst: Should we disallow overlapping tiles? What if a user specifically wants that? 

On invalid tiles: see my comment above, how would we identify when a tile is 'invalid'?

setup.py
VincentVerelst: https://github.com/Open-EO/openeo-python-client/issues/717

It's needed for as long as the backend supports it

openeo/extra/job_management/_job_splitting.py
VincentVerelst: Added a check for negative tile sizes. Float is indeed needed for lat-lon

tests/extra/job_management/test_job_splitting.py
VincentVerelst: Yes the `TileGrid` tests covered this case. Also added a test for `split_area`.  Current behaviour: only `dict`, `Polygon` or `MultiPolygon` is allowed

openeo/extra/job_management/_job_splitting.py
VincentVerelst: Added a validation an raise an error when user specified an antimeridian crossing AOI

VincentVerelst:
@HansVRP , except for addressing your comments, one big change: there is now a standardized reprojecting happening to the CRS of the tile grid that is being used for splitting. 

To re-iterate what will be exposes to the public API: the `split_area` function now takes as input an AOI, which can be split in either of two ways:

- By providing a tile-grid as `geopandas.GeoDataFrame` or a list of `(Multi)Polygon`. In the background `_PredefinedTileGrid` will be used for splitting
- By providing a `projection` and `tile_size`. In the background `_SizeBasedTileGrid` will be used for splitting. The origin will be taken to be the bottom-left of the AOI provided by the user. 

Advanced users can always define their own splitting method by defining a subclass of `_TileGridInterface`, which is still experimental and prone to change

HansVRP:
will look into the update; thanks. 

can you also include a usecase example in: https://github.com/Open-EO/openeo-python-client/blob/master/docs/cookbook/job_manager.rst ?

openeo/extra/job_management/_job_splitting.py
HansVRP: so we do not support antimeridian AOIs then? 

is there any way in which we could support it?

setup.py
HansVRP: I am probably overlooking something; but this runs only client side? 

or is the limitation due to the geometry we send into the back-end

openeo/extra/job_management/_job_splitting.py
HansVRP: maybe good to mention this explicitly in the docstring (might have overread it)

tests/extra/job_management/test_job_splitting.py
HansVRP: performance wise I can image that splitting up france, germany or Australia is going to be demanding. 

Having a country based unit test could help us track regression if the splitting becomes slow?

openeo/extra/job_management/_job_splitting.py
HansVRP: valid comment; perhaps we can just add in the documentation a good example where we explain best practices, that way users are informed on how we see the 'ideal' grid?

tests/extra/job_management/test_job_splitting.py
HansVRP: perfect

HansVRP: if the dict contains for instance a line or point

the error will come when openEO tries to run it I assume?

openeo/extra/job_management/_job_splitting.py
VincentVerelst: I propose to log this in a separate issue as a further improvement. I prefer to move this out of scope for this PR

tests/extra/job_management/test_job_splitting.py
VincentVerelst: No the job splitter will fail, because the bbox is validated to must have east>west and north>south

openeo/extra/job_management/_job_splitting.py
HansVRP: sounds good; please link the ticket and make clear in the documentation that this is a known limitation

tests/extra/job_management/test_job_splitting.py
VincentVerelst: Instead of adding a deliberately time-consuming unit test, I now print a warning when the resulting grid will be larger than 500x500 tiles

openeo/extra/job_management/_job_splitting.py
VincentVerelst: added

tests/extra/job_management/test_job_splitting.py
HansVRP: also be wary with coordinates that extend past ±180° longitude, it might be these slip through your guard

tests/extra/job_management/test_job_splitting.py
HansVRP: There still might be some niche cases we will encounter: 
- e.g. inputs in a geographic CRS with coordinates that extend past ±180° longitude, this breaks the assumption that west < east
- Predefined tiles in a geographic CRS that wrap past the antimeridian

Couple of things we could do as follow up: 

if west>east --> spliting two polygons and further split those seperatly. This seems low effort to me.

input polygons with lon past +-180 will be harder; I believe @soxofaan was also looking into this for the python client.

HansVRP: (made a small commit which I believe handles this: https://github.com/Open-EO/openeo-python-client/pull/886/changes/1e0c1781cbe4f6e9fddfe243092846d27060559e)

openeo/extra/job_management/_job_splitting.py
VincentVerelst: The crs from `self._gdf` is used, unless it's not present, then the optional `crs` parameter must be given and the crs is derived from there, otherwise an error is thrown

openeo/extra/job_management/_job_splitting.py
VincentVerelst: an empty dataframe

openeo/extra/job_management/_job_splitting.py
VincentVerelst: There is a warning thrown if the amount of rows or columns is larger than 500, with an indication that it's better to select a lager tile size

openeo/extra/job_management/_job_splitting.py
VincentVerelst: There will be a warning thrown if the grid will be larger than 500x500. Even then it's only a matter of a few seconds.

tests/extra/job_management/test_job_splitting.py
VincentVerelst: This is now covered by your commit I believe

openeo/extra/job_management/_job_splitting.py
VincentVerelst: This is indeed outdated. There is no fixed anchor point anymore. The anchor point is taken as the bottom left of the user's AOI

openeo/util.py
```diff
@@ -534,6 +534,10 @@ def __init__(self, *, west: float, south: float, east: float, north: float, crs:
 
     # TODO: provide west, south, east, north, crs as @properties? Read-only or read-write?
 
+    def as_polygon(self) -> shapely.geometry.Polygon:
```
HansVRP, line 537: @soxofaan okay to add this here?

HansVRP (approved):

openeo/util.py
soxofaan: yes, fine to add that here.

However, note that it's tricky to get this correct+intuitive for antimeridian use cases.
e.g. in antimeridian use cases, you have `west > east` (so min and max longitude are switched). Your polygon will then probably the opposite of what you want.

There are several possible solutions/guardrails here:
- explicit `assert west < east`  so that you get hard fail on anti-meridian cases, instead of blindly producing wrong results
- produce a multi-polygon with two parts, one for each side of the antimeridian
- add 360 to `east` so you preserve the `west < east` assumption

soxofaan: FYI: also look at comparable utility in openeo_driver:
https://github.com/Open-EO/openeo-python-driver/blob/b7d0f7bea15a2c4bec03ee71f11c416854bc775e/openeo_driver/util/geometry.py#L590-L617

VincentVerelst: Comitted something that uses the exact same antimeridian crossing logic as the python driver and a split between `as_polygon` and `as_geometry`

PR #898, opened 2026-05-11
Support openeo.api.process.Parameter as resolution in DataCube.resamp…

### LINKED ISSUE #897: Allow `openeo.api.process.Parameter` object as `resolution` in `DataCube.resample_spatial()`
Version: 0.49.0

**Problem**
The following script will fail

```python
import openeo 
from openeo.api.process import Parameter

c = openeo.connect("openeo.dataspace.copernicus.eu").authenticate_oidc()

s2 = c.load_collection("SENTINEL2_L2A", spatial_extent={"west": 16.0, "south": 48.0, "east": 17.0, "north": 49.0}, temporal_extent=["2021-01-01", "2021-12-31"])

res_param = Parameter.number(
    name='resolution'
)

s2 = s2.resample_spatial(
    resolution=res_param
)
```

Because `normalize_resample_resolution` doesn't allow for an `openeo.api.process.Parameter`: https://github.com/Open-EO/openeo-python-client/blob/63ac7837f1b41e30950fdc5cc986b0137962bca5/openeo/utils/normalize.py#L4

**Solution**

Support `openeo.api.process.Parameter` in `openeo.utils.normalize.normalize_resample_resolution()` by returning something like `(None, None)` in that case. 

CC: @Pratichhya

VincentVerelst:
Fixes https://github.com/Open-EO/openeo-python-client/issues/897


openeo/metadata.py
```diff
@@ -512,15 +516,25 @@ def resample_spatial(
...
+                    step=r if r != 0.0 else dim.step,
+                )
+        elif projection:
+            for i in spatial_indices:
+                dim: SpatialDimension = dimensions[i]
+                dimensions[i] = SpatialDimension(
+                    name=dim.name,
+                    extent=dim.extent,
+                    crs=projection,
+                    step=dim.step,
```
soxofaan: I think it's better to set `step` to `None` to more clearly signal that it is unknown at this point, instead of leaving it unchanged (which is the opposite of what is probably happening)

openeo/metadata.py
```diff
@@ -494,10 +495,13 @@ def drop_dimension(self, name: str = None) -> CubeMetadata:
 
     def resample_spatial(
         self,
-        resolution: Union[float, Tuple[float, float], List[float]] = 0.0,
+        resolution: Union[float, Tuple[float, float], List[float], "Parameter"] = 0.0,
         projection: Union[int, str, None] = None,
```
soxofaan: Also allow projection to be a parameter here, while we're at it?

openeo/metadata.py
```diff
@@ -494,10 +495,13 @@ def drop_dimension(self, name: str = None) -> CubeMetadata:
...
     ) -> CubeMetadata:
-        resolution = normalize_resample_resolution(resolution)
+        if isinstance(resolution, Parameter):
+            normalized_resolution = None
```
soxofaan:
```suggestion
            normalized_resolution = None, None
```

which will allow to simplify the logic below to practically the original code, I guess

tests/test_metadata.py
```diff
@@ -1279,6 +1280,35 @@ def test_metadata_resample_spatial(cube_metadata, kwargs, expected_x, expected_y
     assert metadata.band_dimension == cube_metadata.band_dimension
 
 
+@pytest.mark.parametrize("cube_metadata", [CUBE_METADATA_XYTB, CUBE_METADATA_TBXY])
+def test_metadata_resample_spatial_parameter_resolution_only(cube_metadata):
+    """When resolution is a Parameter, step should remain unchanged and crs should stay as-is."""
+    param = Parameter.number("res", description="The spatial resolution.")
+    metadata = cube_metadata.resample_spatial(resolution=param)
+    assert isinstance(metadata, CubeMetadata)
+    # step must not change because the resolution is unknown at build time
```
soxofaan: I think step must be made explicitly unknown: e.g. `None`

openeo/metadata.py
VincentVerelst: Added, with the same logic that the metadata crs should become None, if `projection` is a `Parameter`

CHANGELOG.md
```diff
@@ -20,7 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Removed
 
 ### Fixed
-
+- `DataCube.resample_spatial()` now supports `openeo.api.process.Parameter` objects for the `resolution` and `projection` parameter.  ([#897](https://github.com/Open-EO/openeo-python-client/issues/897))
```
soxofaan:
```suggestion
- `DataCube.resample_spatial()` now supports parameterized `resolution` and `projection` arguments.  ([#897](https://github.com/Open-EO/openeo-python-client/issues/897))
```

CHANGELOG.md
```diff
@@ -20,7 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Removed
 
 ### Fixed
-
+- `DataCube.resample_spatial()` now supports parameterized `resolution` and `projection` arguments.  ([#897](https://github.com/Open-EO/openeo-python-client/issues/897))
```
soxofaan: I would put this under "added" as this feels more like an added feature than a fixed bug

also: I think this added changelog entry is under the wrong version (0.50 has been released in the meantime)

I think you should merge master in this feature branch first

openeo/metadata.py
```diff
@@ -512,13 +517,14 @@ def resample_spatial(
...
-        for i, r in zip(spatial_indices, resolution):
+
+        assert len(normalized_resolution) == 2
+        for i, r in zip(spatial_indices, normalized_resolution):
             dim: SpatialDimension = dimensions[i]
             dimensions[i] = SpatialDimension(
                 name=dim.name,
                 extent=dim.extent,
-                crs=projection or dim.crs,
+                crs=None if isinstance(projection, Parameter) else projection or dim.crs,
```
soxofaan: to avoid confusion:
```suggestion
                crs=None if isinstance(projection, Parameter) else (projection or dim.crs),
```

soxofaan (commented):
some more quick nitpicks

PR #900, opened 2026-05-27
clarification

HansVRP:


VincentVerelst (approved):

docs/cookbook/job_manager.rst
```diff
@@ -266,6 +266,42 @@ This will create a GeoDataFrame with one row per tile, and a geometry column con
...
+    import geopandas as gpd
+    from openeo.extra.job_management import split_area
+
+    bbox = {"west": 5.0, "south": 51.0, "east": 5.2, "north": 51.2, "crs": "EPSG:4326"}
+
+    # Load a pre-computed tile grid (e.g. a pixel-aligned UTM grid stored as GeoParquet)
+    tile_grid = gpd.read_parquet("my_tile_grid.parquet")
+
+    # Only tiles intersecting the AOI are returned
+    gdf = split_area(aoi=bbox, tile_grid=tile_grid)
```
soxofaan: UI design consideration: 
you are using/mixing `aoi` and `bbox` here, while in most comparable parts of the openEO API we use `spatial_extent` (or just `extent` when it's already obvious from the context that it is a spatial extent).
It might be beneficial to consider improving consistency

soxofaan (commented):
minor note to consider

docs/cookbook/job_manager.rst
HansVRP: updated bbox to extent

HansVRP:
@soxofaan would you propose to also revise the aoi parameter name in split_area itself to spatial_extent?

soxofaan:
I don't know these utilities to be honest, I just want to promote internal consistency and alignment with the terminology of the openEO API. I have no idea of the impact of changing these things at the moment

HansVRP:
@VincentVerelst what do you think of revising aoi to spatial_extent ? or would it be confusing that a spatial_extent gets splitted up into spatial_extents

VincentVerelst:
@HansVRP, I missed your comment. `aoi` was just an initial proposal, for me it's fine to use `spatial_extent`, however `aoi` in `split_area` has already been made public, so it would introduce a breaking change

HansVRP:
given that its just new I do not think we'd break anything apart from a few community examples. I would however not do it in this PR. 

thoughts @soxofaan or would aoi be acceptable to stay here?

soxofaan:
if aoi is already part of public API, it's indeed out of scope here

PR #906, opened 2026-07-03
[codex] add DataCube inspect helper

### LINKED ISSUE #795: Easy way to invoke 'inspect' in datacube
The 'inspect' method allows to log properties of a datacube, which is useful for debugging.
It is however not so easy to invoke it. What I would intuitively expect based on other processes, is a simple `cube.inspect()`. 

Note that importing 'inspect' from `openeo.processes` is an option, but invoking that method returns a `ProcessBuilder` object, which does not have the convenience functionality to start jobs.

The current workaround I found was to do this:

`connection.datacube_from_process("inspect",data=cube)`

therealclvn:
Closes #795.

## What changed

- Added `DataCube.inspect(message=..., code=..., level=...)` as a convenience wrapper around the openEO `inspect` process.
- The helper keeps returning a `DataCube`, so existing job/download/execute convenience methods remain available.
- Added process graph tests for both `cube.inspect()` and `cube.inspect(...)` with log fields.

## Why

Users currently have to call `cube.process("inspect", data=cube)` or use `connection.datacube_from_process(...)` for this debugging process. A direct `cube.inspect()` matches the expected DataCube method style and keeps the result in the DataCube API.

## Validation

- `python -m pytest tests/rest/datacube/test_datacube100.py -k 'custom_process or datacube_from_process or test_inspect' -q`
- `python -m pytest tests/rest/datacube/test_datacube100.py -q`
- `python -m compileall openeo/rest/datacube.py tests/rest/datacube/test_datacube100.py`
- `git diff --check`

This PR was drafted with help from OpenAI Codex.


tests/rest/datacube/test_datacube100.py
```diff
@@ -2896,6 +2896,55 @@ def test_custom_process_arguments_namespacd(con100: Connection, test_data):
...
+            "result": True,
+        },
+    }
+
+
+def test_inspect_without_log_fields(con100: Connection):
+    cube = con100.load_collection("S2")
+    res = cube.inspect()
+
+    assert res.flat_graph() == {
```
soxofaan: instead of `flat_graph`: use `get_download_graph` with `drop_load_collection=True` to make these asserts to the point

openeo/rest/datacube.py
```diff
@@ -137,6 +137,25 @@ def process(
...
+    def inspect(self, message=_UNSET, code=_UNSET, level=_UNSET) -> DataCube:
+        """
+        Add information to the logs for this data cube.
+
+        :param message: A message to send in addition to the data.
+        :param code: A label to help identify one or more log entries originating from this process in the list
+            of all log entries. It can help to group or filter log entries and is usually not unique.
+        :param level: The severity level of this message, defaults to `info`.
+        :return: new DataCube instance
+        """
```
soxofaan, line 151: add a `versionadded` annotation to the docs

soxofaan (commented):
thanks for contributing!

some notes for further finetuning.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Add `DataCube.inspect()` helper for the openEO `inspect` process. ([#795](https://github.com/Open-EO/openeo-python-client/issues/795))
```
soxofaan: append to the bottom of the item list

therealclvn: Moved it to the bottom of the Added list in 73621c00.

PR #908, opened 2026-07-06
Convert setup.py to pyproject.toml

pjonsson:
The same files are included before/after
when doing "python3 -m build" on
my machine.


pjonsson:
I can't see the log from [continuous-integration/jenkins/pr-head](https://jenkins.vgt.vito.be/job/openEO/job/openeo-python-client/job/PR-908/1/display/redirect), so not sure why that is red.

soxofaan:
from the error logs:

```
+ setuptools==75.3.4
 + setuptools-lint==0.6.0
 + toml==0.10.2
 + typing-extensions==4.13.2
 + wrapt==1.13.3
+ '[' -f requirements.txt ']'
+ '[' -f setup.py ']'
+ '[' -f setup.cfg ']'
+ uv pip install '.[tests]'
Using Python 3.8.8 environment at: venv38
  × Failed to build `openeo @
  │ [file:///var/lib/jenkins/workspace/enEO_openeo-python-client_PR-908`](file:///var/lib/jenkins/workspace/enEO_openeo-python-client_PR-908%60)
  ├─▶ The build backend returned an error
  ╰─▶ Call to `setuptools.build_meta.build_wheel` failed (exit status: 1)

      [stdout]
      configuration error: `project.license` must be valid exactly by one
      definition (2 matches found):

          - keys:
              'file': {type: string}
            required: ['file']
          - keys:
              'text': {type: string}
            required: ['text']

      DESCRIPTION:
          `Project license <https://peps.python.org/pep-0621/#license>`_.

      GIVEN VALUE:
          "Apache-2.0"
```



as far as I understand the license field should not be a string but a mapping with field "text" containing that string.
This is probably because the build chain used here is stuck at outdated conventions

soxofaan:
probably related to 
- https://github.com/Open-EO/openeo-python-client/issues/717

pjonsson:
@soxofaan ok, this revision of the PR works in a Python 3.8 container for me, and I've rebased to latest master.

pjonsson:
Still red in the CI despite building locally, so I'm guessing the CI logs that I can't see are trying to run `uv run pytest` after installing the tests dependencies.

Accepted any version of sphinx-autodoc-typehints for Python versions that have reached end of life, and skipped `openeo_pg_parser_networkx`/`openeo_processes_dask[implementations]` on Python 3.8: https://github.com/Open-EO/openeo-python-client/compare/15d8ce848ba2ec05ead03de7db0a5398b6d0ce4f..c18db737723efc375efd4d75916f1ec87a470eff and after that I can run the tests in the Python 3.8 container locally as well (and they pass).

pjonsson:
Apparently still red, but I can't guess what the CI does next, so I need to see some more logs.

soxofaan:
yes I was afraid this would be bit of a rabbit whole and wrestling with CI contexts built on outdated assumptions

the failure on https://github.com/Open-EO/openeo-python-client/pull/908/commits/8900d432d66156befeae3bb66dc6ad8f1d614e66  is an actual unit test failure:

```
tests.internal.test_warnings.test_user_deprecation_warning (from pytest)

assert 'This is a UserDeprecationWarning (stacklevel 1)' in 'Traceback (most recent call last):\n  File "myscript.py", line 1, in <module>\n    from openeo.internal.warnings impo... from openeo.rest.connection import Connection, connect, session\nModuleNotFoundError: No module named \'openeo.rest\''

Stacktrace

pytester = <Pytester PosixPath('/var/lib/jenkins/workspace/enEO_openeo-python-client_PR-908/pytest-tmp/pytest-of-jenkins/pytest-0/test_user_deprecation_warning0')>

    def test_user_deprecation_warning(pytester):
        pytester.makepyfile(myscript="""
            from openeo.internal.warnings import test_warnings
            test_warnings()
            test_warnings(2)
        """)
    
        result = pytester.runpython("myscript.py")
        stderr = "\n".join(result.errlines)
>       assert "This is a UserDeprecationWarning (stacklevel 1)" in stderr
E       assert 'This is a UserDeprecationWarning (stacklevel 1)' in 'Traceback (most recent call last):\n  File "myscript.py", line 1, in <module>\n    from openeo.internal.warnings impo... from openeo.rest.connection import Connection, connect, session\nModuleNotFoundError: No module named \'openeo.rest\''

/var/lib/jenkins/workspace/enEO_openeo-python-client_PR-908/tests/internal/test_warnings.py:17: AssertionError

```

soxofaan:
On a more general note: is there any particular reason you work on this? Are the current setup.py based wheel releases at e.g. https://pypi.org/project/openeo/  not sufficient in some sense? Do you want to make custom builds or something like that?

But to be clear: I definitively also want to move to pyproject.toml, but was planning to do that after we can drop the python 3.8 support (because that would eliminate some annoying intricacies in our internal CI setup)

pyproject.toml
```diff
@@ -1,7 +1,144 @@
...
+openeo-auth = "openeo.rest.auth.cli:main"
+
+[project.urls]
+"Bug Tracker" = "https://github.com/Open-EO/openeo-python-client/issues"
+Changelog = "https://github.com/Open-EO/openeo-python-client/blob/master/CHANGELOG.md"
+Documentation = "https://open-eo.github.io/openeo-python-client/"
+Homepage = "https://github.com/Open-EO/openeo-python-client"
+"Source Code" = "https://github.com/Open-EO/openeo-python-client"
+
+[build-system]
```
soxofaan: Isn't the general convention to put `build-system` at the top of the file?

pjonsson: I used hatch to generate `pyproject.toml` from `setup.py`, so this was where the previously commented out build-system ended up. I've moved it to the top like you suggest.

pjonsson:
>`         result = pytester.runpython("myscript.py")`
>`         stderr = "\n".join(result.errlines)`
>` >       assert "This is a UserDeprecationWarning (stacklevel 1)" in stderr`
>` E       assert 'This is a UserDeprecationWarning (stacklevel 1)' in 'Traceback (most recent call last):\n  File "myscript.py", line 1, in <module>\n    from openeo.internal.warnings impo... from openeo.rest.connection import Connection, connect, session\nModuleNotFoundError: No module named \'openeo.rest\''`

Since it can't find `openeo.rest` I'm guessing the CI doesn't install the package before starting the tests, so `PYTHONPATH` is probably faulty.

The test seems to still pass in the Python 3.8 container when I run with uv:
```
openeo-python-client $ uv run --python 3.8 pytest -vvv tests/internal/test_warnings.py::test_user_deprecation_warning
============================= test session starts ==============================
platform linux -- Python 3.8.20, pytest-8.3.5, pluggy-1.5.0 -- /files/openeo-python-client/.venv/bin/python3
cachedir: .pytest_cache
rootdir: /files/openeo-python-client
configfile: pytest.ini
plugins: time-machine-2.15.0, requests-mock-1.12.1
collected 1 item

tests/internal/test_warnings.py::test_user_deprecation_warning PASSED    [100%]

============================== 1 passed in 0.20s ===============================
```
so I think the test/package itself is fine.

> On a more general note: is there any particular reason you work on this?

Well, the immediate thing I wanted is a `py.typed` in this package (just opened #921), but my Python journey started with 3.10 so I don't know anything about Python package building predating that. I figured it would be easier to convert this package to `pyproject.toml` now, and since that is presumably where things will be moving in the long-term it wouldn't be wasted effort to make the conversion now.

Switching to `pyproject.toml` would also make it easier to use uv on this package, and that would make my existing PyCharm configuration work as-is more or less.

I think it's the first time I've seen a private CI used on GitHub and I assume you have good reasons for why things are set up the way they are, but the black-box turning red does raise the bar for third-party contributions (and seems to give the maintainers extra work) compared to regular GitHub actions.

.github/workflows/unittests.yml
```diff
@@ -42,6 +40,6 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        python -m pip install .[tests]
+        python -m pip install -e .[tests]
```
pjonsson: The error message looks similar so I think this is the root cause for the failing non-public CI as well, but I'm not sure I can fix that.

soxofaan: I'm not sure this change would solve anything

pyproject.toml
```diff
@@ -1,7 +1,144 @@
...
+Changelog = "https://github.com/Open-EO/openeo-python-client/blob/master/CHANGELOG.md"
+Documentation = "https://open-eo.github.io/openeo-python-client/"
+Homepage = "https://github.com/Open-EO/openeo-python-client"
+"Source Code" = "https://github.com/Open-EO/openeo-python-client"
+
+[tool.setuptools.dynamic]
+version = {attr = "openeo._version.__version__"}
+
+[tool.setuptools.packages.find]
+include = ["openeo"]
```
soxofaan: I dumped the failure logs in Claude and it suggested this line is the culprit for the failing test:

> include patterns are matched with fnmatch-style glob rules against dotted package names. "openeo" matches only the literal top-level package openeo, not any of its subpackages. So the built wheel ends up containing just openeo/__init__.py and friends, but drops openeo.rest, openeo.internal, etc. entirely.
> ...
> Fix: in PR [#908](https://github.com/Open-EO/openeo-python-client/pull/908)'s pyproject.toml, change:
>     [tool.setuptools.packages.find]
>     include = ["openeo*"]

FYI original from setup.py: https://github.com/Open-EO/openeo-python-client/blob/9d225d98dcaa880ea654fc385c2e8aead1b88199/setup.py#L76

pjonsson: As the PR description mentions, I checked that both files generated from `python3 -m build` contains the same files (except setup.py) before and after this change before I made a PR out of it.

I don't know enough to say whether the suggested change is better or worse, so I've updated the PR with your suggestion.

.github/workflows/unittests.yml
pjonsson: Well, that was the line I changed that made the GitHub actions go green for this PR, before that they had the same failure log as your internal CI does.

soxofaan: I think usage of `-e` hides potential packaging issues, as it uses a symlink of the full source tree to "install" the package instead of selectively copying files that are part of the packaging scope,  so the tests won't signal problems with that packaging scope

pjonsson: Based on your description of what happens, that sounds reasonable. Would you like me to drop this change?

I still don't understand how I can get a `dist/openeo-0.52.0a2.tar.gz` and a `dist/openeo-0.52.0a2-py3-none-any.whl` that have the same files/layout as the files I get on the current main branch, but somehow those "identical" packages install to something that differs from what is currently installed from the main branch. Do you have an explanation for what might be causing that?

Edit: and just to be clear, `openeo/internal/warnings.py`, `openeo/rest/connection.py`, and `openeo/rest/__init__.py` are all present in `dist/openeo-0.52.0a2-py3-none-any.whl`.

soxofaan: > Would you like me to drop this change

Yes, `-e` makes no sense in an automated CI environment, so I would avoid it.

> Do you have an explanation for what might be causing that?

I think it has to do with subtle differences between `pip install .` (install from source tree) and `pip install PACKAGE` (install built wheel) and how package discovery and file collection is handled in both cases.

So a better fix would actually be to first build a wheel and install that wheel. Then you are actually testing what you package and ship.

While this is possible in the GitHub actions workflow, our internal CI tool chain (which is Jenkins based instead of GitHub actions) is too opinionated to make that feasible at the moment.

So to not get lost in a rabbit hole here I would keep this PR focused:

- avoid `-e` in `pip install .`
- use `include=["openeo*"]` (as already done)

I think that is enough to get all checks green

docs/development.rst
```diff
@@ -289,7 +289,7 @@ we will use a concrete version ``0.8.0`` in the examples below.
         -   *Or, if you know what you are doing* and you're sure you have a clean
             local checkout, you can also build it locally::
 
-                python setup.py bdist_wheel
+                python -m build -w
```
soxofaan: why the `-w` here and not stick to the defaults?

also note that the "build-wheel.yml" workflow does not use `-w`

.github/workflows/unittests.yml
pjonsson: Ok, updated the PR to not include `-e`.

pjonsson:
Since there was a release about to happen this week or next, can this get merged now that you provided the fix to make it pass the internal CI?

soxofaan:
> Since there was a release about to happen this week or next, can this get merged now that you provided the fix to make it pass the internal CI?

I'm going to do that after that release, as quickly switching to pyproject.toml might jeopardize that (I just encountered some internal tooling that is not pyproject.toml-ready)

pyproject.toml
```diff
@@ -1,7 +1,144 @@
...
+    "sphinx-copybutton",
+]
+jupyter = [
+    "ipyleaflet>=0.17.0",
+    "ipython",
+]
+localprocessing = [
+    "openeo_pg_parser_networkx>=2023.5.1; python_version>='3.9'",
+    "openeo_processes_dask[implementations]>=2023.7.1; python_version>='3.9'",
+    "pyproj",
```
soxofaan, line 94: these `; python_version>='3.9'` constraints are new, compared with the original in setup.py.
why is that?

pyproject.toml
```diff
@@ -1,7 +1,144 @@
...
+    "types-boto3-sts",
+    "urllib3<2.3.0",
+]
+docs = [
+    "myst-parser",
+    "pydata_sphinx_theme",
+    "sphinx",
+    "sphinx-autodoc-annotation",
+    "sphinx-autodoc-typehints >=2.2.3; python_version>='3.10'",
+    "sphinx-autodoc-typehints; python_version<='3.9'",
```
soxofaan: these `python_version` constaints are new, compared with the original in setup.py
why is that?

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Lookup collection queryables only when property filtering is actually used in `load_collection` ([#926](https://github.com/Open-EO/openeo-python-client/issues/926))
+- Convert setup.py to pyproject.toml ([#920](https://github.com/Open-EO/openeo-python-client/issues/920))
```
soxofaan: this probably has to be moved, now that a new section was started on the master branch

soxofaan (commented):
some more notes

docs/development.rst
pjonsson: I'm just converting the existing things into using the build package, `python setup.py bdist_wheel` builds a wheel, and so does `python -m build -w`.

You're right that I missed the `-w` in `build-wheel.yml`, I'll update the pull request with that.

pyproject.toml
pjonsson: Version 2.1.0 of sphinx-autodoc-typehints dropped the support for Python 3.8, so I'm not sure how you manage to install the documentation extras with Python 3.8 right now.

pyproject.toml
pjonsson: It's been a couple of months since I made this PR and I don't remember the details, but some package in this bunch made the constraints non-solvable with Python 3.8. I'm guessing the places that still use Python 3.8 don't install this extra.

soxofaan: the problem is that expressing it like that suggests that these packages are not necessary for python 3.8.
So you will be able to install `openeo[localprocessing]` on python 3.8, but not be able to use it on python 3.8

I think it is better to *not* specify `python_version` constraints so the user gets a failure at install time (which is more correct) than at run time.

pyproject.toml
soxofaan: Like the other case: I think it is more correct to *not* specify a `python_version` constraint so that the user gets a failure at install time instead of installing a possibly broken dependency situation. 


or another way to put it: the openeo package should only specify what it minimum version of `sphinx-autodoc-typehints` it expects (`2.2.3` here). That this introduces extra constraints on `python_version` is an implementation detail of `sphinx-autodoc-typehints` that should not leak into the openeo package. 

> I'm not sure how you manage to install the documentation extras with Python 3.8 right now.

I think it's fine that it's not possible to build the docs on python 3.8. The claimed python 3.8 support is purely for the core package, not any of the extra's.

docs/development.rst
soxofaan: understood.

I however I think this change from `setup.py bdist_wheel` to `python -m build`  is an opportunity to do the building more correctly, which is without the `-w`(and which is not really possible with classic `setup.py` as far as I understand).
So I would remove the `-w` from both places instead of trying to replicate the suboptimal old way

.github/workflows/build-wheel.yml
```diff
@@ -20,9 +20,9 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install wheel
-    - name: Build bdist_wheel
-      run: python setup.py bdist_wheel
+        pip install build wheel
+    - name: Build wheel
+      run: python3 -m build -w
```
soxofaan: these three lines look bit inconsistent now:
```
python -m pip ...
pip ...
python3 -m build ....
```

let's change them all to `python3 ...` style

pyproject.toml
pjonsson: That makes the repository unusable with uv since the version constraints aren't solvable, but I've updated the pull request.

soxofaan: how do you mean unusable? 
the dependencies won't be solvable if you install the "localprocessing" extra in a python 3.8 env, indeed, but that's the point of what I propose.

pjonsson:
```
$ uv sync
warning: Ignoring existing virtual environment linked to non-existent Python interpreter: .venv/bin/python3 -> python
Using CPython 3.14.4 interpreter at: /usr/bin/python3
Removed virtual environment at: .venv
Creating virtual environment at: .venv
  × No solution found when resolving dependencies for split (markers:
  │ python_full_version >= '3.8.1' and python_full_version < '3.9'):
  ╰─▶ Because the requested Python version (>=3.8) does not satisfy
      Python>=3.9 and sphinx-autodoc-typehints>=2.2.3,<=2.3.0
      depends on Python>=3.9, we can conclude that
      sphinx-autodoc-typehints>=2.2.3,<=2.3.0 cannot be used.
      And because only the following versions of sphinx-autodoc-typehints
      are available:
          sphinx-autodoc-typehints<=2.2.3
          sphinx-autodoc-typehints==2.3.0
          sphinx-autodoc-typehints==2.4.0
          sphinx-autodoc-typehints==2.4.1
          sphinx-autodoc-typehints==2.4.2
          sphinx-autodoc-typehints==2.4.3
          sphinx-autodoc-typehints==2.4.4
          sphinx-autodoc-typehints==2.5.0
          sphinx-autodoc-typehints==3.0.0
          sphinx-autodoc-typehints==3.0.1
          sphinx-autodoc-typehints==3.1.0
          sphinx-autodoc-typehints==3.2.0
          sphinx-autodoc-typehints==3.3.0
          sphinx-autodoc-typehints==3.4.0
          sphinx-autodoc-typehints==3.5.0
          sphinx-autodoc-typehints==3.5.1
          sphinx-autodoc-typehints==3.5.2
          sphinx-autodoc-typehints==3.6.0
          sphinx-autodoc-typehints==3.6.1
          sphinx-autodoc-typehints==3.6.2
          sphinx-autodoc-typehints==3.6.3
          sphinx-autodoc-typehints==3.7.0
          sphinx-autodoc-typehints==3.8.0
          sphinx-autodoc-typehints==3.9.0
          sphinx-autodoc-typehints==3.9.1
          sphinx-autodoc-typehints==3.9.2
          sphinx-autodoc-typehints==3.9.3
          sphinx-autodoc-typehints==3.9.4
          sphinx-autodoc-typehints==3.9.5
          sphinx-autodoc-typehints==3.9.6
          sphinx-autodoc-typehints==3.9.7
          sphinx-autodoc-typehints==3.9.8
          sphinx-autodoc-typehints==3.9.9
          sphinx-autodoc-typehints==3.9.10
          sphinx-autodoc-typehints==3.9.11
          sphinx-autodoc-typehints==3.10.0
          sphinx-autodoc-typehints==3.10.1
          sphinx-autodoc-typehints==3.10.2
          sphinx-autodoc-typehints==3.10.3
          sphinx-autodoc-typehints==3.10.4
          sphinx-autodoc-typehints==3.10.5
          sphinx-autodoc-typehints==3.10.6
          sphinx-autodoc-typehints==3.11.0
          sphinx-autodoc-typehints==3.11.1
          sphinx-autodoc-typehints==3.12.0
          sphinx-autodoc-typehints==3.12.1
          sphinx-autodoc-typehints==3.13.0
          sphinx-autodoc-typehints==3.13.1
          sphinx-autodoc-typehints==3.13.2
          sphinx-autodoc-typehints==3.13.3
          sphinx-autodoc-typehints==3.13.4
          sphinx-autodoc-typehints==3.13.5
      we can conclude that sphinx-autodoc-typehints>=2.2.3,<2.4.0 cannot be
      used. (1)

      Because the requested Python version (>=3.8) does not satisfy
      Python>=3.10 and sphinx-autodoc-typehints>=2.4.0,<=3.0.1
      depends on Python>=3.10, we can conclude that
      sphinx-autodoc-typehints>=2.4.0,<=3.0.1 cannot be used.
      And because we know from (1) that sphinx-autodoc-typehints>=2.2.3,<2.4.0
      cannot be used, we can conclude that
      sphinx-autodoc-typehints>=2.2.3,<3.1.0 cannot be used. (2)

      Because the requested Python version (>=3.8) does not satisfy
      Python>=3.10 and sphinx-autodoc-typehints>=3.1.0,<=3.6.1
      depends on Python>=3.11, we can conclude that
      sphinx-autodoc-typehints>=3.1.0,<=3.6.1 cannot be used.
      And because we know from (2) that sphinx-autodoc-typehints>=2.2.3,<3.1.0
      cannot be used, we can conclude that
      sphinx-autodoc-typehints>=2.2.3,<3.6.2 cannot be used. (3)

      Because the requested Python version (>=3.8) does not satisfy
      Python>=3.10 and sphinx-autodoc-typehints>=3.6.2 depends on
      Python>=3.12, we can conclude that sphinx-autodoc-typehints>=3.6.2
      cannot be used.
      And because we know from (3) that sphinx-autodoc-typehints>=2.2.3,<3.6.2
      cannot be used, we can conclude that sphinx-autodoc-typehints>=2.2.3
      cannot be used.
      And because openeo[dev] depends on sphinx-autodoc-typehints>=2.2.3 and
      your project requires openeo[dev], we can conclude that your project's
      requirements are unsatisfiable.

hint: While the active Python version is 3.14, the resolution failed for other Python versions supported by your project. Consider limiting your project's supported Python versions using `requires-python`.
hint: The `requires-python` value (>=3.8) includes Python versions that are not supported by your dependencies (e.g., sphinx-autodoc-typehints>=2.2.3,<=2.3.0 only supports >=3.9). Consider using a more restrictive `requires-python` value (like >=3.9).
```

Edit: I realize this is what you want, I just think the cost of making uv unusable for package development outweighs the benefit of supporting Python 3.8 for the documentation and localprocessing optionals.

soxofaan: Also note that the intended unsolvability is already the case with the current (setup.py based) releases. 

For example, with the `docs` extra (which is lighter to experiment with):

installing the `docs` extra with uv doesn't work on python 3.8:
```
❯ uv run --no-project --python=3.8 --with 'openeo[docs]==0.52.0' python -m pip freeze
  × No solution found when resolving `--with` dependencies:
  ╰─▶ Because the current Python version (3.8.20) does not satisfy Python>=3.9 and sphinx-autodoc-typehints>=2.2.3,<=2.3.0 depends on Python>=3.9, we can conclude that
      sphinx-autodoc-typehints>=2.2.3,<=2.3.0 cannot be used.
....
```

but it works on python 3.9
```
❯ uv run --no-project --python=3.9 --with 'openeo[docs]==0.52.0' python -m pip freeze
accessible-pygments==0.0.5
alabaster==0.7.16
babel==2.18.0
...
```

soxofaan: (FYI: I am working on the train and apparently my view of this thread was bit out of sync)

I see now you are talking about locking and syncing with uv (I assumed it was just about `uv pip install`), 
but then it is important to describe your project environment a bit

soxofaan: I can reproduce your problem with a temp project and a pyproject.toml file that has
```
requires-python = ">=3.8"
dependencies = [
    "openeo[docs]==0.52.0",
]
```

Which is indeed not solvable. But I don't think it's the responsibility of the openeo package to "fix" that e.g. by declaring misleading constraints leading to a broken environment

I can fix the  `uv sync`  in my  temp project by doing
```
requires-python = ">=3.8"
dependencies = [
    "openeo==0.52.0",
    "openeo[docs]==0.52.0; python_version>='3.9'",
]
```

pjonsson: It's this package that can't be developed with uv (even when I have Python 3.14 in my path) with the non-resolvable version constraints:
```
ubuntu@ea784ccf698b:~/openeo-python-client$ uv run pytest
  × No solution found when resolving dependencies for split (markers:
  │ python_full_version >= '3.8.1' and python_full_version < '3.9'):
  ╰─▶ Because the requested Python version (>=3.8) does not satisfy
      Python>=3.9 and sphinx-autodoc-typehints>=2.2.3,<=2.3.0
...
```
But I have updated the PR according to your comments, so let's get this merged, a 10 line local diff is still easier to manage than a 200+ line diff when working on this package.

soxofaan: did some ping-pong with Claude AI about this and got this suggestion to add 
```
[tool.uv]
environments = ["python_version >= '3.10'"]
```
(version 3.9 might also work)

which decouples the scope of the uv locking from the generic `requires-python = ">=3.8"` constraint

Seems like that will be the escape hatch  here

pjonsson: It works if I change that to `python_full_version`, so I've updated the PR with that, thank you!

soxofaan: `python_version` worked for me if after removing existing `.venv` and `uv.lock`

pjonsson: I had removed those as well before I tested with uv 0.12.1, what version of uv did you test with?

(I can update to a more recent version of uv if that is required for this to work, they just release so frequently that I don't have time to keep up with all their updates.)

pyproject.toml
```diff
@@ -1,7 +1,147 @@
...
+    "pyproj",
+    "rioxarray>=0.13.0",
+]
+# Install oschmod even when platform is not Windows, e.g. for testing in CI.
+oschmod = [
+    "oschmod>=0.3.12",
+]
+tests = [
+    "boto3",
+    "boto3~=1.37.38; python_version<'3.9'",
```
soxofaan: this entry is a temp hack to solve extreme slow dependency resolution on python 3.8,
added by https://github.com/Open-EO/openeo-python-client/commit/b8f5403b28068e76573f1977c277626242f75769
but the related comment is lost because of alphabetic reordering here.
So please mark it (so it can be cleaned up when dropping python 3.8 support):

```suggestion
    "boto3~=1.37.38; python_version<'3.9'",  # Pin to speed up slow dependency resolution on Python 3.8
```

pyproject.toml
```diff
@@ -1,7 +1,147 @@
...
+    "botocore",
+    "botocore~=1.37.38; python_version<'3.9'",
```
soxofaan: same:

```suggestion
    "botocore~=1.37.38; python_version<'3.9'",  # Pin to speed up slow dependency resolution on Python 3.8
```

pyproject.toml
```diff
@@ -1,7 +1,147 @@
...
+    "dirty_equals>=0.8.0",
+    "flake8>=5.0.0",
+    "httpretty>=1.1.4",
+    "mock",
+    "moto>=5.0.0",
+    # Some pins to speed up slow dependency resolution in Python 3.8 venvs
+    "moto~=5.0.28; python_version<'3.9'",
```
soxofaan:
```suggestion
    "moto~=5.0.28; python_version<'3.9'",  # Pin to speed up slow dependency resolution on Python 3.8
```

pyproject.toml
```diff
@@ -1,7 +1,147 @@
...
+]
+
+[project.optional-dependencies]
+artifacts = [
+    "boto3",
+    "botocore",
+]
+dev = [
+    "boto3",
+    "boto3~=1.37.38; python_version<'3.9'",
```
soxofaan:
```suggestion
    "boto3~=1.37.38; python_version<'3.9'",  # Pin to speed up slow dependency resolution on Python 3.8
```

pyproject.toml
```diff
@@ -1,7 +1,147 @@
...
+    "botocore",
+    "botocore~=1.37.38; python_version<'3.9'",
```
soxofaan:
```suggestion
    "botocore~=1.37.38; python_version<'3.9'",  # Pin to speed up slow dependency resolution on Python 3.8
```

pyproject.toml
```diff
@@ -1,7 +1,147 @@
...
+    "dirty_equals>=0.8.0",
+    "flake8>=5.0.0",
+    "httpretty>=1.1.4",
+    "mock",
+    "moto>=5.0.0",
+    "moto~=5.0.28; python_version<'3.9'",
```
soxofaan:
```suggestion
    "moto~=5.0.28; python_version<'3.9'",  # Pin to speed up slow dependency resolution on Python 3.8
```

pyproject.toml
```diff
@@ -1,7 +1,147 @@
...
+    "types-boto3-sts",
+    "urllib3<2.3.0",
+]
+docs = [
+    "myst-parser",
+    "pydata_sphinx_theme",
+    "sphinx",
+    "sphinx-autodoc-annotation",
+    "sphinx-autodoc-typehints >=2.2.3; python_version>='3.10'",
+    "sphinx-autodoc-typehints; python_version<='3.9'",
```
soxofaan: as discussed elsewhere in this PR, please revert to
```suggestion
    "sphinx-autodoc-typehints>=2.2.3",
```

pjonsson: I'm fairly sure I had removed this, I wonder if it snuck in again and I accidentally committed it when testing with uv.

It's gone again.

pyproject.toml
soxofaan: I'm now at 0.12.10 (upgraded yesterday to be sure I could properly reproduce the problems you reported).

but after downgrading to 0.12.1, `environments = ["python_version >= '3.10'"]` still seems to work for me. How does  it fail for you?

pyproject.toml
```diff
@@ -1,7 +1,146 @@
...
+    "types-boto3-s3",
+    "types-boto3-sts",
+    "urllib3<2.3.0",
+]
+docs = [
+    "myst-parser",
+    "pydata_sphinx_theme",
+    "sphinx",
+    "sphinx-autodoc-annotation",
+    "sphinx-autodoc-typehints >=2.2.3; python_version>='3.10'",
```
soxofaan:
```suggestion
    "sphinx-autodoc-typehints >=2.2.3",
```

pyproject.toml
pjonsson: It failed with the same message about not being able to solve the constraints, and the error message using `python_full_version` rather than `python_version`.

For some reason I can no longer reproduce the failure in my container. I haven't changed anything in my setup, but I got a Docker version update yesterday which forced me to start my container again, so I'm wondering if the failure was somehow related to the uv cache hanging on to something.

I just pushed the update to this PR that uses `python_version` instead.

.github/workflows/build-wheel.yml
```diff
@@ -19,10 +19,10 @@ jobs:
         python-version: '3.11'
     - name: Install dependencies
       run: |
-        python -m pip install --upgrade pip
-        pip install wheel
-    - name: Build bdist_wheel
-      run: python setup.py bdist_wheel
+        python3 -m pip install --upgrade pip
+        python3 -m pip install build wheel
```
soxofaan: Claude suggests `wheel` is redundant here
```suggestion
        python3 -m pip install build
```

pyproject.toml
```diff
@@ -1,7 +1,146 @@
+[build-system]
+requires = ["setuptools>=69"]
```
soxofaan: any particular reason you picked `setuptools>=69` here?

current https://packaging.python.org/en/latest/tutorials/packaging-projects/#choosing-a-build-backend suggests `setuptools >= 77.0.3`

Claude mentions something about future proofing the license entry related to this version:
> `license = { text = "Apache-2.0" }` with `setuptools>=69`: fine now; `setuptools ≥77` will nudge toward the SPDX string form `license = "Apache-2.0"`.

soxofaan (commented):
a couple of more notes after asking Claude for a review too

pyproject.toml
pjonsson: It's version 69 that started including `py.typed`, so I know I want at least that version.

I don't think requiring even the most recent version of setuptools would cause problems for me personally, so I'm happy to bump it to whatever version `>= 69` that you want.

pjonsson: Bumping to `>= 77` caused the 3.8 job to fail: https://github.com/Open-EO/openeo-python-client/actions/runs/34452633595/job/102791734321?pr=908

Lowering to `>= 75` makes the 3.8 job fail with the new license specifier: https://github.com/Open-EO/openeo-python-client/actions/runs/34452962901/job/102792794422?pr=908

soxofaan:
merged!
thanks for all the effort

PR #912, opened 2026-07-13
Preserve metadata when merging cubes without bands

### LINKED ISSUE #783: merge_cubes: merge metadata even without band_dimensions
related code:
https://github.com/Open-EO/openeo-python-client/blob/77c3e87e44a199949306d230339a92863c96de40/openeo/rest/datacube.py#L2210-L2224

Currently if one or both of the cubes have missing band data the resulting metadata is set to None.
Suggested approaches:
1. cube1 has bands and cube2 has bands -> current implementation
2. cube1 has bands but cube2 has no bands -> take over bands of cube1 (cube2 will be merged with overlap on all bands)
3. cube1 has no bands but cube 2 has bands -> See step 2 I think
4. both have no bands -> use metadata from first cube

If one of both cubes has none as metadata I think it is still safer to put the result to none as well

In a better scenario the extents of all dimensions are merged but that might be the scope of a separate issue rather than include it in this one.

Silvren:
## Summary

- preserve cube metadata when `merge_cubes` receives cubes without band dimensions
- retain the available band metadata when only one input cube has a band dimension
- keep the existing band-union behavior when both cubes have band dimensions
- keep metadata unknown when either input has no metadata

## Testing

- `python -m pytest tests/rest/datacube/test_datacube100.py -q` (347 passed, 2 skipped)
- `pre-commit run --files CHANGELOG.md openeo/rest/datacube.py tests/rest/datacube/test_datacube100.py`

Fixes #783

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Better error message when passing a bounding box with field typos in `Connection.load_collection()` and alike  ([#910](https://github.com/Open-EO/openeo-python-client/issues/910))
-
+- Preserve available band metadata when merging data cubes if one cube has no band dimension. ([#783](https://github.com/Open-EO/openeo-python-client/issues/783))
 
 ## [0.50.0] - 2026-05-18
 
--- a/openeo/rest/datacube.py
+++ b/openeo/rest/datacube.py
@@ -2183,18 +2183,16 @@ class DataCube(_ProcessGraphAbstraction):
         arguments = {"cube1": self, "cube2": other}
         if overlap_resolver:
             arguments["overlap_resolver"] = build_child_callback(overlap_resolver, parent_parameters=["x", "y"])
-        if (
-            self.metadata
-            and self.metadata.has_band_dimension()
-            and isinstance(other, DataCube)
-            and other.metadata
-            and other.metadata.has_band_dimension()
-        ):
-            # Minimal client side metadata merging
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -22,7 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Better error message when passing a bounding box with field typos in `Connection.load_collection()` and alike  ([#910](https://github.com/Open-EO/openeo-python-client/issues/910))
-
+- Preserve available band metadata when merging data cubes if one cube has no band dimension. ([#783](https://github.com/Open-EO/openeo-python-client/issues/783))
```
soxofaan, line 25: changelog has been bumped in meantime, this has to be moved

openeo/rest/datacube.py
```diff
@@ -2183,18 +2183,16 @@ def merge_cubes(
...
             merged_metadata = self.metadata
-            for b in other.metadata.band_dimension.bands:
-                if b not in merged_metadata.bands:
-                    merged_metadata = merged_metadata.append_band(b)
+            if self.metadata.has_band_dimension() and other.metadata.has_band_dimension():
+                for b in other.metadata.band_dimension.bands:
+                    if b not in merged_metadata.bands:
+                        merged_metadata = merged_metadata.append_band(b)
+            elif other.metadata.has_band_dimension():
+                merged_metadata = other.metadata
```
soxofaan, line 2195: It's not ideal to use the other.metadata in full here, just to copy the band metadata

PR #913, opened 2026-07-13
Compare GeoJSON assets in job result tests

### LINKED ISSUE #862: support vector cube comparison in openeo.testing.results
https://github.com/Open-EO/openeo-python-client/blob/bfcd173f03c9be3d02eaced8dbc57544083161d2/openeo/testing/results.py#L450

support vector cubes here as well

e.g. from https://github.com/ESA-APEx/apex_algorithms/actions/runs/21517683654/job/62000109521:
```
WARNING  openeo.testing.results:results.py:450 Unhandled job result asset 'vectorcube.geojson'
```

Silvren:
## Summary

- recognize `.geojson` assets in `assert_job_results_allclose` instead of treating them as unhandled files
- compare feature attributes with the existing relative and absolute tolerances
- compare geometry types, geometry values, feature order, and coordinate reference systems through GeoPandas
- report attribute and geometry mismatches separately for actionable failures

## Testing

- `python -m pytest tests/testing -q` (84 passed)
- `pre-commit run --files CHANGELOG.md openeo/testing/results.py tests/testing/test_results.py`

Fixes #862

### DIFF
--- a/openeo/testing/results.py
+++ b/openeo/testing/results.py
@@ -10,6 +10,7 @@ from pathlib import Path
 from typing import List, Optional, Union
 
 import numpy
+import pandas
 import xarray
 import xarray.testing
 from xarray import DataArray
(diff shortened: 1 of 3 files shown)


Silvren:
Thanks for the notification. I tried to inspect the Jenkins console, but `jenkins.vgt.vito.be` currently returns NXDOMAIN through both my system DNS and public DNS, so the console is not reachable from my network.

I reproduced the PR branch locally:
- `python -m pytest tests/testing/test_results.py -q`: 56 passed
- `python -m pytest tests/testing -q`: 84 passed
- `git diff --check`: passed

GitHub also shows two workflows awaiting maintainer approval. Could you please approve/re-run the workflows and retry the Jenkins build? If Jenkins still fails, sharing the first relevant traceback from the console would help me address it. Thanks!

openeo/testing/results.py
```diff
@@ -95,6 +96,55 @@ def _as_xarray_dataarray(data: Union[str, Path, xarray.DataArray]) -> xarray.Dat
     return data
 
 
+def _load_geodataframe(path: Union[str, Path]):
+    """Load a vector data file as a GeoPandas GeoDataFrame."""
+    try:
+        import geopandas
+    except ImportError as e:
```
soxofaan, line 103: why is this? geopandas is a base requirement, not an extra, 
no need for this guard

openeo/testing/results.py
```diff
@@ -95,6 +96,55 @@ def _as_xarray_dataarray(data: Union[str, Path, xarray.DataArray]) -> xarray.Dat
...
+    actual: Union[str, Path],
+    expected: Union[str, Path],
+    *,
+    rtol: float = _DEFAULT_RTOL,
+    atol: float = _DEFAULT_ATOL,
+) -> List[str]:
+    """Compare vector geometries and their attribute data."""
+    actual_gdf = _load_geodataframe(actual)
+    expected_gdf = _load_geodataframe(expected)
+    import geopandas.testing
```
soxofaan, line 118: avoid local import when not necessary

openeo/testing/results.py
```diff
@@ -459,6 +509,11 @@ def _compare_job_results(
             if issues:
                 all_issues.append(f"Issues for file {filename!r}:")
                 all_issues.extend(issues)
+        elif expected_path.suffix.lower() == ".geojson":
```
soxofaan, line 512: I'm not sure having ".geojson" is enough as an indicator for being loadable as geodataframe.
e.g. will a GeoJSON file with just a Polygon load here?
I think only Features and FeatureCollection objects can be loaded

PR #921, opened 2026-07-15
Add py.typed

pjonsson:
There are type signatures in this
package, so include a py.typed.


soxofaan:
can you add a changelog entry about this too?

pjonsson:
@soxofaan I've added a changelog entry and rebased on top of latest master.

PR #924, opened 2026-07-21
CI: pin actions by sha256

pjonsson:
Update the actions to the latest
version and pin them by sha256
to avoid supply-chain attacks
if the action provider gets
compromised.


soxofaan:
already handled as part of https://github.com/Open-EO/openeo-python-client/issues/927

PR #925, opened 2026-07-21
test_warnings.py: fix SyntaxWarnings

pjonsson:
Add the missing r-qualifier on the strings
to fix the two warnings about invalid
escape sequences.


soxofaan:
thanks!

Issue #928, opened 2026-08-10
Support document based "derived_from" links in openeo.testing.results


soxofaan:
openeo.testing.results has support for extracting "derived_from" links, but only the legacy style where the "href" field contains just an item id or local path (instead of being a real href), e.g.

```
"links": [
    {
        "href": "c_gls_NDVI300_202509210000_GLOBE_OLCI_V3.0.1_cog",
        "rel": "derived_from",
        "title": "Derived from c_gls_NDVI300_202509210000_GLOBE_OLCI_V3.0.1_cog",
        "type": "application/json"
    },
```

Support has to be added for "derived_from" links that point to a STAC ItemCollection, as added in https://github.com/Open-EO/openeo-geopyspark-driver/issues/1488 and https://github.com/Open-EO/openeo-geopyspark-driver/issues/1618 e.g.

```
"links": [
    {
        "href": "https://openeo.example.com/.../stac-item-collection-loadcollection1.json?signature=...",
        "rel": "derived_from",
        "type": "application/geo+json"
    },
```

soxofaan:
implemented by https://github.com/Open-EO/openeo-python-client/pull/929

PR #929, opened 2026-08-10
Support document based "derived_from" links in `openeo.testing.results`

soxofaan:
ref #928


openeo/testing/results.py
```diff
@@ -465,14 +467,76 @@ def _compare_job_results(
...
+        self._request_session: requests.Session = session_with_retries(retry)
+
+    def from_links(self, links: Iterable[dict]) -> Iterable[_ProductRef]:
+        """
+        Extract product references from a list of links (e.g. as part of result metadata).
+        """
+        for link in links:
+            if link["rel"] == "derived_from":
+                link_href = link["href"]
+                if self._HREF_HTTP_SCHEME_REGEX.match(link_href):
```
soxofaan: Also support relative paths  (e.g. when reference data is local file on disk)

PR #932, opened 2026-08-11
Fix: don't drop requested OIDC scopes not listed in scopes_supported 

### LINKED ISSUE #930: Cannot authenticate against a Microsoft Entra ID protected backend: Entra never advertises custom scopes in `scopes_supported`
## The problem

Microsoft Entra ID cannot advertise an application's custom API scopes in its OIDC discovery document. Its `scopes_supported` is tenant-wide and permanently fixed at `["openid", "profile", "email", "offline_access"]`. This is by design, per Microsoft in [AzureAD/microsoft-identity-web#1689](https://github.com/AzureAD/microsoft-identity-web/discussions/1689):

> The OIDC metadata endpoint just returns Azure AD's STS metadata. There's no concept of custom resources or custom scopes here.

`OidcProviderInfo.__init__` intersects the scopes a backend declares at `/credentials/oidc` against that list (`openeo/rest/auth/oidc.py`, 0.48.0):

```python
self._scopes = {"openid"}.union(scopes or []).intersection(self._supported_scopes)
```

A sensible default against a well-behaved provider but with Entra it drops the one scope that matters:

```python
provider = OidcProviderInfo(
    issuer=f"https://login.microsoftonline.com/{TENANT}/v2.0",
    scopes=["openid", "profile", "email", f"api://{CLIENT_ID}/openeo"],
)
provider.get_scopes_string()
# -> 'email openid profile'      the api:// scope never reaches Entra
```

With no resource scope requested, Entra issues an access token unusable and the backend (here titiler-openeo) rejects the token, and the user sees a signature error with no obvious connection to a dropped scope.

The net effect is that **no openEO backend protected by Entra can be authenticated against with this client**, and nothing the backend or the app registration can configure will change that.

## Possible accommodations

Any of these would be enough:

1. **Warn instead of dropping** keep the declared scopes, `log.warning` when one is not advertised. Preserves the diagnostic value without the breakage.
2. **Opt out** e.g. `OidcProviderInfo(..., strict_scopes=False)`, plumbed through `Connection.authenticate_oidc*`.
3. **Do not intersect at all** [RFC 8414](https://www.rfc-editor.org/rfc/rfc8414#section-2) makes `scopes_supported` RECOMMENDED and descriptive, and RFC 6749 §3.3 puts the grant decision on the authorization server, which reports the result in the `scope` response parameter.

Options 1 and 3 also help any other provider with incomplete metadata; option 2 leaves each user to discover the problem first.

## Workaround

```python
from openeo.rest.auth.oidc import OidcProviderInfo

_orig_init = OidcProviderInfo.__init__

def _patched_init(self, issuer=None, discovery_url=None, scopes=None, **kwargs):
    _orig_init(self, issuer=issuer, discovery_url=discovery_url, scopes=scopes, **kwargs)
    self._scopes = {"openid"} | set(scopes or [])

OidcProviderInfo.__init__ = _patched_init
```

## Environment

`openeo` 0.48.0 · Python 3.13 · Entra ID v2.0 endpoint, single tenant · backend: [titiler-openeo](https://github.com/sentinel-hub/titiler-openeo)

emmanuelmathot:
## What / why

Fixes #930.

`OidcProviderInfo.__init__` was intersecting the scopes an openEO backend declares (e.g. `api://<client-id>/openeo`) against the discovery document's `scopes_supported` field. Microsoft Entra ID always reports a fixed, tenant-wide `scopes_supported` (`openid`, `profile`, `email`, `offline_access`) regardless of the app's actual custom scopes, so the intersection silently dropped the one scope needed to get a usable access token. No openEO backend behind Entra ID could be authenticated against.

Per [@soxofaan comment](https://github.com/Open-EO/openeo-python-client/issues/930#issuecomment-5252379168), this implements **option 3** from the issue: stop filtering requested scopes against `scopes_supported` altogether. It's only a RECOMMENDED, descriptive field per RFC 8414 §2

## Changes

- `openeo/rest/auth/oidc.py`: `OidcProviderInfo` now requests `{"openid"} | scopes` directly, without intersecting against `scopes_supported`. The `offline_access`-if-supported behavior for refresh tokens is unchanged.
- `tests/rest/auth/test_oidc.py`: added a regression test asserting a requested scope not present in `scopes_supported` is preserved.
- `CHANGELOG.md`: entry under `Unreleased / Fixed`.

## Disclosure

This PR was drafted with the help of Claude Code, per `AGENTS.md`.

### DIFF
--- a/openeo/rest/auth/oidc.py
+++ b/openeo/rest/auth/oidc.py
@@ -279,9 +279,11 @@ class OidcProviderInfo:
         except Exception as e:
             raise OidcException(f"Failed to obtain OIDC discovery document from {self.discovery_url!r}: {e!r}") from e
         self.issuer = issuer or self.config["issuer"]
-        # Minimal set of scopes to request
         self._supported_scopes = self.config.get("scopes_supported", ["openid"])
-        self._scopes = {"openid"}.union(scopes or []).intersection(self._supported_scopes)
+        # Don't filter requested scopes against `scopes_supported`: it's only a RECOMMENDED
+        # discovery field (RFC 8414 section 2), and some providers (e.g. Microsoft Entra ID)
+        # report a fixed, incomplete list there regardless of which scopes they actually accept.
+        self._scopes = {"openid"}.union(scopes or [])
         log.debug(f"Scopes: provider supported {self._supported_scopes} & backend desired {scopes} -> {self._scopes}")
         self.default_clients = default_clients
         self.authorization_parameters = authorization_parameters or {}
--- a/openeo/rest/auth/oidc.py
+++ b/openeo/rest/auth/oidc.py
@@ -279,12 +279,12 @@ class OidcProviderInfo:
         except Exception as e:
             raise OidcException(f"Failed to obtain OIDC discovery document from {self.discovery_url!r}: {e!r}") from e
         self.issuer = issuer or self.config["issuer"]
-        self._supported_scopes = self.config.get("scopes_supported", ["openid"])
-        # Don't filter requested scopes against `scopes_supported`: it's only a RECOMMENDED
-        # discovery field (RFC 8414 section 2), and some providers (e.g. Microsoft Entra ID)
-        # report a fixed, incomplete list there regardless of which scopes they actually accept.
+        # Note: we don't filter requested scopes against the discovery document's
+        # `scopes_supported`: it's only a RECOMMENDED discovery field (RFC 8414 section 2),
+        # and some providers (e.g. Microsoft Entra ID) report a fixed, incomplete list there
+        # regardless of which scopes they actually accept.
         self._scopes = {"openid"}.union(scopes or [])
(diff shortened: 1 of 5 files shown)


openeo/rest/auth/oidc.py
```diff
@@ -279,9 +279,11 @@ def __init__(
         except Exception as e:
             raise OidcException(f"Failed to obtain OIDC discovery document from {self.discovery_url!r}: {e!r}") from e
         self.issuer = issuer or self.config["issuer"]
-        # Minimal set of scopes to request
         self._supported_scopes = self.config.get("scopes_supported", ["openid"])
-        self._scopes = {"openid"}.union(scopes or []).intersection(self._supported_scopes)
+        # Don't filter requested scopes against `scopes_supported`: it's only a RECOMMENDED
```
soxofaan: I wonder if it is even necessary to keep `self._supported_scopes` at all if it's only for discovery.
It was originally added defensively to avoid doing invalid requests, but that's apparently unnecessarily cautious and counter-productive.

Removing `_supported_scopes` would reduce the dependency on the config document

Issue #747, opened 2025-03-10
robust/ranged download support


soxofaan:
EO data downloads can be pretty big, and big transfers can be brittle in some situations.
If the server supports ranged downloads, it's possible to do this more robustly.

soxofaan:
inspiration:  https://gist.github.com/pvbouwel/f7d23dbe8e7207f1b7ede8e7f924d868  by @pvbouwel

mbuchhorn:
we solved that in the WEED project for cases were we have to download the data locally in this way: 
openEO pipeline always dump the files into a S3 bucket and then in the jobmanager we have a threaded download to get the files from S3 and after successful download (etag check) to delete the file on S3. This is even way faster then directly downloading from openEO.
example: 4GB produced file with direct download needs roughly 20 - 25 min for the download to local disk after the file is successful processed. With our setup the download is roughly 3-4 minutes.
https://github.com/ESA-WEED-project/eo_processing/blob/main/src/eo_processing/utils/jobmanager.py#L260

if you have questions to the `storage` object we are using then just ask :)

jdries:
Pointer to where this probably needs to be added:
https://github.com/Open-EO/openeo-python-client/blob/9b92d6f59566ec2d32deaaddae261e683bdd9861/openeo/rest/job.py#L405

PR #933, opened 2026-08-11
Introduce job result downloading with item walking and ref rewriting

### LINKED ISSUE #931: "deep" download functionality of job results
`job.get_results().download_files()` is the existing way to download job result assets
but lacks some features:
- it only supports collection level assets, not item level assets. While openEO 1.1 requires both to be the same, collection-level assets is sub-optimal (e.g. naming-wise) and is scheduled to be removed in openEO 2.0. Item level assets also eliminate ambiguity about which asset belongs to which item
- download of item STAC metadata
- rewrite hrefs to local paths (to make metadata self-contained)
- download additional linked documents like derived_from documents

soxofaan:
ref #931


openeo/rest/job.py
```diff
@@ -722,6 +671,122 @@ def download_files(
 
         return downloaded
 
+    def download_collection(
```
soxofaan: "download_collection" might look confusing

rename to "download_as_collection"?

openeo/rest/job.py
```diff
@@ -722,6 +671,122 @@ def download_files(
...
+        self,
+        target: Union[Path, str] = None,
+        *,
+        rewrite_references: bool = True,
+    ) -> List[Path]:
```
soxofaan, line 687: expose other args:
- extra_link_rels_to_download
- download_collection_assets
- json style options

openeo/rest/job.py
```diff
@@ -722,6 +671,122 @@ def download_files(
...
+
+
+class _JobResultDownloader:
+    """
+    Helper class to download batch job results as a STAC collection,
+    recursively walking through items, assets and additional linked metadata.
+    """
+
+    # TODO: make this a public API that users can implement for custom download behavior (e.g. download to S3, ...)
+    # TODO: API to warn about or skip existing/previously downloaded files?
```
soxofaan, line 761: TODO: API to skip download failures (after retries) and just warn?

openeo/rest/job.py
```diff
@@ -722,6 +671,122 @@ def download_files(
...
+        """
+        Download the job results as a self-contained STAC collection.
+        """
+        result_metadata = self._connection.get(self._job.get_results_metadata_url(), expected_status=200).json()
+        if result_metadata.get("type") != "Collection":
+            raise OpenEoClientException(f"Result metadata is not a STAC Collection, but {result_metadata.get('type')}")
+
+        result_metadata_path = self._root_path / DEFAULT_JOB_RESULTS_FILENAME
+        # Initial write of metadata, will possibly be updated later if rewrite_references is True
+        self._json_dump(data=result_metadata, path=result_metadata_path)
```
soxofaan: give initial file a ".inprogress" suffix?
and only swap to final filename at the end

openeo/rest/job.py
```diff
@@ -722,6 +671,122 @@ def download_files(
...
+    # TODO: API to warn about or skip existing/previously downloaded files?
+
+    def __init__(self, *, job: BatchJob, target: Union[Path, str, None] = None, rewrite_references: bool = True):
+        # TODO: dedicated request session (with appropriate retry strategy) for downloading?
+        # TODO: customization of JSON format (compact vs "pretty" vs ...)?
+        self._job = job
+        self._connection = job.connection
+        self._root_path = Path(target or Path.cwd() / job.job_id)
+        self._rewrite_references = rewrite_references
+        self._downloaded = []
```
soxofaan: Use this list of downloaded files to avoid overwrites

openeo/rest/job.py
```diff
@@ -722,6 +671,122 @@ def download_files(
...
+
+    def _json_dump(self, data: dict, path: Union[str, Path]) -> Path:
```
soxofaan: make this a more general utility

tests/rest/test_job.py
```diff
@@ -707,6 +716,62 @@ def setup(*, job_id="jj1", assets: dict, media_type: str = "image/tiff; applicat
...
+        item_href = f"{api_url}/items/{item_id}.json"
+        item_doc = StacDummyBuilder.item(
+            id=item_id,
+            stac_version="1.1.0",
+            assets=assets,
+        )
+        requests_mock.get(item_href, json=item_doc)
+        return item_href
+
+    def setup(*, job_id="job-123", items: dict, api_url: str = API_URL):
```
soxofaan: make this a class instead of closured functions

openeo/rest/job.py
soxofaan: let's keep it internal to allow deeper customization

openeo/rest/_connection.py
```diff
@@ -262,3 +296,60 @@ def put(self, path: str, headers: Optional[dict] = None, data: Optional[dict] =
...
+                    try:
+                        range_headers = {"Range": f"bytes={from_byte_index}-{to_byte_index}"}
+                        with self.get(path=url, headers=range_headers, stream=True) as r:
+                            r.raise_for_status()
+                            for block in r.iter_content(chunk_size=chunk_size):
+                                f.write(block)
+                        break
+                    except OpenEoApiPlainError as error:
+                        tries_left -= 1
+                        if tries_left > 0 and error.http_status_code in RETRIABLE_DOWNLOAD_STATUSCODES:
```
pjonsson, line 349: There's no exponential backoff with retries here from what I can see, it will just hammer the server again and again if it fails.

I don't have the code for openeo-python-client in front of me, but I would suggest using something like https://requests.readthedocs.io/en/latest/user/advanced/#example-automatic-retries instead to get a solid implementation of all the routine-aspects of retries (the HTTP codes to retry are configurable). As a bonus, that also respects the `Retry-After` sent by the server.

openeo/rest/_connection.py
```diff
@@ -262,3 +296,60 @@ def put(self, path: str, headers: Optional[dict] = None, data: Optional[dict] =
...
+        head = self.head(url, stream=True)
+        if head.ok and head.headers.get("Accept-Ranges") == "bytes" and "Content-Length" in head.headers:
+            file_size = int(head.headers["Content-Length"])
+            self._download_ranged(
+                url=url, target=target, file_size=file_size, chunk_size=chunk_size, range_size=range_size
+            )
+        else:
+            self._download_all_at_once(url=url, target=target, chunk_size=chunk_size)
+
+    def _download_all_at_once(self, url: str, target: Path, *, chunk_size: int = DEFAULT_DOWNLOAD_CHUNK_SIZE) -> None:
```
pjonsson, line 317: All callers for these internal methods pass all parameters, are the default values for the arguments useful?

openeo/rest/_connection.py
soxofaan: indeed, good point,
however, that's existing code I just moved in this PR to make it more generally available (originally introduced for #747).

I'd want to avoid too much scope creep as this PR is already way larger than initially intended, so let's move this to another ticket -> https://github.com/Open-EO/openeo-python-client/issues/934

pjonsson: GitHub had collapsed the file that you moved things from so I didn't notice, sorry about that. I completely agree that moving code and fixing issues in that code should be in separate PRs.

openeo/rest/_connection.py
soxofaan: I think that's fine. 
E.g. `_download_all_at_once` can't know (and shouldn't) all the ways it will be called

openeo/rest/job.py
```diff
@@ -722,6 +674,258 @@ def download_files(
...
+        - job result metadata (the root STAC collection)
+        - linked STAC items: metadata and the result assets
+        - additionally linked metadata (e.g. "derived_from" documents)
+
+        .. warning:: this is an experimental API, subject to change.
+
+        :param target: folder path to download to
+        :param rewrite_references: whether to rewrite (item/asset/...) HREFs in the STAC documents.
+            in the downloaded STAC collection to point to the local files
+            instead of the original URLs.
```
soxofaan, line 700: doc mixup here

openeo/rest/job.py
```diff
@@ -722,6 +674,258 @@ def download_files(
...
+        self._root_dir = Path(target or Path.cwd() / job.job_id)
+        if self._root_dir.exists() and not self._root_dir.is_dir():
+            raise OpenEoClientException(f"Download target {self._root_dir} exists but isn't a folder.")
+        self._rewrite_references = rewrite_references
+        # TODO: also support passing a `json.dump`-style callable to customize json dumping
+        self._json_dumping = {"ensure_ascii": False, **(json_dumping or {})}
+        self._downloaded: List[Path] = []
+        self._on_download_failure = on_download_failure
+        self._path_templates = {
+            "collection": "job-results.json",
```
soxofaan: expose this default as a class constant

openeo/rest/job.py
```diff
@@ -722,6 +674,258 @@ def download_files(
...
+            if self._on_download_failure in {"warn"}:
+                logger.warning(message, exc_info=True)
+            else:
+                # TODO: other handling strategies?
+                #       e.g. collect all failures and raise a single exception at the end,
+                if self._on_download_failure != "raise":
+                    logger.warning(f"Unknown on_download_failure strategy {self._on_download_failure!r}")
+                raise JobResultDownloadException(message) from e
+
+    def _check_download_path(self, path: Path):
```
soxofaan: rename to something like `assert_new_download` ?

or encapsulate this repeated pattern just better (e.g. with context manager):
```
self._check_download_path(path)
...
self._downloaded.append(path)
```

openeo/rest/job.py
soxofaan: let's make this a future work item

openeo/rest/job.py
soxofaan: I think this download collision feature is also not covered with tests

soxofaan: I tried the context manager approach, but it felt too much overkill at the moment (in terms of LoC). Kept it a bit more simple, but still encapsulated in `_DownloadTracker`

openeo/rest/job.py
```diff
@@ -722,6 +674,275 @@ def download_files(
 
         return downloaded
 
+    def download_as_collection(
```
soxofaan, line 677: while there is plenty of coverage of `_JobResultDownloader`, this `download_as_collection` wrapper need some basic tests too

PR #938, opened 2026-09-17
Use standard retry logic in _download_ranged

### LINKED ISSUE #934: Use standard retry logic in _download_ranged
https://github.com/Open-EO/openeo-python-client/blob/88d793beb7b824f336e2d9b061e8ecafc0b0d6f1/openeo/rest/job.py#L557-L571

as suggested in https://github.com/Open-EO/openeo-python-client/pull/933#discussion_r3770142673, replace this DIY retry logic with standard retry utilities like `urllib3.util.Retry`.

Actually, `RestApiConnection` already has such a retry config by default since https://github.com/Open-EO/openeo-python-client/pull/770 (for https://github.com/Open-EO/openeo-python-client/issues/441 and https://github.com/Open-EO/openeo-python-client/issues/764)
so the DIY retry logic can probably just be dropped

niukanen1:
Fixes #934

Removes the hand-rolled per-range retry loop from `_download_ranged`. Transient failures are already retried by the urllib3 Retry configuration mounted on the connection's session, so the DIY loop only added divergent behavior: different retry counts, a different status-code set (408/500/501 retried here but not by the session), and `OpenEoApiPlainError` where the session raises `RetryError`

The unused `MAX_DOWNLOAD_RETRIES_PER_RANGE` / `RETRIABLE_DOWNLOAD_STATUSCODES` constants are gone

Tested with `pytest tests/rest/test_download_ranged_retry.py tests/rest/test_job.py`: 133 passed — including a new httpretty-based test proving a 503 on a range is retried by the session and the download completes


tests/rest/test_download_ranged_retry.py
```diff
@@ -0,0 +1,57 @@
...
+from unittest import mock
+
+import httpretty
+import pytest
+
+from openeo.rest._connection import RestApiConnection
+from openeo.utils.http import session_with_retries
+
+
+class TestDownloadRangedStandardRetry:
```
soxofaan: we usually don't create new, separate test files per feature, can you move this to an existing  test module?

e.g. there are already existing tests for the chunked/ranged downloading in `test_job.py`

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- `_download_ranged` no longer implements its own retry loop; transient failures are retried by the standard urllib3 retry configuration of the connection's session ([#934](https://github.com/Open-EO/openeo-python-client/issues/934))
```
soxofaan: append new changelog items  to the bottom of a section

tests/rest/test_download_ranged_retry.py
```diff
@@ -0,0 +1,57 @@
...
+    configuration of the connection's session.
+    """
+
+    @pytest.fixture(autouse=True)
+    def _auto_httpretty_enabled(self):
+        with httpretty.enabled(allow_net_connect=False):
+            yield
+
+    def test_transient_503_on_range_is_retried_by_session(self, tmp_path, time_sleep):
+        content = b"0123456789" * 100  # 1000 bytes -> two 500-byte ranges
```
soxofaan: this content (repetition of the same 10 chars) does not allow to uniquely identify the individual 500char chunks as they will be exactly the same

tests/rest/test_download_ranged_retry.py
```diff
@@ -0,0 +1,57 @@
...
+                    status=206,
+                    body=content[500:],
+                    adding_headers={"Content-Range": f"bytes 500-999/{len(content)}"},
+                ),
+            ],
+        )
+
+        connection = RestApiConnection(root_url="https://example.test", session=session_with_retries())
+        target = tmp_path / "file.bin"
+        connection._download_ranged(url=url, target=target, file_size=len(content), range_size=500)
```
soxofaan: I would avoid calling the internal `_download_ranged` here, I'd call the public `download_url` method instead

soxofaan (commented):
hi, thanks for contributing

some notes on you PR

niukanen1:
Thanks for the review — addressed all four points in d6a57c6:

- the test now lives in `test_job.py`, next to the other ranged download tests
- it drives the public `download_url` method instead of `_download_ranged`
- the test content is built so each range chunk is unique (distinct byte pattern per 500-byte block, asserted in the test)
- the changelog entry moved to the bottom of the Changed section

tests/rest/test_job.py
```diff
@@ -821,6 +821,59 @@ def test_get_results_download_file_ranged(job_with_chunked_asset_using_head: Bat
...
+                body=content[256:].decode("latin-1"),
+                adding_headers={"Content-Range": f"bytes 256-351/{len(content)}"},
+            ),
+        ],
+    )
+
+    httpretty.register_uri(
+        httpretty.GET,
+        uri=API_URL + "/.well-known/openeo",
+        body=json.dumps({"api_version": "1.0.0", "endpoints": [{"path": "/credentials/basic", "methods": ["GET"]}]}),
```
soxofaan: this is not the correct body to return on `/.well-known/openeo`
I'd try with not mocking this request, I think that should still work

tests/rest/test_job.py
```diff
@@ -821,6 +821,59 @@ def test_get_results_download_file_ranged(job_with_chunked_asset_using_head: Bat
...
+    """#934: transient failures during ranged download are retried by the
+    standard urllib3 Retry of the connection's session, not a custom loop."""
+    content = b"R4ng3d-D4t4" * 32  # 352 bytes -> one full and one partial 256-byte range
+    assert len(set(content[i : i + 256] for i in range(0, len(content), 256))) == 2
+
+    # HEAD body is ignored, but httpretty validates registered Content-Length against it.
+    httpretty.register_uri(
+        httpretty.HEAD,
+        uri=API_URL + "/dl/ranged.bin",
+        body=content.decode("latin-1"),
```
soxofaan: a HEAD request doesn't need a body, or am I misunderstanding?

tests/rest/test_job.py
```diff
@@ -821,6 +821,59 @@ def test_get_results_download_file_ranged(job_with_chunked_asset_using_head: Bat
         assert f.read() == TIFF_CONTENT
 
 
+@httpretty.activate(allow_net_connect=False)
+def test_download_url_ranged_retries_transient_503(tmp_path):
+    """#934: transient failures during ranged download are retried by the
+    standard urllib3 Retry of the connection's session, not a custom loop."""
+    content = b"R4ng3d-D4t4" * 32  # 352 bytes -> one full and one partial 256-byte range
+    assert len(set(content[i : i + 256] for i in range(0, len(content), 256))) == 2
```
soxofaan: I still find this `content` juggling a bit convoluted

why not return just  `"A" * 200` on the first GET, `"B" * 100` on the second and  check that you got `"A" * 200 + "B" * 100` in the end?

openeo/rest/_connection.py
```diff
@@ -331,25 +310,15 @@ def _download_ranged(
         chunk_size: int = DEFAULT_DOWNLOAD_CHUNK_SIZE,
         range_size: int = DEFAULT_DOWNLOAD_RANGE_SIZE,
     ) -> None:
+        # Retries on transient failures (429/502/503/504) are handled by the
+        # urllib3 Retry mounted on this connection's session (see
+        # session_with_retries), so no per-range retry loop is needed here.
```
soxofaan: I find this comment a bit misleading as it is (at this point) not guaranteed what kind of session or mounts are active

soxofaan (commented):
I also think the changelog entry needs alignment with latest master

PR #942, opened 2026-09-22
Drop support for Python 3.8

### LINKED ISSUE #717: Drop support for Python 3.8

soxofaan:
Drop support for python 3.8 and related dependency pins, adaptation logic, and other hacks

for #717


soxofaan:
@copilot can you review this PR (no need to run code, run tests or make commits).
focus on dropping python 3.8 support and see if everything is covered

pyproject.toml
```diff
@@ -1,5 +1,4 @@
 [build-system]
-# Setuptools 75 is the latest version that works with Python 3.8.
```
pjonsson, line 2: Removing this comment eliminates all the traces of us wanting to bump setuptools to a more recent version from the source code. You should probably make a note somewhere else so the bumping doesn't get lost.

soxofaan: indeed, I was hesitant about this

the problem is that keeping the comment as is (i.e. the "Python 3.8" part)  makes no sense when dropping Python 3.8.
Updating the comment to align with the new minimum Python requirement (Python 3.9) would also require to bump actual setuptools constraint (to `>=82` I think, but there is no urgent reason as far as I know to also drag this in here.

So the easiest solution is just removing the comment (as it will lose relevance).
Note that this not removes all traces, the original comment will still be discoverable when doing a git blame on the `requires = ["setuptools>=75"]` line.

But again, no strong opinion here. I'm also fine with bumping the setuptools constraint to `>=82` if you think that makes more sense

pjonsson: I think it makes sense to limit the scope as much as possible so we can get rid of the EOL versions of Python as quickly as possible. I guess this is the reminder to bump it "soon" after this PR is merged.

(The Ceph August updates tightening the checks on signed S3 URIs makes me keen on getting a new release of this package that doesn't run HEAD on them to make our open-EO usable at all, so I don't want to block your next release on a bunch of nice to haves here.)

soxofaan:
Note: this PR conflicts with https://github.com/Open-EO/openeo-python-client/pull/867 as that one fiddles with `_PYSTAC_1_9_EXTENSION_INTERFACE` which is removed here.

Getting https://github.com/Open-EO/openeo-python-client/pull/867 merged first is probably better in term of ease of resolving conflicts

soxofaan:
@copilot I had to do a rebase and resolve conflicts, can you check I didn't miss or overlook anything?

soxofaan:
@copilot OMG please don't make commits unless I explicitly ask for that. Is there a setting to tweak that default behavior?

soxofaan:
Ding dong, the witch is dead

